PR #29, opened 2019-08-02
compat for CSTParser/Tokenize

pfitzseb:


domluna:
We require this commit https://github.com/julia-vscode/CSTParser.jl/commit/541c609c7dc7d652f94034035daa75194c02094c

So would it be 0.6.2-DEV instead of 0.6.1?

pfitzseb:
Mh, we should just bump them to tag a new version and then switch to 0.6.2.

domluna:
@pfitzseb zac tag v0.6.2

Issue #27, opened 2019-08-01
`=` additional prettification


domluna:
### 1. When `=` is part of a function definition, i.e. `foo() = body`

(A) Consider the following code:

```julia
foo() = 
    if cond
       a
    else
       b
    end
```

(B) Currently this is formatted as (unless it exceeds the allowed margin length):

```julia
foo() = if cond
   a
else
   b
end
```

It's highly unlikely this will exceed the margin required for the if expression to nest, but IMO it's clear A looks better than B. What if A was the default during the prettify stage? This would include more than just `If` expressions:

```
CSTParser.If
CSTParser.Do
CSTParser.Begin
CSTParser.Quote
CSTParser.Try
CSTParser.For
CSTParser.While
CSTParser.Let
```

### 2. When `=` is an assignment, i.e. `foo = body`

I'm not sure this occurs frequently but I have seen it come up. When `body` is one of the above types you can make a good argument body should start on the next line with an indent. If it's not the default during the prettify stage then it should at least be nest-able.

```
foo =
    body
```

cc @StefanKarpinski

StefanKarpinski:
Both seem reasonable to me.

domluna:
done

PR #30, opened 2019-08-04
Prettify eq

domluna:
ref #27


domluna:
@JuliaRegistrator register()

Issue #34, opened 2019-08-14
for = vs for in normalization


StefanKarpinski:
The rule of thumb that I use for `for` loops is that if the thing you're iterating over is a literal range then you use `=` whereas if it's anything else (e.g. an object) then I use `in`. So, for example, these would be "correct":
```jl
for i = 1:n
    println(i)
end

for i in itr
    println(i)
end
```
These, on the other hand would be "incorrect":
```jl
for i in 1:n
    println(i)
end

for i = itr
    println(i)
end
```
Part of the reasoning is that `for i = 1:n` looks like it assigns `i` to 1, then 2, then ... `n` — at least in common mathematical / pseudocode notation, whereas `for i = itr` looks like it assigns `i = itr` which is not what it does. So one could argue "why use/allow `for i = 1:n` at all? Why not just use `in` uniformly and write `for i in 1:n` and `for i in itr`? Some people surely would prefer this. However, this superficial syntax difference can help catch a common error, which is writing `for i = n` when you meant to write `for i = 1:n`. If we require `in` there then the programmer will notice that `for i in n` is not what they meant to write and catch the bug.

So I would propose normalizing `for =` versus `for in` based on this rule. Thoughts?

domluna:
I'm in favour of this. 

Throwback https://github.com/julia-vscode/DocumentFormat.jl/pull/26/files#diff-8f8b649677f922163282f3d951be629aR89-R104.

Additionally this is a best practice for the prevalent style guides:

- http://www.juliaopt.org/JuMP.jl/latest/style.html
- https://github.com/jrevels/YASGuide

KwatME:
Good idea.

diegozea:
> why use/allow for i = 1:n at all? Why not just use in uniformly and write for i in 1:n and for i in itr? Some people surely would prefer this.

I'm some people. I've found `for =` difficult to read, so I always use `for in`. I would prefer `for ∈` instead of `for =` if a difference should be made.

KwatME:
It's more difficult to type ∈. Also, ∈ is rarely found in other languages.

singularitti:
I am "some people" also. In fact, some [nonofficial guide](https://github.com/jrevels/YASGuide/blob/master/README.md#L71) has suggested not to use `for ... = ...`.

KwatME:
I'm for always using `in`.

odow:
Going to chime in on this dead issue that
[...]

PR #35, opened 2019-08-17
Eq in normalization

domluna:
Transforms `=` to `in` if part of a for loop condition, ref https://github.com/domluna/JuliaFormatter.jl/issues/34. Kind of funny, calling format on the package after this changed a loop condition. I guess it works 😄 

@StefanKarpinski any other test cases you think would be good?

### DIFF
--- a/src/pretty.jl
+++ b/src/pretty.jl
@@ -131,7 +131,7 @@ function length_to_newline(x::PTree, start = 1)
     x.typ === PLACEHOLDER && (return 0, true)
     is_leaf(x) && (return length(x), false)
     len = 0
-    for i = start:length(x.nodes)
+    for i in start:length(x.nodes)
         n = x.nodes[i]
         ln, nl = length_to_newline(n)
         len += ln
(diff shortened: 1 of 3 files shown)


StefanKarpinski:
Isn't this the opposite rule from what I proposed?

domluna:
Hmmm, I guess misinterpreted it and now I'm kind of confused.

> However, this superficial syntax difference can help catch a common error, which is writing `for i = n` when you meant to write `for i = 1:n`. If we require `in` there then the programmer will notice that `for i in n` is not what they meant to write and catch the bug.

Doesn't argue for replacing `=` with `in`?

Follow up - would we use `=` only if it's a range like `1:10`?

domluna:
Using the correct/incorrect examples #34 

Correct (these remain the same)

```julia
for i = 1:n
    println(i)
end

for i in itr
    println(i)
end
```
Incorrect

```julia
# remains the same
for i in 1:n
    println(i)
end

# `=` is replaced with `in`
for i = itr
    println(i)
end
```

So only the last example would warrant an actual transformation. Is this in line with what you meant?

StefanKarpinski:
> Follow up - would we allow `=` only if it's a range like `1:10`?

Yes, and only allow `=` if the RHS is a range. So both of the later "incorrect" examples would be transformed:
* `for i in 1:n` would be turned into `for i = 1:10` and
* `for i = itr` would be turned into `for i in itr`.

domluna:
@StefanKarpinski I think this does what it should now.

Issue #42, opened 2019-08-26
Change formal.jl script to show help when called without arguments


karlwessel:
At the moment, the `format.jl` script formats every `.jl`-file it can find in the current path when it is called without any argument. 

For stupid people like me this can lead to format every `.jl`-file in their whole home directory when they call it without any argument.

Maybe it would be more safe to print some usage information when called without any arguments instead of happily formating the .julia directory (and everything else)? :-)

Also some output on which file it is currently formating would probably be good.

domluna:
fixed by #48

karlwessel:
Thanks!

PR #43, opened 2019-09-01
display help when command line script is invoked without parameters

karlwessel:
This is my take to solve ticket #42 

I used [ArgParse.jl](https://github.com/carlobaldassi/ArgParse.jl) for argument parsing since it also automatically displays help when no or wrong arguments are given.

I had to move the `parse_commandline` method to the module itself ("src/commandline.jl") to make sure that `ArgParse` package is available in the scope when the command line script is executed.

The script changed in the way thath it can only handle one file to format as argument now. I don't know how necessary it is for it to be able to handle multiple files as arguments at once.

I also added some tests for the script.

Let me know if everything looks ok to you.

### DIFF
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -1,5 +1,11 @@
 # This file is machine-generated - editing it directly is not advised
 
+[[ArgParse]]
+deps = ["Compat", "Test", "TextWrap"]
+git-tree-sha1 = "14d5789a99e6bea3e258d668ec09359a9feaf2d1"
+uuid = "c7e460c6-2fb9-53a9-8c5b-16f535851c63"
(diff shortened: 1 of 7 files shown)


domluna:
Thanks for this @karlwessel!

I'm not a fan of having to adding `ArgParse` to to the module just for utility in the CLI script. I think we could extend the current script to accommodate all of this and more. I've made some improvements to the current script in #48.

karlwessel:
Sorry for taking so long to answer, I was a bit lazy the last two week.

Although I am a big fan of using packages that do small things right I can totally understand that you don't want to add another dependency for this.

Thanks for fixing this!

PR #49, opened 2019-09-07
Add format github action

domluna:
whyyy is this soo hard

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -6,6 +6,7 @@ jobs:
   format:
     runs-on: ubuntu-latest
     steps:
+    - uses: actions/checkout@master
     - uses: domluna/JuliaFormatter-action@master
       with:
         args: -v .
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -1,6 +1,6 @@
 name: Format
 
-on: [push, pull_request]
+on: [push]
 
 jobs:
   format:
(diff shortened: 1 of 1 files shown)


odow:
Why not just put the
```Julia
import Pkg
Pkg.add("JuliaFormatter")
```
at the top of the `format.jl` file?

That's only going to get the latest tagged version though. Is there an environment variable that let's you know which PR the action is being called from? Then you could use that instead.

odow:
You could also try printing the package status and project name etc at the top of `format.jl` in addition to `pwd()` etc.

odow:
Oh, so `action/checkout` clones the PR `$GITHUB_WORKSPACE`. So then you just need to start with `julia --project=$GITHUB_WORKSPACE`. 

That means there's probably no need to clone `JuliaFomatter` in your docker file?

domluna:
Looks this is is working now 😄

odow:
Is there a plan to make it commit the changes to the PR? Or throw an error if it detects that the formatter changed something?

domluna:
@odow There are a bunch of ways it could go. 

The most minimal thing would be to just format the diffs, but we don't have functionality for that at the moment. I think formatting the changed files is a good way to go since for large projects formatting everything at once can be pretty jarring. On that note I'm not sure how to get the changed files, `actions/checkout` clones the repo so I'm sure we can use some git commands to figure that out.

Issue #44, opened 2019-09-01
Either always add or never all tail comma, when formatting items within an iterable


KwatME:
Sometimes the formatter outputs:
```julia
f(
    a,
    b,
)
```

Some other times the formatter outputs:
```julia
if true

    f(
        a,
        b
    )

end
```

Always adding tail comma is a good idea but, as you can see in the follow before-after screenshots, the formatter is not adding the tail comma at the cursor site.

## Before
<img width="589" alt="Screen Shot 2019-09-01 at 11 26 34" src="https://user-images.githubusercontent.com/7741548/64080662-9c951280-ccab-11e9-964c-c1a70f26bfca.png">

## After
<img width="510" alt="Screen Shot 2019-09-01 at 11 28 02" src="https://user-images.githubusercontent.com/7741548/64080666-a74fa780-ccab-11e9-94cd-cfc69292b7d7.png">

I like this project. Many thanks :)

domluna:
Right now it adds a trailing comma only if there was one already present in the code. 

I think @StefanKarpinski mentioned before he was in favour of always adding a trailing comma. The main benefit I see is when adding a new argument you don't have to add a new comma, although that would only apply if the arguments are on different lines. Perhaps the rule could be to always add a trailing comma if it's nested otherwise follow the current behavior.

KwatME:
Sounds good. Also, if the implementation can be simplified, always adding tailing comma is a good idea too (to alleviate workload for you).

KwatME:
Is it possible to try, as @StefanKarpinski also favors, always adding tailing comma, please? I think the simplicity of this rule is elegant and generalizes to type, list, tuple, and all other iterables. Thank you @domluna.

diegozea:
I do not like the idea, trailing commas makes you think that there's still another element in the list. Also, if you do not use formatter, when you add the element you need to type the new trailing comma. To alleviate the workload, it would be better make Juno to automatically add a trailing commas when you aren't closed your list and automatically delete it once you close it.

KwatME:
In an ideal world, world without any mistake, I think being smart about when to have or omit such comma is good. So I agree with you @diegozea.

But it may be easier to code and maintain the simpler rule of always having (or not having) the tailing comma. I can imagine the complexity of [], (), {}, + to perhaps increase in the future. So I'm for simpler rule of always having the tailing comma.

Finally, this simpler rule is not only more robust but generalizable across languages such as Python.

[...]

PR #50, opened 2019-09-09
Trailing comma

domluna:
Adds a trailing comma to iterable types if they're nested. Ref #44


test/files/PProf.jl
```diff
@@ -258,10 +258,11 @@ Start or restart the go pprof webserver.
 - `file::String`: Profile file to open.
 """
 function refresh(
+    ;
```
StefanKarpinski: Funky case here seems to add a double semicolon!

Issue #65, opened 2019-09-14
Unicode in comments causes StringIndexError


odow:
JuliaFormatter cannot deal with unicode in comments. This also happens with multiline comments.
```Julia
julia> JuliaFormatter.format_text("α = 1")
"α = 1"

julia> JuliaFormatter.format_text("1 # a")
"1 # a"

julia> JuliaFormatter.format_text("1 # α")
ERROR: StringIndexError("1 # α", 6)
Stacktrace:
 [1] string_index_err(::String, ::Int64) at ./strings/string.jl:12
 [2] getindex at ./strings/string.jl:248 [inlined]
 [3] getline at /Users/oscar/.julia/packages/JuliaFormatter/ttR0T/src/JuliaFormatter.jl:98 [inlined]
 [4] print_inlinecomment at /Users/oscar/.julia/packages/JuliaFormatter/ttR0T/src/print.jl:64 [inlined]
 [5] print_leaf(::Base.GenericIOBuffer{Array{UInt8,1}}, ::JuliaFormatter.PTree, ::JuliaFormatter.State) at /Users/oscar/.julia/packages/JuliaFormatter/ttR0T/src/print.jl:14
 [6] print_tree(::Base.GenericIOBuffer{Array{UInt8,1}}, ::JuliaFormatter.PTree, ::JuliaFormatter.State) at /Users/oscar/.julia/packages/JuliaFormatter/ttR0T/src/print.jl:29
 [7] #format_text#52(::Int64, ::Int64, ::Function, ::String) at /Users/oscar/.julia/packages/JuliaFormatter/ttR0T/src/JuliaFormatter.jl:163
 [8] format_text(::String) at /Users/oscar/.julia/packages/JuliaFormatter/ttR0T/src/JuliaFormatter.jl:131
 [9] top-level scope at none:0

julia> JuliaFormatter.format_text("# α")
"# α"
```

domluna:
@odow can you provide the multiline comment test case?

odow:
Here you go:
```Julia
"""
#=
α
=#
x = 1
"""
```

devmotion:
This is a very old issue but I think I just ran into a similar Unicode-related indexing problem. I don't have a MWE yet but the error seems to be caused by https://github.com/domluna/JuliaFormatter.jl/blob/8a1e69229fe74d331528678ae007b25670675a8e/src/styles/default/pretty.jl#L323 and the error message is
```julia
ERROR: TaskFailedException

[...]

PR #67, opened 2019-09-15
 Handle Unicode comments better

domluna:
Handle Unicode comments better - ref #65. Use Integer and AbstractString in for type/function definitions

### DIFF
--- a/src/JuliaFormatter.jl
+++ b/src/JuliaFormatter.jl
@@ -58,8 +58,14 @@ function Document(text::AbstractString)
                 end
             end
         elseif t.kind === Tokens.COMMENT
-            ws = prev_tok.kind !== Tokens.WHITESPACE ? 0 :
-                 count(c -> c == ' ', prev_tok.val)
+            ws = 0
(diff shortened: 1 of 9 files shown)


odow:
Why is it necessary to use `Integer` and `AbstractString`?

domluna:
Best practice to use wider types? There was a mix of Int/Integer and AbstractString/String in type/function defs so I figured I'd normalize those to use the wider type. In practice I think this is only useful for the `val` type of a PTree, which is sometimes a `SubString` instead of a `String`.

StefanKarpinski:
The integer fields at least, I would leave as `Int`: they are indices into strings which should be `Int` for any kind of abstract string and guaranteed for `String`. Unless you expect the strings to be very large and you want to avoid transcoding, I would also just use the `String` and let various kinds of strings be converted to `String` everywhere (or most often, just used as-is since `String` is common). The general rule of thumb is that for dispatch, it's good to leave things nice and generic, only requiring what you need for methods to work, but for places like fields or data structures, use the most concrete type you can.

PR #70, opened 2019-09-16
Normalize all `=` to `in` in for loops

odow:
Since #34 was closed, there has been a steady stream of people disagreeing. I get that it is a contentious issue with no settled style in the community, but this simplifies the rule:

- Use `for i in iter` except it `iter` is a literal range, then use `for i = 1:n`.

to

- Always use `for i in iter`.

Revises https://github.com/domluna/JuliaFormatter.jl/pull/35. Re-closes the already closed issue: https://github.com/domluna/JuliaFormatter.jl/issues/34

### DIFF
--- a/src/pretty.jl
+++ b/src/pretty.jl
@@ -861,33 +861,12 @@ function p_let(x, s)
     t
 end
 
-
-
-# Transforms
-#
-# for i = iter body end
-#
-# =>
-#
-# for i in iter body end
-#
-# AND
-#
-# for i in 1:10 body end
-#
-# =>
-#
-# for i = 1:10 body end
-#
-# https://github.com/domluna/JuliaFormatter.jl/issues/34
+# Transform all `for i = iter body end` to `for i in iter body end` with no
+# exceptions.
 function eq_to_in_normalization!(x)
     if x.typ === CSTParser.BinaryOpCall
-        op = x.args[2]
-        arg2 = x.args[3]
-        if op.kind === Tokens.EQ && !is_colon_op(arg2)
+        if x.args[2].kind === Tokens.EQ
             x.args[2].kind = Tokens.IN
-        elseif op.kind === Tokens.IN && is_colon_op(arg2)
-            x.args[2].kind = Tokens.EQ
         end
     elseif x.typ === CSTParser.Block || x.typ === CSTParser.InvisBrackets
         for a in x.args
--- a/src/pretty.jl
+++ b/src/pretty.jl
@@ -77,7 +77,7 @@ function add_node!(t::PTree, n::PTree, s::State; join_lines = false, max_padding
         join_lines = true
         loc = s.offset > length(s.doc.text) && t.typ === CSTParser.TopLevel ?
               loc = cursor_loc(s, s.offset - 1) : cursor_loc(s)
-        for l = t.endline:loc[1]
+        for l in t.endline:loc[1]
             if has_semicolon(s.doc, l)
                 # @info "found semicolon" l
                 n.startline = l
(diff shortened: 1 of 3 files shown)


StefanKarpinski:
How do you propose catching the common issue where people write `for i = n` when they meant `for i = 1:n`?

StefanKarpinski:
Looking through more cases, Base + stdlib Julia code is actually remarkably consistent about following the rule that is currently implemented.

odow:
> How do you propose catching the common issue where people write for i = n when they meant for i = 1:n?

Do you mean from a correctness standpoint? Or a stylistic standpoint?

```Julia
julia> function foo(n)
           for i in n
               println(i)
           end
       end
foo (generic function with 1 method)

julia> function bar(n)
           for i = n
               println(i)
           end
       end
bar (generic function with 1 method)

julia> foo(5)
5

julia> bar(5)
5
```

StefanKarpinski:
Yes, I mean from a correctness standpoint. If the convention is that one uses `in` with an iterator and `=` with a literal range, then you have an additional bit of information about user intent. That bit is redundant, but it acts as a "parity bit" and the user gets a warning when the parity bit doesn't match the usage. For example if the user writes `for i = n` then the `=` doesn't match the nature of the `n` which is not a literal range, so the linter or formatter can call their attention to the mismatch. You are proposing stylistically disallowing the usage of that parity bit to convey intent.

odow:
Okay, fair point. I see the benefit, and I've lost some zeal for this change.

To summarize: the trade off is therefore between people making the mistake `for i = n` _and_ using a linter to find catch this error, and making everyone writing Julia learns when to use the difference between `=` and `in`. Against standardizing on `in` and having more incorrect code. 

I'm not sure which I prefer...

I guess the root issue is that numbers are iterable. (Slightly off-topic, but was there a reason for this?)

mlubin:
> For example if the user writes for i = n then the = doesn't match the nature of the n which is not a literal range, so the linter or formatter can call their attention to the mismatch.

FWIW, given a good auto-formatter, the default workflow for me would be "auto format and forget", which means that if I wrote "for i = n", it would just get corrected without raising a flag to "for i in n" regardless of if I intended to write a literal range.

StefanKarpinski:
Agree, it's probably not very effective just as part of the standard formatting. To be effective, it needs to also be something that linters can warn about, but for that to be possible, it needs to be standard. Even in situation where one blindly runs the formatter and later goes bug hunting, at least the fact that it's `for i in n` gives more visual clue that the RHS isn't the intended range.

> I guess the root issue is that numbers are iterable. (Slightly off-topic, but was there a reason for this?)

https://github.com/JuliaLang/julia/issues/7903

domluna:
I think generally `=` is used for literal ranges outside of repos that strictly follow https://github.com/jrevels/YASGuide or http://www.juliaopt.org/JuMP.jl/dev/style/.

At first I was in the use `in` for everything camp as well but the describing the intent and possibly catching an error as @StefanKarpinski mentioned above has grown on me.

Thoughts on allowing use of either `=` or `in` for a range. So these would both be correct:

```julia
for i = 1:10
end

for i in 1:10
end
```

But this would be incorrect:

```julia
for i = I
end
```

I believe this still captures the initiative for #34 (might be wrong) and keeps the `in` folks happy.

mlubin:
@domluna I would prefer that you chose exactly one of the two options and make it a rule. We shouldn't carve out exceptions for special interest groups :)

domluna:
Oh man, "with great power comes great responsibility" a wise man once said.

We're gonna stick to the current implementation then. @odow I appreciate the effort ❤️

Issue #80, opened 2019-09-27
Formatting of let blocks


saschatimme:
I just stumbled about this

The following code (which I consider nicely formatted)
```
s = let
    tracker, starts = pathtracker_startsolutions([x^2+y^2+z^2 - 1; L₁]; system = FPSystem)
    solution(track(tracker, first(starts)))
end
```
gets formatted to

```
s =
        let
                tracker, starts = pathtracker_startsolutions(
                        [x^2 + y^2 + z^2 - 1; L₁];
                        system = FPSystem,
                )
                solution(track(tracker, first(starts)))
        end
```

domluna:
This is intentional #27. 

Site note: Is your indentation set to 8 or is that just GH formatting?

saschatimme:
I think GitHub messed it up.  Still think that the white spaces are somehwat excessive here, but if this is the decision made I am also fine.

domluna:
For this example it’s a toss up but I think In the majority of cases the
latter returns a nicer output

StefanKarpinski:
Having the `s = let` on the same line would be reasonable.

PR #82, opened 2019-09-28
RFC: Assignment op RHS block nesting

domluna:
This spun off of #80.

If `Begin`, `Quote`, or `Let` (with no args) are on the RHS of an assignment op (not function def). Then don't nest over the assignment op. The idea is if it's *just* a keyword, keep it on the same line. `Try` is exempted from this, I think it fits better with being nested, but that's just my opinion.

### DIFF
--- a/src/pretty.jl
+++ b/src/pretty.jl
@@ -1105,10 +1105,8 @@ end
 block_type(x::CSTParser.EXPR) =
     x.typ === CSTParser.If ||
     x.typ === CSTParser.Do ||
-    x.typ === CSTParser.Begin ||
-    (x.typ === CSTParser.Quote && x.args[1].kind === Tokens.QUOTE) ||
     x.typ === CSTParser.Try ||
-    x.typ === CSTParser.For || x.typ === CSTParser.While || x.typ === CSTParser.Let
+    x.typ === CSTParser.For || x.typ === CSTParser.While || (x.typ === CSTParser.Let && length(x) > 3)
 
 nest_assignment(x::CSTParser.EXPR) = CSTParser.is_assignment(x) && block_type(x.args[3])
 
--- a/src/pretty.jl
+++ b/src/pretty.jl
@@ -1106,7 +1106,8 @@ block_type(x::CSTParser.EXPR) =
     x.typ === CSTParser.If ||
     x.typ === CSTParser.Do ||
     x.typ === CSTParser.Try ||
-    x.typ === CSTParser.For || x.typ === CSTParser.While || (x.typ === CSTParser.Let && length(x) > 3)
+    x.typ === CSTParser.For ||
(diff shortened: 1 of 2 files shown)


StefanKarpinski:
This looks great to me! Much more natural this way. I like the insight about keywords.

StefanKarpinski:
Were there other RHS constructs to which the same rule could be applied? Doesn't it naturally apply to `A = [ ... ]` for example? Or even `d = Dict( ... )` where the RHS needs to be split across many lines? (Maybe that's already handled that way.)

domluna:
I'm not sure what you mean but

```julia
d = Dict(a ,b , c)
```

is nested to

```julia
d = Dict(
    a,
    b,
    c,
)
```

odow:
https://github.com/odow/SDDP.jl/blob/a11366d33590234c893ff3631ae932518607caaf/examples/FAST_hydro_thermal.jl#L14-L35

One thing I didn't bring up earlier, because I assumed it was a deliberate design decision is
```Julia
model =
    SDDP.PolicyGraph(
        SDDP.LinearGraph(2),
        bellman_function = SDDP.BellmanFunction(lower_bound = 0.0),
        optimizer = with_optimizer(GLPK.Optimizer),
    ) do sp, t
        @variable(sp, 0 <= x <= 8, SDDP.State, initial_value = 0.0)
        @variables(sp, begin
            y >= 0
            p >= 0
            ξ
        end)
        @constraints(sp, begin
            p + y >= 6
            x.out <= x.in - y + ξ
        end)
        RAINFALL = (t == 1 ? [6] : [2, 10])
        SDDP.parameterize(sp, RAINFALL) do ω
            JuMP.fix(ξ, ω)
        end
        @stageobjective(sp, 5 * p)
    end
```
I would prefer that it was formatted as
```julia
model = SDDP.PolicyGraph(
    SDDP.LinearGraph(2),
    bellman_function = SDDP.BellmanFunction(lower_bound = 0.0),
    optimizer = with_optimizer(GLPK.Optimizer),
) do sp, t
    @variable(sp, 0 <= x <= 8, SDDP.State, initial_value = 0.0)
    @variables(sp, begin
        y >= 0
        p >= 0
        ξ
    end)
    @constraints(sp, begin
        p + y >= 6
        x.out <= x.in - y + ξ
    end)
    RAINFALL = (t == 1 ? [6] : [2, 10])
    SDDP.parameterize(sp, RAINFALL) do ω
        JuMP.fix(ξ, ω)
    end
    @stageobjective(sp, 5 * p)
end
```

domluna:
@odow that's a bug, it should reformat to the latter, since the arguments of `SDDP.PolicyGraph` nested this fits in the width margin. I think it's because the `Call`s don't reset their length when nested so in the eyes of the formatter it's still on one line and thus doesn't move `SDDP.PolicyGraph` back up.

Issue #85, opened 2019-10-02
Custom style rules


nickrobinson251:
Thanks for making this! I would very much like to use this package to implement a style guide (https://github.com/invenia/BlueStyle/issues/7)

- How may I specify custom style rules be used?  e.g. If I would like add the rule that `|>` must never be used ;)
- If not possible right now, is there an intention/plan to support custom style rules?

domluna:
I was thinking about integrating linting pertaining to style earlier this week. I think could could be added as a custom pass done similar to `pretty` or `nest!`. The signature could be:

```julia
struct BlueLinter end

# maybe CSTParser.EXPR instead of PTree
function (linter::BlueLinter)(x::PTree, s::State)
    ...
end
```

Then `lint` performs a check dependent on the type, `x.typ` and potentially adds a message.

```
struct LintError
    line::Int
    err::String
end
```

https://github.com/domluna/JuliaFormatter.jl/blob/master/src/JuliaFormatter.jl#L102-L109

```julia
mutable struct State
    doc::Document
    indent_size::Int
    indent::Int
    offset::Int
    line_offset::Int
    margin::Int
    # new field
    lints::Vector{LintError}
end
```

Then outputs the lint error along with the formatted result. The formatter could match the old line with the new line (from formatted output).

domluna:
It could be fairly straightforward depending on the how complicated some of the style checks are.

nickrobinson251:
I guess there's a more general question here: 

- Should we try to seperate the core formatting tools from any particular style, i.e. don't "hard code" the style, as much as possible?
- In other words,  prioritise making a configurable tool that allows automating "style guides" (rather than `JuliaFormatter` just being a single automation of a particular style guide)?

In particular I am thinking similar in spirit to https://github.com/invenia/PkgTemplates.jl (especially https://github.com/invenia/PkgTemplates.jl/pull/61)

nickrobinson251:
Related to the topic of extensability /  JuliaFormatter having a similar philisphy to PkgTemplates, I am wondering "how opinionated should JuliaFormatter be?"

In particular
- Should `JuliaFormatter` have a default style? (probably)
- Should that default just be the (fairly minimal) set specificied in [the Julia base style guide](https://docs.julialang.org/en/v1/manual/style-guide/index.html) and [contributing guidelines](https://github.com/JuliaLang/julia/blob/master/CONTRIBUTING.md)? (I think so)
[...]

PR #89, opened 2019-10-05
Options - Always use `for in`

domluna:
Adds an initial option `always_for_in` which always uses `in` for `for` loops, by default this will be false. I think this is a very reasonable one to have, some people REALLY like always using `in` and I don't want them not using the formatter solely because it doesn't support this.

TODO

- [x] update README
- [x] update GH action script

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,9 +12,30 @@ Width-sensitive formatter for Julia code. Inspired by gofmt and refmt.
 `JuliaFormatter` exports `format_text`, `format_file` and `format`:
 
 ```julia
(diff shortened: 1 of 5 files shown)


nickrobinson251:
Maybe this is a sign that the default isn't quite right for most people?

domluna:
Perhaps. It's somewhat related to #85 but there's been a `=` vs `in` saga #34 #70. It's no secret the default is a little controversial so at least now there will be an, in some cases, much more desired alternative.

Issue #92, opened 2019-10-09
Testing/support for Julia 1.0


odow:
JuliaFormatter isn't tested on Julia 1.0. It doesn't support Julia 1.0 because it requires `splitpath`. Is this intentional? If so, `Project.toml` should be updated with the correct lower bound on the Julia version.

Ref: see this travis log for breakage: https://travis-ci.org/odow/SDDP.jl/jobs/595685471#L413

domluna:
I'm not sure if it's intentional, afaik @StefanKarpinski wrote that part. Stefan any suggestions?

StefanKarpinski:
Either bump the compat entry (see my PR) or stop using `splitpath`. The use of `splitpath` seems pretty simple to replace, so if that's all that's preventing this running on 1.0 it might be worth it.

PR #93, opened 2019-10-10
Project.toml: bump lower bound on Julia compat

StefanKarpinski:
"Fixes" #92. For a real fix, you could not use `splitpath` or copy its definition into here.


odow:
I'd prefer to redefine `splitpath`. I'll make a PR.

PR #96, opened 2019-10-18
Formatting Toggle

### LINKED ISSUE #91: Skipping lines within a file
So we have the `# nofmt` header, but there are a few cases where I want to exclude some lines from the formatter, e.g.,"
```Julia
... code I want formatted ...

# JuliaFormatter: off
data = [ ... very long line I don't want wrapped ... ]
# JuliaFormatter: on

... code I want formatted ...
```

The idea is that the `# JuliaFormatter: XXX` directives are file-local, so we should replace the `# nofmt` header with `# JuliaFormatter: off`, and this would be equivalent to excluding every line in the file.

domluna:
Extends the previous "nofmt" functionality to allow regions of a file not to be formatted. Formatting can now be turned off with the following comment:

```julia
#! format: off
# code from this point will not be formatted
...
```

and back on with:

```julia
#! format: on
# code from this point will be formatted
...
```

Closes #91

### DIFF
--- a/src/JuliaFormatter.jl
+++ b/src/JuliaFormatter.jl
@@ -96,11 +96,11 @@ function Document(text::AbstractString)
 
             # There should not be more than 1 
             # "off" tag on the stack at a time.
-            if t.val == "#! format: off" && length(stack) == 0
+            if occursin(r"^#!\s*format\s*:\s*off\s*$", t.val) && length(stack) == 0
                 push!(stack, t.startpos[1])
             # If "#! format: off" has not been seen
             # "#! format: on" is treated as a normal comment.
-            elseif t.val == "#! format: on" && length(stack) > 0
+            if occursin(r"^#!\s*format\s*:\s*on\s*$", t.val) && length(stack) == 0
                 push!(format_skips, (pop!(stack), t.startpos[1]))
             end
 
--- a/src/JuliaFormatter.jl
+++ b/src/JuliaFormatter.jl
@@ -38,7 +38,7 @@ struct Document
     # a few node types.
     semicolons::Set{Int}
 
-    format_skips::Vector{UnitRange{Int}}
+    format_skips::Vector{Tuple{Int,Int}}
 end
 
 function Document(text::AbstractString)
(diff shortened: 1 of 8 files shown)


src/JuliaFormatter.jl
```diff
@@ -87,6 +93,17 @@ function Document(text::AbstractString)
                 # last comment
                 comments[sl] = (ws, cs)
             end
+
+            # There should not be more than 1 
+            # "off" tag on the stack at a time.
+            if t.val == "#! format: off" && length(stack) == 0
```
StefanKarpinski: Maybe a little more flexibility here would be good? A line matching `^#!\s*format\s*:\s*(on|off)\s*$`? One could potentially reformat the command line itself to be exactly `#! format: off` or `#! format: on` but I worry that it would be too easy to accidentally have trailing spaces or something and then be confused at why it doesn't work.

domluna: makes sense, I'm just not too familiar with regexes and this was super simple :).

I guess `startswith(...)` could work too but then people could write whatever after the comment which could be weird. Might not actually be an issue in practice though.

StefanKarpinski:
```suggestion
            if occursin(r"^#!\s*format\s*:\s*off\s*$", t.val) && length(stack) == 0
```

StefanKarpinski:
```suggestion
            if occursin(r"^#!\s*format\s*:\s*off\s*$", t.val) && length(stack) == 0
```

src/JuliaFormatter.jl
```diff
@@ -87,6 +93,17 @@ function Document(text::AbstractString)
...
+                push!(stack, t.startpos[1])
+            # If "#! format: off" has not been seen
+            # "#! format: on" is treated as a normal comment.
+            elseif t.val == "#! format: on" && length(stack) > 0
```
StefanKarpinski:
```suggestion
            if occursin(r"^#!\s*format\s*:\s*on\s*$", t.val) && length(stack) == 0
```

StefanKarpinski:
There, I wrote the regexs :) But you may still want to normalize the `#! format` lines.

PR #100, opened 2019-10-26
Normalize file line endings

### LINKED ISSUE #98: Line endings on Windows files
On Windows, line endings are given by CRLF (two characters, ascii 13 then ascii 10) whereas on Unix it's simply LF. When operating on Windows files, `format_file` produces a file with mixed line endings, some CRLF, some LF,  which I then need to correct (using Notepad++).

Could the function respect Windows line endings please?

domluna:
Should fix #98 

This also should do https://github.com/domluna/JuliaFormatter.jl/issues/98#issuecomment-546602507

I'm not 100% sure if this is what you meant @StefanKarpinski


StefanKarpinski:
Not quite: I didn't mean to end with a blank line (`\n\n`) but to end with a single trailing newline (`\n`) so that every line, including the last one, ends with an `\n`. Some editors strip the trailing newline on the last line, which is annoying when you do things like in-place editing with `sed`, `awk` or `perl` because those tools by default add the final trailing newline, so if you edit files with them, it spuriously causes files without the final newline to get edited even if nothing has changed.

src/JuliaFormatter.jl
```diff
@@ -268,8 +272,9 @@ function format_file(
         error("$filename must be a Julia (.jl) source file")
     end
     verbose && println("Formatting $filename with indent = $indent, margin = $margin")
-    str = read(filename) |> String
+    str = String(read(filename))
     str = format_text(str, indent = indent, margin = margin, always_for_in = always_for_in)
+    !endswith(str, "\n\n") && (str *= "\n")
```
domluna:
```suggestion
    !endswith(str, "\n") && (str *= "\n")
```

Would that work then?

StefanKarpinski: Yes, that seems right 👍

bin/format.jl
```diff
@@ -81,3 +81,4 @@ if isempty(ARGS) || haskey(opts, :help)
     exit(0)
 end
 format(ARGS; opts...)
+
```
StefanKarpinski: You're going to want to get rid of these trailing newlines too, I'd guess (even though I think they're technically ok).

src/JuliaFormatter.jl
```diff
@@ -8,6 +8,9 @@ export format, format_text, format_file, format_dir
 is_str_or_cmd(x::Tokens.Kind) =
     x in (Tokens.CMD, Tokens.TRIPLE_CMD, Tokens.STRING, Tokens.TRIPLE_STRING)
 
+# on Windows lines can end in "\r\n"
+normalize_line_ending(s::AbstractString) = replace(s, "\r\n" => "\n")
```
StefanKarpinski: You could also do this here:
```jl
replace(s, r"\n*$", "\n")
```
That should replace zero or more trailing newlines at the end of the file with a single newline. Then you can skip the bit at line 277.

domluna: regex sensei

StefanKarpinski: I've spent a _lot_ of time writing regexes 😬

PR #101, opened 2019-10-27
Multi variable `for`/`let`

### LINKED ISSUE #95: Multi-variable `for` and `let`
Currently, we have

```julia
julia> format_text("""
       for a in x,
           b in y,
           c in z
           body
       end""") |> println
for a in x,
b in y,
c in z
    body
end

julia> format_text("""
       let a = x,
           b = y,
           c = z
           body
       end""") |> println
let a = x,
b = y,
c = z
    body
end
```

It's great that these are consistent but this is somewhat unfortunate that scoping of the variables `b` and `c` are not reflected in the indentation level.  Can you perhaps use an empty line as a separator?  Something like

```julia
for a in x,
    b in y,
    c in z

    body
end
```

while

```julia
for a in x
    body
end
```

still is valid?

A nice consequence is that indentation of the variables in

```julia
let aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa = x,
    bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb = y,
    cccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccc = z

    body
end
```

is compatible with the equivalent "immediately invoked function expression" form:

```julia
(function (; aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa,
    bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb,
    cccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccc)
    body
end)(
    aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa = x,
    bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb = y,
    cccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccc = z,
)
```

domluna:
Fixes #95 

At the moment this joins the arguments on 1 line and, if necessary, nests them like it would arguments in a tuple. 

TODO:

- [x] empty newline after argument and before body

### DIFF
--- a/src/nest.jl
+++ b/src/nest.jl
@@ -255,7 +255,7 @@ function n_for!(x, s; extra_width = 0)
     n = x.nodes[idx]
     if n.typ === NOTCODE && n.startline == n.endline
         res = get(s.doc.comments, n.startline, (0, ""))
-        res == (0, "") && deleteat!(x.nodes, idx-1)
+        res == (0, "") && deleteat!(x.nodes, idx - 1)
     end
 end
 
--- a/src/nest.jl
+++ b/src/nest.jl
@@ -96,6 +96,10 @@ function nest!(x::PTree, s::State; extra_width = 0)
         n_tuple!(x, s, extra_width = extra_width)
     elseif x.typ === CSTParser.Block
         n_block!(x, s, extra_width = extra_width)
+    elseif x.typ === CSTParser.For
+        n_for!(x, s, extra_width = extra_width)
+    elseif x.typ === CSTParser.Let
+        n_for!(x, s, extra_width = extra_width)
     elseif x.typ === CSTParser.TypedVcat
         n_call!(x, s, extra_width = extra_width)
     elseif x.typ === CSTParser.StringH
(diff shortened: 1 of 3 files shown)


StefanKarpinski:
Question: does this force multiple let or for "assignments" onto multiple lines regardless of length or only if the "assignment" expressions are long?

domluna:
Only if it’s long

StefanKarpinski:
Perfect. Just checking since some of the test cases are quite short, but I guess they have very small margins.

PR #113, opened 2019-11-05
Format comprehensions

### LINKED ISSUE #112: Balance line breaks across the CST nodes??
Here is an example of something that seems to be formatted poorly:

```julia
    weights = Dict((file, i) => w for (file, subject) in subjects for (
        i,
        w,
    ) in enumerate(weightfn.(eachrow(subject.events))))
```
Manually, I would format this as:

```julia
    weights = Dict((file, i) => w 
        for (file, subject) in subjects 
        for (i, w) in enumerate(weightfn.(eachrow(subject.events))))
```

It is much easier to read the second form (at least for me). Not knowing anything about the internals here, I'm not sure what the specific rule is that I'd like, but it seems like it should be express-able in terms of the depth of the CST at which line breaks occur. The first (actual) result is pretty asymmetric in terms of where along the CST the line break occurs, and it seems like you would want to favor line breaks that balance the nodes of the tree across lines.

domluna:
fixes #112 

Need to add some more test cases.

### DIFF
--- a/src/nest.jl
+++ b/src/nest.jl
@@ -132,8 +132,7 @@ function n_do!(x, s; extra_width = 0)
 end
 
 function n_generator!(x, s; extra_width = 0)
-    @info "" x.indent
-    n_block!(x, s, extra_width = extra_width)
+    n_block!(x, s, extra_width = extra_width, custom_indent = x.indent)
 end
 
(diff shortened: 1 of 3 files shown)


domluna:
alright seems g2g

PR #133, opened 2019-11-24
Imports/Exports consistent indent

### LINKED ISSUE #128: Suggestion: always use multiple-of-`indent` indentation levels?
Currently, JuliaFormatter.jl uses indentation levels that are not multiple of four (or whatever `indent` is) in some cases.  Is there strong philosophy/opinion behind this choice?  If not, why not always restrict the indentation levels to multiple of four?  I think it has some benefits:

* As indentation level becomes less dependent on surrounding code, it produces less diff after editting code.
* Structure of nested data literals is clearer.

OTOH, if I were to guess, the benefit of current approach is that it saves width of the code and makes less line breaks.

### Example: named tuple

Consider that you have LISPy `return`-less function definition:

```julia
f(x) =
    if x
        (
         aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa = 1,
         bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb = 2,
         cccccccccccccccccccccccccccccccccccccccc = 3,
        )
    else
        nothing
    end
```

You may want to change to function with explicit `return` later.  However, it would introduce a large diff because the indentation of the named tuple would be changed:

```julia
function f(x)
    x = something(x, false)
    if x
        return (
            aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa = 1,
            bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb = 2,
            cccccccccccccccccccccccccccccccccccccccc = 3,
        )
    else
        return nothing
    end
end
```

This can be avoided by using multiple-of-four indentation level:

```julia
f(x) =
    if x
        (
            aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa = 1,
            bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb = 2,
            cccccccccccccccccccccccccccccccccccccccc = 3,
        )
    else
        nothing
    end
```

### Example: `using`

Currently, explicit imports with many names produces indentation based on the name of the module:

```julia
using .Some.Internal.Module: aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa,
                             bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb,
                             cccccccccccccccccccccccccccccccccccccccc
```

This also produces a large diff when the module name is changed (e.g., `.Some.Internal.Module` -> `.Internal.Module`).

This can be avoided by using multiple-of-four indentation level:

```julia
using .Some.Internal.Module:
    aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa,
    bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb,
    cccccccccccccccccccccccccccccccccccccccc
```

### Example: Nested data structure

Consider some nested data written using literals:

```julia
[
 (
  aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa = 1,
  bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb = 2,
  cccccccccccccccccccccccccccccccccccccccc = (
      dddddddddddddddddddddddddddddddddddddddd = 3,
      eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee = 4,
  ),
 ),
 (
  ffffffffffffffffffffffffffffffffffffffff,
  (
   gggggggggggggggggggggggggggggggggggggggg,
   hhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhh,
   iiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiii,
  ),
 ),
 jjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjj,
]
```

Personally I find that it is hard to grasp nested level and it is not very satisfactory that nested level of `d...` and `g...` are inconsistent.  With multiple-of-four indentation levels, I think it would make the nested structure clearer:

```julia
[
    (
        aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa = 1,
        bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb = 2,
        cccccccccccccccccccccccccccccccccccccccc = (
            dddddddddddddddddddddddddddddddddddddddd = 3,
            eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee = 4,
        ),
    ),
    (
        ffffffffffffffffffffffffffffffffffffffff,
        (
            gggggggggggggggggggggggggggggggggggggggg,
            hhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhh,
            iiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiii,
        ),
    ),
    jjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjj,
]
```

domluna:
closes #128 I'll have to look at the macro one a little closer


StefanKarpinski:
Nice! This seems much preferable!

PR #164, opened 2020-01-02
Styles

domluna:
This implements custom overriding most (if not all - WIP!) of the formatter via styles. In a nutshell multiple dispatch is used to allow control over every part of the formatter.

## Overview

`FST` is now type parameterized with `AbstractFormatNode`. Some of these types are copied over from `CSTParser.Head` enum and some are custom, added just for formatting purposes. A very nice ergonomic improvement is with this we can keep reusing the same functions with different types instead of having to write `p_if`, `p_for`, etc.

Styles are subtypes of `AbstractStyle`:

```julia
abstract type AbstractStyle end

struct DefaultStyle <: AbstractStyle
  innerstyle::Union{Nothing,AbstractStyle}
end
DefaultStyle() = DefaultStyle(nothing)

getstyle(s::DefaultStyle) = s.innerstyle === nothing ? s : s.innerstyle
```

`innerstyle` allows a custom style to be propagated recursively. There might be a "nicer" way to do this but this works.

### Example: CustomStyle

Suppose we want to implement a style which has no spaces on keyword arguments.

First define the style:

```julia
struct CustomStyle <: AbstractStyle end
getstyle(s::CustomStyle) = s
```

Then write a new `pretty` function for our new custom style and the keyword type:

```julia
function pretty(style::CustomStyle, ::Kw, cst::CSTParser.EXPR, s::State)
    t = FST{Kw}(cst, nspaces(s))
    for a in cst
        add_node!(t, pretty(style, a, s), s, join_lines = true)
    end
    t
end
```

Lastly, we need to fallback to `DefaultStyle` for all other node types:

```julia
pretty(style::CustomStyle, cst::CSTParser.EXPR, s::State; kwargs...) = pretty(style, nodetype(cst), cst, s; kwargs...)
pretty(style::CustomStyle, node::T, cst::CSTParser.EXPR, s::State; kwargs...) where T <: AbstractFormatNode = pretty(DefaultStyle(style), node, cst, s; kwargs...)
```

`nodetype(cst)` converts the `CSTParser.Head` to a type `<: AbstractFormatNode`.

There a more involved example in `styles/yas.jl` implementing YAS style (WIP).

closes https://github.com/domluna/JuliaFormatter.jl/issues/85

### DIFF
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -5,9 +5,9 @@ uuid = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
 
 [[CSTParser]]
 deps = ["Tokenize"]
-git-tree-sha1 = "e8166cc32a3bfedc97eda16db428b1faba29ca8e"
+git-tree-sha1 = "2d85216856653e583207af1662e8b38e628cbec4"
 uuid = "00ebfdb7-1f24-5e51-bd34-a7502290713f"
-version = "2.0.0"
+version = "2.1.0"
 
 [[Distributed]]
 deps = ["Random", "Serialization", "Sockets"]
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -10,6 +10,7 @@ makedocs(
         "Code Style" => "style.md",
         "Skipping Formatting" => "skipping_formatting.md",
         "Syntax Transforms" => "transforms.md",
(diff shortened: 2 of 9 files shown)


domluna:
Going to close this in favour of #171. If the compilation times can be reduced at a later date this will be revisited.

PR #168, opened 2020-01-07
Change behaviour of `overwrite = false`

kdheepak:
- `return str` instead of `write(path * "_fmt" * ext, str)`


kdheepak:
Thoughts on this PR? If you are interested in making this change, then we can change to the command line flag `-o / --overwrite` to `-o / --output` so that the user can provide the file they want the output to go to. And if the command line argument is not provided, it can be the return value of the function.

domluna:
If we give the user the option of where the output should go this looks like a nice change 👍

kdheepak:
I'm closing this since I have work around.

PR #172, opened 2020-01-15
Support shebang

innerlee:
Tested locally and manually.

### DIFF
--- a/src/JuliaFormatter.jl
+++ b/src/JuliaFormatter.jl
@@ -339,7 +339,8 @@ function format_file(
     whitespace_ops_in_indices::Bool = false,
 )
     path, ext = splitext(filename)
-    if ext != ".jl"
+    shebang_pattern = r"^#!\s*/.*\bjulia[0-9.-]*\b"
+    if ext != ".jl" && match(shebang_pattern, readline(filename)) === nothing
         error("$filename must be a Julia (.jl) source file")
     end
     verbose && println("Formatting $filename")


domluna:
I'm not sure what this does. Can you elaborate a little more on why this is needed? Thanks.

innerlee:
reference: https://en.wikipedia.org/wiki/Shebang_(Unix)

Basically it is a julia script using as a executable file.
For example, I can write a julia script to print the first line of a file.

`printfirstline.jl`:
```julia
println(readline(ARGS[1]))
```

Then you need to `julia printfirstline.jl my.file` to run the script.
To make life easier, one can also rewrite the file as

`printfirstline`:
```julia
#!/usr/bin/env julia
println(readline(ARGS[1]))
```
and make the file executable by `chmod +x printfirstline`.
Then you can type `./printfirstline my.file` to do the same job.

StefanKarpinski:
Looking at the code, I still don’t follow this change. Am aware of shebag lines. Can you explain a bit more what this change does?

innerlee:
It `ext != ".jl"`, then it reads the first line of the file, and match it with the [regex](https://github.com/julia-vscode/julia-vscode/pull/891/files) to check if it is a julia script. And if nothing matches, it throws an error.

innerlee:
Any more info needed to add? Or the code style?

StefanKarpinski:
I see. This _broadens_ the criteria for whether something is considered a Julia file to include the case where it does not have the `.jl` extension, but it has a shebang line, so e.g. a file called `bin/script` that doesn't have the `.jl` extension but does start with `#!julia` will get formatted.

innerlee:
Right! Because I wrote lots of julia scripts. And today I tried this [julia-formatter](https://marketplace.visualstudio.com/items?itemName=singularitti.vscode-julia-formatter) vscode extension, and it errors when trying to use it.

domluna:
Makes sense to me, as long as it doesn't break existing functionality. I'll have to take this for a spin.

Issue #174, opened 2020-01-20
Long expressions with binary operators in return or macrocall


tkf:
Here is an output of JuliaFormatter 0.2.4:

```julia
function _()
    return some_expression *
    some_expression *
    some_expression *
    some_expression *
    some_expression *
    some_expression *
    some_expression
end

@some_macro some_expression *
some_expression *
some_expression *
some_expression *
some_expression *
some_expression *
some_expression

function _()
    return some_expression |> some_expression |> some_expression |> some_expression |>
           some_expression |> some_expression |> some_expression
end

@some_macro some_expression && some_expression && some_expression &&
            some_expression && some_expression && some_expression && some_expression
```

I think there are a few problems:

* There is no indentation with binary operator `*`.
* Indentations for `|>` depend on the macro names. Repeating my argument in #128, I think it makes sense to use multiple-of-`indent` indentations. It would result in a smaller diff when (say) a macro is renamed.
* Indentations of `*` and `|>` are inconsistent.

Combining these points, I think it'd be nice if the output is something like

```julia
function _()
    return some_expression *
        some_expression *
        some_expression *
        some_expression *
        some_expression *
        some_expression *
        some_expression
end

@some_macro some_expression *
    some_expression *
    some_expression *
    some_expression *
    some_expression *
    some_expression *
    some_expression

function _()
    return some_expression |>
        some_expression |>
        some_expression |>
        some_expression |>
        some_expression |>
        some_expression |>
        some_expression
end

@some_macro some_expression |>
    some_expression |>
    some_expression |>
    some_expression |>
    some_expression |>
    some_expression |>
    some_expression
```

domluna:
This is due to parsing. All three of these have different precedence and hence are parsed differently. Now in the case of a formatter it's arguable this doesn't matter and so all 3 of these cases should be normalized s.t the output is consistent.

That being said the formatting for chainopcall is this case should be better.

`*` example:

```julia
FunctionDef  175(175)
 FUNCTION  9(8)
 Call  8(3)
  _  1(1)
  (
  )
 Block  155(154)
  Return  155(154)
   RETURN  7(6)
   ChainOpCall  148(147)
    some_expression  16(15)
    OP: STAR  6(1)
    some_expression  16(15)
    OP: STAR  6(1)
    some_expression  16(15)
    OP: STAR  6(1)
    some_expression  16(15)
    OP: STAR  6(1)
    some_expression  16(15)
    OP: STAR  6(1)
[...]

PR #199, opened 2020-03-09
Consistent indent for Binary and Chain operations

domluna:
This PR does 2 things:

1. BinaryOpCalls and ChainOpCalls now have a consistent indent. For example prior to this PR these would be format outputs.

```julia
julia> format_text(s, margin=1) |> print
return a *
       b
julia> format_text(s0, margin=1) |> print
return a *
b *
c
```

This PR

```julia
julia> format_text(s, margin=1) |> print
return a *
       b
julia> format_text(s0, margin=1) |> print
return a *
       b *
       c
```

This makes progress on #174 but it's doesn't use an indent multiple. In order to make the multiple of indent work well for these cases I think we need to surround with expressions with parenthesis. But, nonetheless I think this is a better outcome than before. Also, this PR makes that follow-up work much easier (if worthwhile).

2. For most binary operations CSTParser creates a ChainOpCall node if 2+ operations of the same kind are used consecutively. There are some exceptions such as, `&&`, `||`, `|>`, etc. This is due to precedence but leads to weird formatting, and required some hacky un-nesting to avoid results like this.

```julia
if argument1 &&
   argument2 && argument3 && argument4
    ...
end
```

This PR has a transform for these kind of binary op structures so that above exceeded the margin it would be formatted to:

```julia
if argument1 &&
   argument2 &&
   argument3 &&
   argument4
    ...
end
```

### DIFF
--- a/src/JuliaFormatter.jl
+++ b/src/JuliaFormatter.jl
@@ -286,6 +286,7 @@ function format_text(
     s = State(Document(text), indent, margin, opts)
     t = pretty(style, x, s)
     hascomment(s.doc, t.endline) && (add_node!(t, InlineComment(t.endline), s))
+    flatten_fst!(t)
     nest!(style, t, s)
 
     s.line_offset = 0
@@ -404,7 +405,7 @@ if VERSION < v"1.1.0"
             dir, base = _splitdir_nodrive(p)
             dir == p && (pushfirst!(out, dir); break)  # Reached root node.
             if !isempty(base)  # Skip trailing '/' in basename
-                pushfirst!(out, base)
+                pushfirst!(out, base)common
             end
             p = dir
         end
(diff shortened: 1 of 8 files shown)


domluna:
@StefanKarpinski @tkf what do you guys think about this? The only thing I'm a little pensive about is short circuit uses of `&&` and `||` but I think the output of this is still better even in those cases, i.e.  https://github.com/domluna/JuliaFormatter.jl/pull/199/files#diff-9915f66306b42c958fcd076dc0eee523L1234-L1236.

tkf:
Thanks! I played with this branch a little bit and it works nicely so far.

By "pensive" you mean `cst[3].typ !== CSTParser.BracesCat` in

https://github.com/domluna/JuliaFormatter.jl/blob/03eeef81d1cdece9df8102f11b3dc293a07b556a/src/pretty.jl#L1241-L1244

is not in a separate line? I think it's an improvement in the sense that you'll have a smaller diff next time you improve it more.

StefanKarpinski:
Looks good to me!

PR #220, opened 2020-04-09
doc: add YAS style doc to index

aviatesk:
I also would like to point out here that the current stable document still points to v0.3's one.


domluna:
oops

ty! 👏

domluna:
do I have to tag a new release for this to show ?

aviatesk:
hm not sure, Documenter is supposed to update stable documentation after a new one tagged. I'm not familiar with GitHub actions that you're using for documentation generation, but maybe there is something lacked in its setup ?

domluna:
it shows on dev, I think have to tag a new release show it shows on stable docs

PR #221, opened 2020-04-09
configuration file implementation

### LINKED ISSUE #184: discussion: configuration file
What do you think about a configuration file, that specifies formatting options (and possibly custom styles), like [`.prettierrc` for Prettier](https://prettier.io/docs/en/configuration.html) ?

For remote formatting task, we already can use GitHub actions and thus the same configurations is used within a project.
But AFAIU, in order to share those configurations for local formatting task, we currently don't have any good way other than each project member specifies them by themselves.
If we have a configuration file, (that may or not locate at the root of project directory) JuliaFormatter.jl itself or certain frontends like Juno can respect the configuration if exist, and then all the project member can easily share the same configuration even for local formatting.

aviatesk:
closes #184 

TODOs (help very appreciated):
- [x] write documentation
- [x] add test
- [x] search _up_ the file tree ?

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -17,4 +17,4 @@ makedocs(
     ],
 )
 
-deploydocs(repo = "github.com/domluna/JuliaFormatter.jl.git")
+deploydocs(repo = "github.com/domluna/JuliaFormatter.jl.git", push_preview = true)
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -2,7 +2,7 @@ using Documenter, JuliaFormatter
 
 makedocs(
     sitename = "JuliaFormatter",
-    format = Documenter.HTML(prettyurls = true),
+    format = Documenter.HTML(prettyurls = get(ENV, "CI", nothing) == "true"),
     modules = [JuliaFormatter],
     pages = [
         "Introduction" => "index.md",
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -12,6 +12,7 @@ makedocs(
         "Syntax Transforms" => "transforms.md",
         "Custom Styles" => "custom_styles.md",
         "YAS Style" => "yas_style.md",
+        "Configuration File" => "config.md",
         "API Reference" => "api.md",
     ],
 )
--- a/docs/src/config.md
+++ b/docs/src/config.md
@@ -12,7 +12,7 @@ When found, the configurations in the file will overwrite the given options.
     [Juno](https://junolab.org/), a Julia IDE that offers formatting feature using this package, also respects
     configuration file.
     When you use `Julia-Client: Format-Code` command, Juno will automatically search for a configuration file with the
-    same rule as `format` does from the directory fo the current editor.
+    same rule as `format` does from the directory of current editor.
 
 
 ## Basic Configuration
@@ -25,9 +25,9 @@ margin = 100
 ```
 then files under `somedir` will be formatted with 2 spaces indentation and the maximum line length 100.
 
-!!! note
+!!! warning "Custom Style"
     Currently the configuration file doesn't support [Custom Styles](@ref).
-    For the time being, we only provide [YAS Style](@ref) support for the configuration file.
+    For the time being, we only provide [YAS Style](@ref) spec in configuration file.
     In order to use YAS style instead of the default style, you can just specify:
     > .JuliaFormatter.toml
     ```toml
@@ -39,7 +39,7 @@ then files under `somedir` will be formatted with 2 spaces indentation and the m
 
 ## Search Rule
 
-`.JuliaFormatter.toml` will be searched _up_ from the directory of the file begin formatted.
+`.JuliaFormatter.toml` will be searched _up_ from the directory of the file being formatted.
 So if you have:
 ```
 dir
--- a/docs/src/config.md
+++ b/docs/src/config.md
@@ -28,14 +28,13 @@ then files under `somedir` will be formatted with 2 spaces indentation and the m
 !!! note
     Currently the configuration file doesn't support [Custom Styles](@ref).
     For the time being, we only provide [YAS Style](@ref) support for the configuration file.
-    In order to use YAS style, you can just specify:
+    In order to use YAS style instead of the default style, you can just specify:
     > .JuliaFormatter.toml
     ```toml
     ...
     style = "yas"
     ...
     ```
-    Any other value will fallback to the default style.
 
 
 ## Search Rule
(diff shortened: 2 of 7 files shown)


aviatesk:
A possible complain on this approach might be that we can't easily pass custom styles via TOML. One workaround would be:
```toml
style = "path_where_style_defined"
```
but it's fairly complicated. Not sure it's worth to support.

aviatesk:
alright, so now the documentation is the only thing left to do. At this point I would like to hear your opinions on this approach. I think [the testcases](https://github.com/domluna/JuliaFormatter.jl/pull/221/files#diff-bd8f87318916b9ff607e83398ac5961f) are enough descriptive about the current behavior.

/cc @StefanKarpinski

domluna:
> A possible complain on this approach might be that we can't easily pass custom styles via TOML. One workaround would be:
> 
> ```toml
> style = "path_where_style_defined.jl"
> ```
> 
> but it's fairly complicated. Not sure it's worth to support.

yeah it's going to be tricky to support styles JuliaFormatter doesn't already know about from a config file.

src/JuliaFormatter.jl
```diff
@@ -539,4 +570,26 @@ function format(paths; options...)
...
+
+function kwargs(dict)
+    ns = (Symbol.(keys(dict))...,)
+    vs = (collect(values(dict))...,)
+    return pairs(NamedTuple{ns}(vs))
+end
+
+parse_config(tomlfile) = kwargs(parsefile(tomlfile))
+
+overwrote_options(options, config) = kwargs(merge(options, config))
```
domluna, line 593:
```suggestion
overwrite_options(options, config) = kwargs(merge(options, config))
```

aviatesk:
> So right now styles won't work in the config file is that correct?

no. 

In order to support that, we can do something like:
change
https://github.com/domluna/JuliaFormatter.jl/blob/562daea2638c52623d5fd2ca39b3e2211c4206f8/src/JuliaFormatter.jl#L564
to 
```julia
if (stylefile = get(opts, :stylefile, nothing)) !== nothing
    stylefile_path = joinpath(config_dir, stylefile)
    style = include(stylefile_path)
    opts_dict = Dict(opts)
    delete!(opts_dict, :stylefile)
    push!(opts_dict, :style => style)
    opts = kwargs(opts_dict)
    Base.invokelatest(format_text, full_path; opts...)
else
    format_file(full_path; opts...)
end
```

and then an user can have `.JuliaFormatter.toml` and `custom_style.jl` such like
> `.JuliaFormatter.toml`
```toml
...
stylefile = "custom_style.jl"
...
```
> `custom_style.jl`
```julia
using JuliaFormatter, CSTParser
using JuliaFormatter: AbstractStyle, FST, State, add_node!
import JuliaFormatter: pretty, p_kw

struct CustomStyle <: AbstractStyle end

# this must be defined
getstyle(s::CustomStyle) = s

function p_kw(style::CustomStyle, cst::CSTParser.EXPR, s::State)
    t = FST(cst, 0)
    for a in cst
        add_node!(t, pretty(style, a, s), s, join_lines = true)
    end
    t
end

CustomStyle() # custom style object must be returned
```

***

Fairly complicated  ¯\_(ツ)_/¯

aviatesk:
I guess you missed to configure CI for the test ?

domluna:
for some reason it's stopped showing up

domluna:
ok using github actions for testing now, you'll  need to rebase

aviatesk:
nice, seems to work ;)

aviatesk:
So maybe this approach is acceptable, basically ? Then I will write a documentation.

domluna:
I think it's great, we can worry about adding styles later. For the time being we can support default and YAS style

example toml file

```toml
...
style = yas
```

and then in the code call `YASStyle` if it's present.

aviatesk:
yeah, that would be enough for a basic usage case (I actually had done [the exactly same thing](https://github.com/JunoLab/Atom.jl/pull/317/files#diff-b5120071dc5dfc2ab1de1a9e0ceb2187R19-R21) in Juno :P)

Will let you know when I think this is ready.

aviatesk:
ok, I've done the documentation work. Any feedback on this would be very appreciated.

***

Well, while writing documentation I found another thing I want to consider; maybe we want to search _up_ for a config file from the directory of the file begin formatted, [as prettier does](https://prettier.io/docs/en/configuration.html).
Currently this PR looks for config files searching _down_ the file tree.
But when we think about applications outside of `format`, like formatting feature in Juno, following the search rule can be confusing.
For example, let's think about a case like:
```
dir
├─ .JuliaFormatter.toml
├─ code.jl
├─ ...
└─ subdir
   ├─ ...
   └─ sub_code.jl
```
As far as we call `format("dir")`, the formatting rule will be applied as we expect.
But for example in Juno, when we format on `sub_code.jl`, the current search rule doesn't find `dir/.JuliaFormatter.toml` since it's called from the directory of `sub_code.jl`, while we usually expect the configurations in `dir/.JuliaFormatter.toml` to be applied in such a setup.

So I would like to change this PR so that we search up the file tree until a config file is (or isn't) found. Any ideas ?

docs/make.jl
```diff
@@ -17,4 +17,4 @@ makedocs(
     ],
 )
 
-deploydocs(repo = "github.com/domluna/JuliaFormatter.jl.git")
+deploydocs(repo = "github.com/domluna/JuliaFormatter.jl.git", push_preview = true)
```
aviatesk, line 20: um guessing from the error, maybe this change needs to be done with the write access to the repository ?
@domluna this would allow us to preview the documentation of this package without local build, so I would like to just let you know this, but I will erase this from the commit history later ;) (ref: https://discourse.julialang.org/t/ann-documenter-0-24/31381)

domluna: makes sense :+1:

domluna:
Searching up the directory makes sense IMO. I've personally run into this issue in other projects/languages where opening the file from a certain directory causes certain things to work or not work and it's really frustrating. It does make things more complicated but it might be worth it. 

@StefanKarpinski does this make sense for larger projects in particular?

aviatesk:
I've done something [similar to that in Juno](https://github.com/JunoLab/Atom.jl/blob/9004b35ad1bc8c899677e553a499cd99411feba4/src/misc.jl#L40-L51) before, and the implementation won't be so compilicated as far as I can tell (if you're okay with the simple logic).

domluna:
> I've done something [similar to that in Juno](https://github.com/JunoLab/Atom.jl/blob/9004b35ad1bc8c899677e553a499cd99411feba4/src/misc.jl#L40-L51) before, and the implementation won't be so compilicated as far as I can tell (if you're okay with the simple logic).

no that's looks great :smile: The complication I was thinking about wasn't code related but behavior related and thinking more about it my concern makes no sense lol, carry on :+1:

docs/src/config.md
```diff
@@ -0,0 +1,60 @@
+# Configuration File
+
+From v0.4.2, JuliaFormatter offers [`.prettierrc` style](https://prettier.io/docs/en/configuration.html) configuration file support.
```
domluna, line 3: i recently tagged v0.4.2 so this should be v0.4.3

aviatesk: thanks for pointing this out ;)

aviatesk:
yey, I've done everything, I think. I found the logic is now rather so much simplified :)

aviatesk:
ah yeah, there should be a test for that (I totally forgot about that !)
Will add the testcase and update documentation as well accordingly later.

test/config.jl
```diff
@@ -0,0 +1,114 @@
...
+        open(io -> write(io, before), sub_code_path, "w")
+        open(io -> write(io, before), subsub_code_path, "w")
+
+        format(sub_code_path)
+        @test read(sub_code_path, String) == after2
+        format(subsub_code_path)
+        @test read(subsub_code_path, String) == after2
+    finally
+        rm(sandbox_dir; recursive = true)
+    end
```
aviatesk, line 53: This case will test a case when a configuration file is in higher position of the file tree.

aviatesk:
I just reviewed the documentation and fixed some typos. I think we can merge this as is for now.

Issue #222, opened 2020-04-09
Format long one-line boolean checks to if statements


sethaxen:
It would be useful to have something like `short_to_long_function_def` for boolean checks. Currently statements like this

```julia
function foo()
    my_really_really_really_long_boolean_check() && return my_really_really_really_long_function_name()
end
```

get formatted to this, so the result is separated from the conditional:

```julia
function foo()
    my_really_really_really_long_boolean_check() &&
    return my_really_really_really_long_function_name()
end
```

But I would like an option for it to be formatted to this:

```julia
function foo()
    if my_really_really_really_long_boolean_check()
        return my_really_really_really_long_function_name()
    end
end
```

Likewise this:

```julia
function foo()
    my_really_really_really_long_boolean_check() || return my_really_really_really_long_function_name()
    return 1
end
```

should be formatted to this:

```julia
function foo()
    if !my_really_really_really_long_boolean_check()
        return my_really_really_really_long_function_name()
    end
    return 1
end
```

domluna:
This looks cool.

Indenting has also been brought up in this case

so

```
function foo()
    my_really_really_really_long_boolean_check() && return my_really_really_really_long_function_name()
end
```

would be

```
function foo()
    my_really_really_really_long_boolean_check() &&
        return my_really_really_really_long_function_name()
end
```

? for this other approach would be should only the last arg be indented or all args after the first arg

sethaxen:
Good question, I think all args that are part of the same statement would be indented. @mateuszbaran, do you have any thoughts?

mateuszbaran:
I also think all parts should be indented, like here:
```julia
function foo()
    my_first_really_really_long_boolean_check() &&
        my_second_really_really_long_boolean_check() &&
        return my_really_really_really_long_function_name()
end
```

penelopeysm:
`short_circuit_to_if` exists now:

```julia
julia> s = """
       function foo()
           my_really_really_really_long_boolean_check() && return my_really_really_really_long_function_name()
           return 1
       end"""
"function foo()\n    my_really_really_really_long_boolean_check() && return my_really_really_really_long_function_name()\n    return 1\nend"

julia> format_text(s; short_circuit_to_if=true) |> println
function foo()
    if my_really_really_really_long_boolean_check()
        return my_really_really_really_long_function_name()
    end
    return 1
end
```

PR #224, opened 2020-04-17
Indent standalone lazy and/or expressions after first arg

domluna:
I'm defining this as a lazy and/or expression that's NOT a component of a larger expression.

ref #222 

For demonstration purposes the maximum margin is set to 1.

Examples:

```julia
a && b

->

a &&
    b
```

```julia
a && b && c

->

a &&
    b &&
    c
```

the testset has more examples.

Something I'm still pondering is whether to treat `||` and `&&` as the same operation in this case.

So instead of `a || b && c` being

```
a ||
    b &&
        c
```

it would be

```
a ||
    b &&
    c
```

but this may be too pessimistic, since with a larger margin, which is what it would be in practice,
the former would be

 ```
a ||
    b && c
```

### DIFF
--- a/docs/src/style.md
+++ b/docs/src/style.md
@@ -199,7 +199,7 @@ e1 :
 e2
 ```
 
-If nesting is required for a `=` binary operation, the RHS is placed on the following line.
+If nesting is required for a `=` binary operation, the RHS is placed on the following line and indented.
 
 ```julia
 foo() = body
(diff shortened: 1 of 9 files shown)


StefanKarpinski:
> Something I'm still pondering is whether to treat `||` and `&&` as the same operation in this case.

Maybe distinguish by precedence level? Which would argue against treating them the same.

domluna:
it already 

> Maybe distinguish by precedence level? Which would argue against treating them the same.

That's currently what it does. It seems to only perhaps cause a worse format in really odd cases and there is another way to make it cleaner rather than mess with the precedence.

Issue #216, opened 2020-04-06
autoformatting of docstrings including embedded code


bramtayl:
Formatting the code in the doctests contained in docstrings would be a nice feature to have, I think.

domluna:
sounds similar to https://github.com/domluna/JuliaFormatter.jl/issues/197 but also sounds a little different. I don't think I'm on the same page, can you give an example?

bramtayl:
Similar, this for doc strings not comments though. https://bramtayl.github.io/LightQuery.jl/latest/ might be a good example. The gray bits are julia code. The source is here: https://github.com/bramtayl/LightQuery.jl/blob/master/docs/src/index.md

domluna:
If you format the text in the code blocks it should just work. But that's definitely trickier than it sounds.

bramtayl:
I tried that but I think that after the `julia> ` prompt there needs to be an indent of 6 spaces for the following lines?

StefanKarpinski:
This is definitely harder than it sounds since I think it's equivalent to parsing docstrings and reformatting them. That said, it would be great to have autoformatting of docstrings, including embedded code, which could recursively call the Julia formatter.

domluna:
this would be significantly easier to do in a .jl file and a .md file

bramtayl:
I'm not sure I understand... The source code of the docs I linked is a .md file, with many of the docstrings taken from various .jl files. I'm guessing the hard part would be autoformatting markdown, and once that's done, it could be broadly applicable?

bramtayl:
https://github.com/sumnow/markdown-formatter

bramtayl:
I could take a stab of porting the javascript to julia?

StefanKarpinski:
Porting a JavaScript markdown parser seems like the wrong way to go about this. We already have a Markdown parser built into Julia. Why not just build on that to emit formatted markdown?

bramtayl:
D'oh yes that makes sense

bramtayl:
Maybe then the best way to go about this would be to make improvements (if they are required) to the show methods for markdown in Base?

StefanKarpinski:
Perhaps, but ideally the methods for rendering back to markdown would be added externally. It's not type piracy if the mime type is defined externally. I guess it would also be acceptable to add a method of rendering markdown back to markdown in the internal markdown module.

bramtayl:
[...]

PR #225, opened 2020-04-17
WIP: Format docstrings

bramtayl:
First stab at #216

Doesn't seem to actually change output...

### DIFF
--- a/src/pretty.jl
+++ b/src/pretty.jl
@@ -347,11 +347,55 @@ end
 p_stringh(style::S, cst::CSTParser.EXPR, s::State) where {S<:AbstractStyle} =
     p_stringh(DefaultStyle(style), cst, s)
 
+
+format_docstrings!(something_else) = nothing
+
+function format_docstrings!(document::MD)
+    for clause in document.content
+        format_docstrings!(clause)
+    end
+    nothing
+end
+
+function format_docstrings!(doctest::Code)
+    language = doctest.language
+    code = doctest.code
+    doctest.code = if startswith(language, "@example") ||
+        startswith(language, "@repl") ||
+        startswith(language, "@eval")
+        format_text(code)
+    elseif startswith(language, "jldoctest")
+        if occursin(r"^julia> "m, code)
+            join(
+                (
+                    string("julia> ", join(
+                        (
+                            line for line in split(format_text(input), '\n')
+                        ),
+                        "\n       "
+                    )) for (input, output) in repl_splitter(code)
+                ),
+                "\n\n"
+            )
+        else
+            an_input, output = split(code, "# output\n", limit = 2)
+            string(format_text(String(an_input)),  "# output\n", output)
+        end
+    else
+        code
+    end
+    nothing
+end
+
 # MacroCall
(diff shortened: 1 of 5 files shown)


src/pretty.jl
```diff
@@ -347,11 +347,55 @@ end
...
+end
+
 # MacroCall
 function p_macrocall(ds::DefaultStyle, cst::CSTParser.EXPR, s::State)
     style = getstyle(ds)
     t = FST(cst, nspaces(s))
     if cst[1].typ === CSTParser.GlobalRefDoc
+        # format docstrings
+        docstring = Markdown.parse(cst[2].val)
+        format_docstrings!(docstring)
```
domluna, line 397: I'll need to check this out but I think this should be called on `.val` returned from p_literal and p_stringh below.

bramtayl: I think doing this right would require passing the parser style and state into the inner call to `format_text`. I tried doing that but got lost

domluna: I checked out the branch and I know what's going on now. So as I mentioned before for strings we don't use `.val` for the CST node. Instead we pull the string directly from the file (this pre-processing is done earlier when the `Document` is initially created).

So I think this should be moved into p_literal and use a keyword arg to label it's from a docstring and extra processing should be done on that string.

https://github.com/domluna/JuliaFormatter.jl/blob/e6152d993b2d723c70246be0e9c8ea050d258da3/src/pretty.jl#L257

this string is wrapped in quotes so it doesn't work right out of the box. We need to extract the inner part and then requote it. 

I can fiddle with this a bit and make a commit.

domluna: It also looks like just parsing in Markdown and then converting back to a string (without formatting) doesn't guarantee to return the original input, so that's a bit problematic.

domluna: I'm not sure how to push to this branch directly but https://github.com/domluna/JuliaFormatter.jl/commit/b83b5309917e76945e7fa94e47cf5385446e46a4 expands on this.

Right now it looks like sometimes format_docstrings eats code

```jldoctest
julia> a = 10
10
```

10 isn't printed

It also looks like Markdown.parse does some string escaping/unescaping 
```

Comment (bramtayl) [reply src/pretty.jl:397]: > 10 isn't printed

Oops sorry that's a bug. I forgot to add the output back in, but you can add an extra `'\n', output` to the end of the string call

> It also looks like just parsing in Markdown and then converting back to a string (without formatting) doesn't guarantee to return the original input

Yeah I thought about that. Could be a feature because you are reformatting the markdown too

File: src/pretty.jl [hunk 2]
```diff
@@ -347,11 +347,55 @@ end
...
+        format_text(code)
+    elseif startswith(language, "jldoctest")
+        if occursin(r"^julia> "m, code)
+            join(
+                (
+                    string("julia> ", join(
+                        (
+                            line for line in split(format_text(input), '\n')
+                        ),
+                        "\n       "
```
domluna, line 375: btw is the spacing here for `julia>` ?

bramtayl: Yeah. All of the code after `julia> ` should be indented by 6 spaces, I think.

PR #236, opened 2020-05-19
re-fixes #234

aviatesk:
refixes #234 , will supercede #235 :

- fix typo(`relpath` to `realpath`), and everything should work well
- ensure we `cd` back to the original directory after test

### DIFF
--- a/src/JuliaFormatter.jl
+++ b/src/JuliaFormatter.jl
@@ -360,22 +360,16 @@ function format(paths; options...)
             format_file(path; opts...)
         else
             for (root, dirs, files) in walkdir(path)
-                base_opts = if (config = find_config_file(path)) !== nothing
-                    overwrite_options(options, config)
-                else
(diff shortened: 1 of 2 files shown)


src/JuliaFormatter.jl
```diff
@@ -360,22 +360,16 @@ function format(paths; options...)
...
-                    options
-                end
-
                 for file in files
                     _, ext = splitext(file)
                     ext == ".jl" || continue
                     full_path = joinpath(root, file)
                     ".git" in splitpath(full_path) && continue
-                    dir = relpath(root)
+                    dir = realpath(root)
```
aviatesk, line 368: the issue was caused by this typo ....
now the config will be detected correctly even if given as relative paths, so we don't need `base_ops` stuff here.

domluna:
nice I'll merge and tag a new release

PR #240, opened 2020-05-26
Make format_file and format return true if file(s) already formatted

tom--lee:
Makes the functions `format` and `format_file` return a boolean: `true` if the files were already formatted and `false` if not.  This simplifies the process of checking whether files are formatted during CI. E.g. `julia -e 'using JuliaFormatter; format(".") || exit(1)'`

Previously those functions return `nothing` so this should not break most code.

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -64,7 +64,7 @@ format(
 )
 ```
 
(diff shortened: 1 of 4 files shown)


domluna:
👍

This may break some existing github actions so I'll have to verify that.

PR #247, opened 2020-06-09
format docstrings, second round same as the first

bramtayl:
Rebased


bramtayl:
The bug was just deleting necessary initial whitespace so that should be fixed

bramtayl:
Whack-a-bug

test/default_style.jl
```diff
@@ -4836,4 +4844,119 @@ some_function(
...
+        formatted = """
+        \"""
+        ```
+        format_text(
+        )
+        ```
+        \"""
+        function format_text() end"""
+
+        @test format_text(unformatted) == formatted
```
domluna, line 4945: What's the rationale for reformatting like this?

domluna:
@bramtayl 

So I think i fixed the bug - https://github.com/domluna/JuliaFormatter.jl/commit/c6150285b1ca4adca74ba470d0170121b7cc3135

Also I've been playing around with CommonMark, it looks promising but I'm not sure it's suitable as of yet.

If this is merged it'll be marked as experimental since it might eat some parts of docstrings, as shown in the test that had to be labelled as broken.

test/default_style.jl
bramtayl: I don't know. `Markdown.plain` does it though

bramtayl:
Yay!!! I'm looking forward to very pretty docstrings

bramtayl:
Close in favor of #254

PR #254, opened 2020-06-29
Rebase #247

bramtayl:


test/files/setup_databases.jl
```diff
@@ -35,24 +37,32 @@ function parse_commandline()
         "--output", "-o"
         help = "path where the database folder is going to be created."
         default = "."
-        "--pdb"
+        """
+–pdb
+"""
```
domluna, line 42: this formatting seems quite a bit worse

src/JuliaFormatter.jl
```diff
@@ -37,24 +41,25 @@ include("styles/yas.jl")
...
+    whitespace_typedefs::Bool = false,
+    whitespace_ops_in_indices::Bool = false,
+    remove_extra_newlines::Bool = false,
+    import_to_using::Bool = false,
+    pipe_to_function_call::Bool = false,
+    short_to_long_function_def::Bool = false,
+    always_use_return::Bool = false,
+    whitespace_in_kwargs::Bool = true,
+)::String
+```
```
domluna, line 60: When you look at the docstring via help in the REPL or docs generated from Documenter is the output identical to how this formatted previously?

domluna: Also I don't think the code in these blocks is formatted, would that be possible?

test/files/setup_databases.jl
bramtayl: Oof yeah that is worse. This isn't even a docstring is it?

domluna: It's parsed that way by CSTParser - I'm not sure if that's correct or not but we want to leave these strings alone, I think even doing the quote normalization thing isn't ideal in this case.

src/JuliaFormatter.jl
bramtayl: This PR only formats code as julia if it is marked as `@example`, `@repl`, `@eval`, `jldoctest`. I could change it to assume that any code without a language listed is julia?

test/files/setup_databases.jl
bramtayl: Where does the code come from? I'm not sure what the `@add_arg_table` macro is doing. I tried parsing it in Julia and it does parse as a docstring...If they are intended to be markdown docstrings, `--pdb` should be wrapped in backticks

bramtayl:
I tried out CommonMark too. This is what I came up with:

```
struct FormatRule{Style, State}
   style::Style
   state::State
end

block_modifier(rule::FormatRule) = Rule(1) do parser, block
    if block.t isa CodeBlock
        style = rule.style
        state = rule.state
        language = block.t.info
        code = block.literal
        if startswith(language, "@example") ||
            startswith(language, "@repl") ||
            startswith(language, "@eval")
            block.literal = format_text(style, state, code)
        elseif startswith(language, "jldoctest")
            block.literal = if occursin(r"^julia> "m, code)
                join(
                    (
                        string(
                            "julia> ",
                            join(
                                (line for line in split(format_text(style, state, input), '\n')),
                                "\n       ",
                            ),
                            '\n',
                            output,
                        ) for (input, output) in repl_splitter(code)
                    ),
                    "\n\n",
                )
            else
                an_input, output = split(code, r"\n+# output\n+", limit = 2)
                string(format_text(style, state, String(an_input)), "\n# output\n", output)
            end
        end
    end
end

function format_docstrings(style, state, text)
    boundaries = findall(text) do character
        character != '"'
    end
    string(
        "\"\"\"\n",
        sprint(
            markdown,
            enable!(Parser(), FormatRule(style, state))(text[boundaries[1]:boundaries[end]]),
        ),
        "\"\"\"",
    )
end
```

It seems to work ok on regular markdown. There's something funny happening to the code blocks though (new lines disappearing?)

domluna:
Removing newlines inside code blocks seems to be something CommonMark does by default. The other thing I encountered when using it is \ is added to snake case, so `foo_bar` becomes `foo\_bar` - this is only relevant to code not in blocks that perhaps should be in a block.

bramtayl:
Given CommonMark's goal is explicitly to be round-trippable, and that doesn't seem to be a priority in Base, I think it's probably the way to go, at least in the long run. I can open an issue about the missing newlines?

domluna:
It may be customizable, there's a lot of things that are on by default but
you can disable them with disable!(parser, ...) or maybe override it
similar to how we're overriding what's done with julia code blocks.

StefanKarpinski:
We'd like to switch the Base parser to the CommonMark one at some point, so using that seems like getting ahead of the ball in a good way.

bramtayl:
Ref https://github.com/MichaelHatherly/CommonMark.jl/issues/5

MichaelHatherly:
> you can disable them with disable!(parser, ...) 

Messing about with that is probably worth trying, mostly to see what it'll break. I've not tried to disable all but code blocks, so haven't got a clue what that might do... :boom: but it may work fine, open some issues if you do run into anything.

Btw,

> ```
> sprint(markdown, ...)
> ```

can be written as `markdown(...)` if the only arg is the parsed ast. It'll give you a string back since it calls `sprint` itself.

bramtayl:
Ok, I've updated the PR to use commonmark. I'm going to try rebasing (yet again...sigh...) because CommonMark seems a bit more conservative about messing with markdown and I want to undo the formatting changes.

PR #256, opened 2020-07-02
Style similar to DocumentFormat.jl

domluna:
At the moment this is a derivative of `YASStyle` but with nesting only valid for binary and conditional calls (at least for now).

UPDATE:

Despite the draft state this is functional right now. With the line offset info for tokens in the source file provided by #282 this could be revised to be even more similar to the document format style.


davidanthoff:
Cool!

I have to admit I don't fully understand this PR :) But I think the main thing from my end is that the "default" formatting in VS Code should leave line breaks alone, i.e. it should just never insert or remove or change any existing line breaks. Is that something that this PR here does?

domluna:
I definitely have to update this but the core idea is that this will line breaks alone. There are exceptions to this - for example if you have something like `try body catch e end` it'll get formatted to

```
try
    body
catch e
end
```

Indentation is kind of a toss up but since this PR was opened we've added functionality to track line offsets in the original source so we could do something like heuristics there. Or we can pass off the indentation to another style.

davidanthoff:
> There are exceptions to this

Would it be complicated to have no exceptions at all? I kind of like the idea that our base formatter in VS Code just never changes line breaks, and I think that is what DocumentFormat does.

> Indentation is kind of a toss up

I think we do change indentation in DocumentFormat, and so I think fixing indentation stuff could certainly be part of the default formatting story?

Also CCing @ZacLN and @pfitzseb on this discussion.

domluna:
> Would it be complicated to have no exceptions at all? I kind of like the idea that our base formatter in VS Code just never changes line breaks, and I think that is what DocumentFormat does.

I'd have to look into it more but I don't think it would be too problematic - mainly just need to overwrite more functions

> I think we do change indentation in DocumentFormat, and so I think fixing indentation stuff could certainly be part of the default formatting story?

What I meant by "toss up" is that I don't know if DocumentFormat has a enforced style of indentation. For example YASStyle
enforces arguments of a function definition to be placed after '(` where as DefaultStyle (if over the margin) would enforce arguments to be placed x spaces after `function`. For a block like

```
begin
a
              b
   c
end
```

it would be indented to whatever the number of spaces is set to in the options (defaults to 4)

davidanthoff:
Bump, can this be merged?

src/JuliaFormatter.jl
```diff
@@ -5,7 +5,7 @@ using Tokenize
 using DataStructures
 using Pkg.TOML: parsefile
 
-export format, format_text, format_file, DefaultStyle, YASStyle
+export format, format_text, format_file, DefaultStyle, YASStyle, DocumentFormatStyle
```
davidanthoff, line 8: Maybe we would just call it `VSCodeStyle`?

domluna:
> Bump, can this be merged?

I'll need to update it but that should be fairly easy

src/JuliaFormatter.jl
domluna: I'm fine with that, although maybe `MinimalStyle` would be more informative.

Also there remains an open question of how the arguments of something like a function call should be formatted. Right now this is a derivative of YASStyle, which might be ok but some might prefer the indentation style of Default/Blue styles instead.

It should be possible to dynamically choose which style to do based on the positioning of the arguments but I haven't thought about it too much.

domluna: Or would be indent of 4 style be preferred  (`DefaultStyle`) and then just stick with that ? assuming the argument appears on a newline since this wouldn't mess with line breaks

davidanthoff: Yes, `MinimalStyle` is much better, I like it!

On the indentation style for function arguments, not really sure :) Is there an easy way for me to try out the options? We could also just merge this for now and then iterate on that particular question a bit?

pfitzseb:
Bump :)

Can I help out here?

domluna:
https://github.com/domluna/JuliaFormatter.jl/compare/dl/423 is similar to this for YASStyle #423

domluna:
I think the indentation approach is the only thing still up in the air though

pfitzseb:
Imho the default JuliaFormatter behavior is fine. DocumentFormatter currently messes up indentation, so we should probably have the JuliaFormatter default fix that :P

davidanthoff:
Yeah, I should say I don't think we need to strive to mimic DocumentFormat in all details (even when it was not broken). I just really feel very strongly that by default we should not add or remove any line breaks, so I think as long as that is sorted out in this new style, we're probably good from my end :) Caveat is that I would like to try JuliaFormatter.jl with this new style on some codebases to see what it actually does, I've actually never run it on anything :)

pfitzseb:
Yeah, we should just run it on all of the julia-vscode codebase.

domluna:
sounds like default style indentation would work fine then

davidanthoff:
What is the best path forward here? Merge this, tag a release, but consider the `MinimalStyle` experimental for now until we have tested it a bit more and iterated on it if needed?

domluna:
> What is the best path forward here? Merge this, tag a release, but consider the `MinimalStyle` experimental for now until we have tested it a bit more and iterated on it if needed?

@pfitzseb have you done anything on this yet. If not I can revive this but use default indentation instead of YAS

pfitzseb:
Sorry, haven't had time to work on this. https://github.com/domluna/JuliaFormatter.jl/compare/dl/423 for `DefaultStyle` should work nicely though, I think (+ a really big margin).

domluna:
sounds about right I'll update this something this week

pfitzseb:
Did you get a chance to look into this? If not I'll try to figure the necessary changes out on Monday so we can get https://github.com/julia-vscode/julia-vscode/pull/2334 into the next minor release.

domluna:
I'll work on this today, I'll rope you in the new PR

PR #265, opened 2020-07-26
Resolve #262

### LINKED ISSUE #262: @ is removed from macrocalls with an interpolation
In the output (`:($(__MODULE__).@macro)`) of the first example, `@` in front of `__MODULE__` is missing.  It seems this happens only with macro call:

```julia
julia> raw":($(@__MODULE__).@macro)" |> format_text |> println
:($(__MODULE__).@macro)

julia> raw":($(@__MODULE__).property)" |> format_text |> println
:($(@__MODULE__).property)

julia> raw":($(@__MODULE__))" |> format_text |> println
:($(@__MODULE__))

(tmpjfx9am02.julia_tmp) pkg> st -m
Status `/tmp/tmpjfx9am02.julia_tmp/Manifest.toml`
  [00ebfdb7] CSTParser v2.5.0
  [864edb3b] DataStructures v0.17.19
  [98e50ef6] JuliaFormatter v0.6.7
  [bac558e1] OrderedCollections v1.3.0
  [0796e94c] Tokenize v0.5.8
  [2a0f44e3] Base64
  [ade2ca70] Dates
  [8ba89e20] Distributed
  [b77e0a4c] InteractiveUtils
  [76f85450] LibGit2
  [8f399da3] Libdl
  [56ddb016] Logging
  [d6f4376e] Markdown
  [44cfe95a] Pkg
  [de0858da] Printf
  [3fa0cd96] REPL
  [9a3f8284] Random
  [ea8e919c] SHA
  [9e88b42a] Serialization
  [6462fe0b] Sockets
  [8dfed614] Test
  [cf7118a7] UUIDs
  [4ec0a83e] Unicode
```

domluna:
During the pass to move @ to the end of the macro call, the FST walk was
descending into inner macro calls. This is problematic since all leaf
nodes with the value "@" were removed. This is solved by modifying `walk` to stop descending into a sub-tree if value other than `nothing` is returned from the passed function.

### DIFF
--- a/test/default_style.jl
+++ b/test/default_style.jl
@@ -4827,6 +4827,6 @@ some_function(
 
         str_ = raw":($(@__MODULE__.macro).@field.macro)"
         str = raw":($(__MODULE__.@macro).field.@macro)"
-        @test fmt(str) == str
+        @test fmt(str_) == str
     end
 end
--- a/test/default_style.jl
+++ b/test/default_style.jl
@@ -4810,4 +4810,23 @@ some_function(
         @test fmt(str_, 2, length(str_) - 1) == str
         @test fmt(str, 2, length(str_)) == str_
     end
+
+    @testset "issue 262 - removal of @ in nested macrocall" begin
+        str = raw":($(@__MODULE__).@macro)"
+        @test fmt(str) == str
+
+        str = raw":($(@__MODULE__).property)"
+        @test fmt(str) == str
+
+        str = raw":($(@__MODULE__))"
+        @test fmt(str) == str
+
+        str_ = raw":($(@__MODULE__).@field.macro)"
+        str = raw":($(@__MODULE__).field.@macro)"
+        @test fmt(str_) == str
+
+        str_ = raw":($(@__MODULE__.macro).@field.macro)"
+        str = raw":($(__MODULE__.@macro).field.@macro)"
+        @test fmt(str) == str
+    end
 end
--- a/test/default_style.jl
+++ b/test/default_style.jl
@@ -4821,12 +4821,36 @@ some_function(
         str = raw":($(@__MODULE__))"
         @test fmt(str) == str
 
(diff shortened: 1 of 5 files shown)


test/default_style.jl
```diff
@@ -4810,4 +4810,23 @@ some_function(
...
+
+        str = raw":($(@__MODULE__).property)"
+        @test fmt(str) == str
+
+        str = raw":($(@__MODULE__))"
+        @test fmt(str) == str
+
+        str_ = raw":($(@__MODULE__).@field.macro)"
+        str = raw":($(@__MODULE__).field.@macro)"
+        @test fmt(str_) == str
```
tkf, line 4826: Huh, interestingly, `a.@b.c` is not the same as `@a.b.c` and `a.b.@c`.

```julia
julia> :(@a.b.c)
:(#= REPL[1]:1 =# a.b.@c)

julia> :(a.@b.c)
:(#= REPL[2]:1 =# a.b.var"@c")

julia> :(a.b.@c)
:(#= REPL[3]:1 =# a.b.@c)
```

Although it looks like `a.@b.c` is a syntax (or rather lowering) error:

```julia
julia> Meta.@lower a.@b.c
ERROR: LoadError: syntax: invalid syntax "a.b.@c" around REPL[8]:1
Stacktrace:
 [1] top-level scope
   @ REPL[8]:1
 [2] lower(m::Module, x::Any)
   @ Base.Meta ./meta.jl:124
 [3] top-level scope
   @ REPL[8]:1
in expression starting at REPL[8]:1
```

Not sure what should happen though.  Maybe the safest option is to leave it as-is?

But, if it's easy to transform this way (to avoid special casing or something), personally I think it's OK to transform to the "correct" macro invocation syntax.

domluna: odd corner case but I've seem to have gotten around it.

PR #282, opened 2020-09-04
Custom Alignment

### LINKED ISSUE #179: detecting manually aligned code
One of my only hangups about using JuliaFormatter universally at this point is that there is occasionally code where I want to manually line things up and autoformatting can mess with this. Here's an example I just ran JuliaFormatter on (the rest of the results look great!):
```diff
 function SuffixArray(s::S) where {S<:AbstractString}
     n = length(s)
-    index = zeros(n <= typemax(Int8)  ? Int8  :
-                  n <= typemax(Int16) ? Int16 :
-                  n <= typemax(Int32) ? Int32 : Int64, n)
+    index = zeros(
+        n <= typemax(Int8) ? Int8 :
+            n <= typemax(Int16) ? Int16 : n <= typemax(Int32) ? Int32 : Int64,
+        n,
+    )
     return SuffixArray(s, n, index)
 end
```
The manually aligned before version is much more legible to me. In pondering what to do about this, I have a few trains of thought that I just want to put out there from least to most fancy:

1. Turn formatting off for blocks like this using `#! format: off` and `#! format: on`.
2. Detect that adjacent lines have an unusual amount of syntactic similarity and automatically turn formatting off.
3. Figure out how to align things like this automatically so that they line up on adjacent lines.

We can already do 1, so that's good. I had considered 3 before but it just seems too smart to actually be implemented reliably. On the other hand 2 seems pretty doable with some clever heuristics. There might even be a 2+ variation that detects syntactic similarity modulo whitespace differences and can even adjust whitespace so that things line up better. My guess is that when things line up it's generally the punctuation that lines up while names, literals and whitespace vary.

domluna:
This closes #179

- [x] struct fields
- [x] consts
- [x] conditionals
- [x] short function definitions
- [x] general assignments
- [x] documentation

Extra

- [ ] matrices
- [x] `=>`

The main feature of this PR is custom alignment, which is explained below. Another change of this PR is conditionals and chain operator calls are only unnested if the entire expression fits on the line.

In other words the following no longer occurs

```julia
var = cond ?
    e1 : e2
```

This change was made in part to reduce the complexity of aligning conditionals but also because it's not a particularly good use case of unnesting and generally makes the code harder to follow.

# Custom Alignment

Custom alignment is determined by a whitespace heuristic:

A token (typically an operator, i.e. `=, ?, ::, etc`) is custom aligned if there are 
`> 1` whitespaces from the previous expression since the formatter only outputs 
0 or 1 whitespaces for separation. If custom alignment is determined then all
expressions in the code block will be aligned to the furthest aligned token.

> NOTE: alignment overrides nesting behavior, meaning it ignores the allowed maximum margin

### Example

Suppose the source text is as follows

```julia
const variable1 = 1
const var2      = 2
const var3 = 3
const var4 = 4
const var5          = 5
```

If the `align_assignment` option is enabled the formatter will detect that `var2`
is aligned to `variable1` AND `var2` has several whitespaces (>1) prior to
`=`. Since `var3`,`var4`, and `var5` are part of the same code block (no comments
or newlines separating code) they will also be aligned.

So the output would be

```julia
const variable1 = 1
const var2      = 2
const var3      = 3
const var4      = 4
const var5      = 5
```

Notice how the `=` operator for `var5` is correctly positioned
despite it being located further to the right than other `=` operators.

However, if the source code is

```julia
const variable1 = 1
const variable2 = 2
const var3 = 3
const var4 = 4
const var5 = 5
```

It's now ambiguous whether this is meant to be aligned and so the formatter will
proceed with normal behavior.

## Alignment Options

In order for alignment to occur the option must be set to `true`. Available options:

- `align_assignment`
- `align_struct_field`
- `align_conditional`
- `align_pair_arrow`

> **Caveat: Since nesting is disabled when alignment occurs be careful when adding comments to the RHS expression. This will be fixed in a future release**

For example:

```julia
const variable1 = 1
const var2      = foo(10,
    # comment,
    20)
```

This will be formatted to

```julia
const variable1 = 1
const var2      = foo(10, # comment, 20)
```

which causes a parsing error.

### `align_assignment`

Align to `=`-like operators. This covers variable assignments and short definition functions.

```julia
const UTF8PROC_STABLE    = (1 << 1)
const UTF8PROC_COMPAT    = (1 << 2)
const UTF8PROC_COMPOSE   = (1 << 3)
const UTF8PROC_DECOMPOSE = (1 << 4)
const UTF8PROC_IGNORE    = (1 << 5)
const UTF8PROC_REJECTNA  = (1 << 6)
const UTF8PROC_NLF2LS    = (1 << 7)
const UTF8PROC_NLF2PS    = (1 << 8)
const UTF8PROC_NLF2LF    = (UTF8PROC_NLF2LS | UTF8PROC_NLF2PS)
const UTF8PROC_STRIPCC   = (1 << 9)
const UTF8PROC_CASEFOLD  = (1 << 10)
const UTF8PROC_CHARBOUND = (1 << 11)
const UTF8PROC_LUMP      = (1 << 12)
const UTF8PROC_STRIP     = (1 << 13)


vcat(X::T...) where {T}         = T[X[i] for i = 1:length(X)]
vcat(X::T...) where {T<:Number} = T[X[i] for i = 1:length(X)]
hcat(X::T...) where {T}         = T[X[j] for i = 1:1, j = 1:length(X)]
hcat(X::T...) where {T<:Number} = T[X[j] for i = 1:1, j = 1:length(X)]

a  = 1
bc = 2

long_variable = 1
other_var     = 2
```

### `align_struct_field`

Align struct field definitions to `::` or `=` - whichever has higher precedence.

```julia
Base.@kwdef struct Options
    indent_size::Int                       = 4
    max_margin::Int                        = 92
    always_for_in::Bool                    = false
    whitespace_typedefs::Bool              = false
    whitespace_ops_in_indices::Bool        = false
    remove_extra_newlines::Bool            = false
    import_to_using::Bool                  = false
    pipe_to_function_call::Bool            = false
    short_to_long_function_def::Bool       = false
    always_use_return::Bool                = false
    whitespace_in_kwargs::Bool             = true
    annotate_untyped_fields_with_any::Bool = true
    format_docstrings::Bool                = false
    align_struct_fields::Bool              = false

    # no custom whitespace so this block is not aligned
    another_field1::BlahBlahBlah = 10
    field2::Foo = 10

    # no custom whitespace but single line blocks are not aligned
    # either way
    Options() = new()
end


mutable struct Foo
    a             :: T
    longfieldname :: T
end
```

### `align_conditional`

Align conditional expressions to either `?`, `:`, or both.

```julia
# This will remain like this if using YASStyle
index = zeros(n <= typemax(Int8)  ? Int8  :
              n <= typemax(Int16) ? Int16 :
              n <= typemax(Int32) ? Int32 : Int64, n)

# Using DefaultStyle
index = zeros(
    n <= typemax(Int8)  ? Int8  :
    n <= typemax(Int16) ? Int16 :
    n <= typemax(Int32) ? Int32 : Int64,
    n,
)

# Note even if the maximum margin is set to 1, the alignment remains intact
index = 
    zeros(
        n <= typemax(Int8)  ? Int8  :
        n <= typemax(Int16) ? Int16 :
        n <= typemax(Int32) ? Int32 : Int64,
        n,
    )

```

### `align_pair_arrow`

Align pair arrows (`=>`).

```julia
pages = [
    "Introduction"        => "index.md",
    "How It Works"        => "how_it_works.md",
    "Code Style"          => "style.md",
    "Skipping Formatting" => "skipping_formatting.md",
    "Syntax Transforms"   => "transforms.md",
    "Custom Alignment"    => "custom_alignment.md",
    "Custom Styles"       => "custom_styles.md",
    "YAS Style"           => "yas_style.md",
    "Configuration File"  => "config.md",
    "API Reference"       => "api.md",
]
```

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -10,6 +10,7 @@ makedocs(
         "Code Style" => "style.md",
         "Skipping Formatting" => "skipping_formatting.md",
         "Syntax Transforms" => "transforms.md",
+        "Custom Alignment" => "custom_alignment.md",
         "Custom Styles" => "custom_styles.md",
         "YAS Style" => "yas_style.md",
         "Configuration File" => "config.md",
--- a/docs/src/custom_alignment.md
+++ b/docs/src/custom_alignment.md
@@ -139,7 +139,7 @@ end
 Align conditional expressions to either `?`, `:`, or both.
 
 ```julia
-# This will remain like if using YASStyle
+# This will remain like this if using YASStyle
 index = zeros(n <= typemax(Int8)  ? Int8  :
               n <= typemax(Int16) ? Int16 :
               n <= typemax(Int32) ? Int32 : Int64, n)
@@ -153,7 +153,6 @@ index = zeros(
 )
 
 # Note even if the maximum margin is set to 1, the alignment remains intact
-
 index = 
     zeros(
         n <= typemax(Int8)  ? Int8  :
--- a/docs/src/custom_alignment.md
+++ b/docs/src/custom_alignment.md
@@ -56,17 +56,38 @@ proceed with normal behavior.
 
 ## Alignment Options
 
+
 In order for alignment to occur the option must be set to `true`. Available options:
 
 - `align_assignment`
 - `align_struct_field`
 - `align_conditional`
+- `align_pair_arrow`
+
+> **Caveat: Since nesting is disabled when alignment occurs be careful when adding comments to the RHS expression. This will be fixed in a future release**
+
+For example:
+
+```julia
+const variable1 = 1
+const var2      = foo(10,
+    # comment,
+    20)
+```
+
+This will be formatted to
+
+```julia
+const variable1 = 1
+const var2      = foo(10, # comment, 20)
+```
+
+which causes a parsing error.
 
 ### `align_assignment`
 
 Align to `=`-like operators. This covers variable assignments and short definition functions.
 
-> **Caveat: since nesting is disabled when alignment occurs be careful when adding comments to the RHS expression.**
 
 ```julia
 const UTF8PROC_STABLE    = (1 << 1)
@@ -162,3 +183,23 @@ index =
     )
 
 ```
+
+### `align_assignment`
+
+Align pair arrows (`=>`).
+
+```julia
+pages = [
+    "Introduction"        => "index.md",
+    "How It Works"        => "how_it_works.md",
+    "Code Style"          => "style.md",
+    "Skipping Formatting" => "skipping_formatting.md",
+    "Syntax Transforms"   => "transforms.md",
+    "Custom Alignment"    => "custom_alignment.md",
+    "Custom Styles"       => "custom_styles.md",
+    "YAS Style"           => "yas_style.md",
+    "Configuration File"  => "config.md",
+    "API Reference"       => "api.md",
+]
+```
+
--- a/docs/src/custom_alignment.md
+++ b/docs/src/custom_alignment.md
@@ -184,7 +184,7 @@ index =
 
 ```
 
-### `align_assignment`
+### `align_pair_arrow`
 
 Align pair arrows (`=>`).
 
--- a/docs/src/custom_alignment.md
+++ b/docs/src/custom_alignment.md
@@ -1,0 +1,165 @@
+# Custom Alignment
+
+> Solution for [issue 179](https://github.com/domluna/JuliaFormatter.jl/issues/179)
+
+Custom alignment is determined by a whitespace heuristic:
+
+A token (typically an operator, i.e. `=, ?, ::, etc`) is custom aligned if there are 
+`> 1` whitespaces from the previous expression since the formatter only outputs 
+0 or 1 whitespaces for separation. If custom alignment is determined then all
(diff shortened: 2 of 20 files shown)


StefanKarpinski:
Very clever and cool!

tkf:
Very clever and cool indeed!

Is there a reason to special-case `const` and not supporting `global` and even "unqualified" assignments?  Naively, the heuristics mentioned in the OP sound applicable to a case like

```julia
variable1 = 1
var2      = 2
var3 = 3
var4 = 4
var5          = 5
```

domluna:
> Very clever and cool indeed!
> 
> Is there a reason to special-case `const` and not supporting `global` and even "unqualified" assignments? Naively, the heuristics mentioned in the OP sound applicable to a case like
> 
> ```julia
> variable1 = 1
> var2      = 2
> var3 = 3
> var4 = 4
> var5          = 5
> ```

Nope, just that const was a simpler test case but the same approach would work for any LHS = RHS situation

aviatesk:
looking forward to this ! I'm glad to be an
end test user :)

Issue #283, opened 2020-09-05
Implement BlueStyle


nickrobinson251:
This issue is to organise and track efforts to implement [BlueStyle](https://github.com/invenia/BlueStyle) as a [Custom Style](https://domluna.github.io/JuliaFormatter.jl/dev/custom_styles/). Similar to how [YASGuide](https://github.com/jrevels/YASGuide) is now available in JuliaFormatter via [YASStyle](https://domluna.github.io/JuliaFormatter.jl/dev/yas_style/). (see issue https://github.com/domluna/JuliaFormatter.jl/issues/198 and PRs https://github.com/domluna/JuliaFormatter.jl/pull/214, https://github.com/domluna/JuliaFormatter.jl/pull/217). 

I think the first step is to come up with a list of cases where BlueStyle recommends a style change that is currently not available via DefaultStyle or YASStyle...

TO DOs (see comment below for details/examples):
- Module Imports
    - [ ] Imports at top of file (see also #284)
    - [ ] One import per line
    - [ ] Order imports alphabetically (see also #257)
    - [ ] Prefer `using` over `import`
-   ~Global Variables (not possible?)~
-   ~Function Naming (not possible)~
- Method Definitions
     - [x] Only use short-form function definitions when they fit on a single line
     - [x] **When using long-form functions always use the return keyword (but not in `do` blocks)** (second highest priority) (#295) 
     - [ ] use `return nothing` rather than bare `return` 
- Keyword Arguments
    - [x] No whitespace in keyword arguments
    - [x]  **Separate keyword arguments from positional arguments with a `;` semicolon** (third highest priority) (#295) 
- Whitespace
    - [x] Avoid extraneous whitespace immediately inside brackets
    - [x] Avoid extraneous whitespace immediately before a comma or semicolon
    - [x] Avoid whitespace around `:` in ranges, add brackets to clarify
    - [x] Avoid whitespace for alignment
    - [x] Surround most binary operators with a single space 
        - [x] except `^`
         - [x] except `//` (https://github.com/domluna/JuliaFormatter.jl/pull/288)
    - [x] Avoid using whitespace between unary operands and the expression
    - [x] Avoid extraneous empty lines
    - [x] Function calls which cannot fit on a single line within the line limit should be broken up such that... (see [below](https://github.com/domluna/JuliaFormatter.jl/issues/283#issuecomment-687620179))
[...]

PR #288, opened 2020-09-12
Give `//` the same binaryop whitespace exception as `^`

nickrobinson251:
As mentioned in #283, BlueStyle recommends exempting both `^` and `//` from the rule that binary operators are surrounded by whitespace. `a ^ b` is already formatted as `a^b`. This PR adds the analogous formatting `x // y` -> `x//y`.

Since `x//y` is a single `Rational` number, i think this makes sense as a general rule, so i've added it to the DefaultStyle - does that seem reasonable?

This is my first attempt at making contributions to the code in the package, so please let me know if this approach seems right 😄 Hopefully tests show the intended behaviour

### DIFF
--- a/src/styles/default/pretty.jl
+++ b/src/styles/default/pretty.jl
@@ -1364,14 +1364,21 @@ function p_binaryopcall(
     elseif op.kind === Tokens.EX_OR
         add_node!(t, Whitespace(1), s)
         add_node!(t, pretty(style, op, s), s, join_lines = true)
-    elseif (is_number(cst[1]) || op.kind === Tokens.CIRCUMFLEX_ACCENT) && op.dot
+    elseif (is_number(cst[1]) || op.kind === Tokens.FWDFWD_SLASH || op.kind === Tokens.CIRCUMFLEX_ACCENT) && op.dot
         add_node!(t, Whitespace(1), s)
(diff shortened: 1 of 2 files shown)


src/styles/default/pretty.jl
```diff
@@ -1364,14 +1364,21 @@ function p_binaryopcall(
...
-        (CSTParser.precedence(op) in (8, 13, 14, 16) && op.kind !== Tokens.ANON_FUNC)
-    ) && op.kind !== Tokens.IN
+    elseif op.kind !== Tokens.IN && (
+        nospace || (
+            op.kind !== Tokens.ANON_FUNC && CSTParser.precedence(op) in (
+                CSTParser.ColonOp,
+                CSTParser.RationalOp,
+                CSTParser.PowerOp,
+                CSTParser.DeclarationOp,
+                CSTParser.DotOp,
```
nickrobinson251, line 1378: I changed `(8, 13, 14, 16)` -> `(CSTParser.ColonOp, CSTParser.PowerOp, CSTParser.DeclarationOp, CSTParser.DotOp)` since it took me quite a while to work out what these mystery numbers were. I think the additional verbosity is worth it.

domluna (approved):
makes sense, thanks for adding it.

PR #300, opened 2020-09-23
Allow `Style`s to set defaults for all settings

### LINKED ISSUE #296: Have defaults settings depend on the Style (dispatch on Style)
I'm wondering about how to give styles ability to fully specific the format, while still exposing the full flexibility to users who want to tweak the style to their own preferences.

Right now, there's one `format` method and one of the keywords is `style` e.g. `style=YASStyle`, and then other keywords e.g. `always_for_in=true` controlling other aspects of the formatting.

We get the situation where to format code following a Style Guide (e.g. YAS Guide) you need to both set a `Style` (e.g. `YASStyle`) _and_ set the keywords to the ones that fulfil the style (e.g. pass `always_for_in=true`).

One small issue with this, is it leads to a slightly awkward API (in my opinion), where passing in a `Style` is not sufficient to have the code formatted in line with the corresponding style guide (since you need to set keywords too). 

A "work-around" for this is to [use a config file](https://domluna.github.io/JuliaFormatter.jl/dev/yas_style/#Configuration-File-Example-1). (Config files are a very useful feature, in my opinion, but for the use-case where you want to exactly follow a style guide, i think we may be able to get an even nicer API without requiring config files, which are still useful for other use-cases.)

A second small issue, since `style` is a keyword / there's one `format` method, all the other `keyword` are the same no matter the `style` (i.e. same keywords `always_for_in`, same defaults `always_for_in=false`), which means we cannot have style-specific keywords (unclear yet if that's flexibility we'd want)

**A suggestion for how to improve this situation is to (i) dispatch on `Style`, and maybe (ii) have `Style` constructors own the keywords**

---
e.g. with option (i)
```julia
function format(path, style::YasStyle; always_for_in=true, etc...)
    # code
end

function format(path, style::DefaultStyle; always_for_in=false, etc...)
    # code
end

format(path; style=DefaultStyle(), kwargs...) = format(path, style; kwargs...)   # (thanks Eric for suggesting this below)
```
then call like
```julia
# to get "full" YAS style
format("path/to/file", YASStyle())

# to "customise" YAS e.g.
format("path/to/file", YASStyle(); always_for_in=false)
```
---
e.g. with both option (i) and option (ii) as well
```julia
function format(path, style::YasStyle)
    # code
end

function format(path, style::DefaultStyle)
    # code
end

format(path; kwargs...) = format(path, DefaultStyle(kwargs...))  # maybe
```
then call like
```julia
# to get "full" YAS style
format("path/to/file", YASStyle())

# to "customise" YAS e.g.
format("path/to/file", YASStyle(always_for_in=false))

# possible to have a YAS specific option e.g.
format("path/to/file", YASStyle(always_for_in=false, use_pipes_to_annoy_alex=true))
```

FWIW, while most of the benefit comes from doing (i), doing both options (i) and (ii) makes most sense to me (and while it's a breaking change to do (ii), it does also allow us to keep the one-arg `format(path; kwargs...)` method if we want, by passing kwargs to `DefaultStyle`).

(This issue is kinda related to #212) 

---
As an aside: I think it's still a little undecided which settings are controlled by `Styles` e.g. `YASStyle`, versus via keywords e.g. `always_for_in=true`, but "what gets it's own keyword?" seems a different question to the one i'm raising here, which is about `Style`s being able to fully specify the format (without losing current flexibility)

nickrobinson251:
- fixes #296 (implements "option (i)" suggested there)

Behaviour:
- Change the results for running `format(str, style=style)` when `style` is BlueStyle or YASStyle, to use the configurable settings that those styles prefer (so tests are affected).
- This doesn't change the API.  
- Updated docs 

Implementation:
- basically this just adds a `format_text(str, ::Style; kwargs...)` method, where the values for the `kwargs` are different depending on the `Style` -- most of the relevant changes are those in `src/JuliaFormatter.jl`.
- This is implemented via having a `options(::Style)` function that returns a NamedTuple which contains the defaults for the style, and and any `kwargs` passed in is `merge`d into that to overwrite them.  
- To be able to re-use more existing code, i changed the field names in the `Options` type (to match the public-facing keywords, so that we can just splat them into the constructor)... which caused big diffs (sorry)

These changes affect the tests, so i fixed them up:
- For YAS tests, since i don't know how to write in YAS style, and there are lots of tests, for those i just changed the internal `yasfmt` function to use `DefaultStyle` settings (i.e. have the same behaviour it had before this PR). 
- ~For Blue tests, since I know how to write BlueStyle, which the tests are short, i just changed the tests to be written in BlueStyle (except for the format change we're testing).~ Same thing for Blue tests as YAS tests.
- ~i've not added dedicated tests, given this is change is exercised by the existing YAS and Blue tests. Would you like me to do that? Not sure how best to do that tbh...~

### DIFF
--- a/src/JuliaFormatter.jl
+++ b/src/JuliaFormatter.jl
@@ -26,8 +26,10 @@ abstract type AbstractStyle end
 """
     DefaultStyle
 
-The default formatting style. See the style section of the documentation
+The default formatting style. See the [Style](@ref) section of the documentation
 for more details.
+
+See also: [`BlueStyle`](@ref), [`YASStyle`](@ref)
 """
 struct DefaultStyle <: AbstractStyle
     innerstyle::Union{Nothing,AbstractStyle}
@@ -446,7 +448,7 @@ See [`format_file`](@ref) and [`format_text`](@ref) for a description of the opt
 This function will look for `.JuliaFormatter.toml` in the location of the file being
 formatted, and searching *up* the file tree until a config file is (or isn't) found.
 When found, the configurations in the file will overwrite the given `options`.
-See ["Configuration File"](@id) for more details.
+See [Configuration File](@ref) for more details.
 
 ### Output
 
--- a/src/JuliaFormatter.jl
+++ b/src/JuliaFormatter.jl
@@ -396,7 +396,7 @@ end
     format_file(filename::AbstractString, style::AbstractStyle; kwargs...)::Bool
 """
 function format_file(filename::AbstractString, style::AbstractStyle; kwargs...)
-    return format_file(filename; style=style, kwargs...)
+    return format_file(filename; style = style, kwargs...)
 end
 
 if VERSION < v"1.1.0"
@@ -517,7 +517,7 @@ format(path::AbstractString; options...) = format((path,); options...)
 """
     format(path, style::AbstractStyle; options...)::Bool
 """
-format(path, style::AbstractStyle; options...) = format(path; style=style, options...)
+format(path, style::AbstractStyle; options...) = format(path; style = style, options...)
 
 function kwargs(dict)
     ns = (Symbol.(keys(dict))...,)
--- a/src/JuliaFormatter.jl
+++ b/src/JuliaFormatter.jl
@@ -392,6 +392,13 @@ function format_file(
     return formatted_str == str
 end
 
+"""
+    format_file(filename::AbstractString, style::AbstractStyle; kwargs...)::Bool
+"""
+function format_file(filename::AbstractString, style::AbstractStyle; kwargs...)
+    return format_file(filename; style=style, kwargs...)
+end
+
 if VERSION < v"1.1.0"
     # We define `splitpath` here, copying the definition from base/path.jl
     # because it was only added in Julia 1.1.
@@ -506,6 +513,11 @@ function format(paths; options...)::Bool
     return already_formatted
 end
 format(path::AbstractString; options...) = format((path,); options...)
+
+"""
+    format(path, style::AbstractStyle; options...)::Bool
+"""
+format(path, style::AbstractStyle; options...) = format(path; style=style, options...)
 
 function kwargs(dict)
     ns = (Symbol.(keys(dict))...,)
--- a/src/JuliaFormatter.jl
+++ b/src/JuliaFormatter.jl
@@ -33,6 +33,7 @@ end
 DefaultStyle() = DefaultStyle(nothing)
 
 @inline getstyle(s::DefaultStyle) = s.innerstyle === nothing ? s : s.innerstyle
+@inline options(s::DefaultStyle) = NamedTuple()
 
 include("document.jl")
 include("options.jl")
@@ -59,9 +60,9 @@ normalize_line_ending(s::AbstractString) = replace(s, "\r\n" => "\n")
 """
     format_text(
         text::AbstractString;
+        style::AbstractStyle = DefaultStyle(),
         indent::Int = 4,
         margin::Int = 92,
-        style::AbstractStyle = DefaultStyle(),
         always_for_in::Bool = false,
         whitespace_typedefs::Bool = false,
         whitespace_ops_in_indices::Bool = false,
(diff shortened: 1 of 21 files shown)


src/JuliaFormatter.jl
```diff
@@ -242,47 +243,13 @@ Markdown is formatted with [`CommonMark`](https://github.com/MichaelHatherly/Com
...
-        always_use_return = always_use_return,
-        whitespace_in_kwargs = whitespace_in_kwargs,
-        annotate_untyped_fields_with_any = annotate_untyped_fields_with_any,
-        format_docstrings = format_docstrings,
-        align_struct_field = align_struct_field,
-        align_conditional = align_conditional,
-        align_assignment = align_assignment,
-        align_pair_arrow = align_pair_arrow,
-    )
+    opts = Options(; merge(options(style), kwargs)...)
```
nickrobinson251, line 252: This is most of the actual change (plus the `options` methods for each `Style`)

test/runtests.jl
```diff
@@ -33,16 +33,21 @@ function run_nest(text::String; opts = Options(), style = DefaultStyle())
...
-    run_nest(text, opts = Options(max_margin = max_margin))
+run_nest(text::String, margin::Int) =
+    run_nest(text, opts = Options(margin = margin))
 
-yasfmt1(s, i, m; kwargs...) = fmt1(s; kwargs..., i = i, m = m, style = YASStyle())
-yasfmt(s, i, m; kwargs...) = fmt(s; kwargs..., i = i, m = m, style = YASStyle())
+function yasfmt(s, i, m; kwargs...)
+    # use DefaultStyle options to make writing tests easier
+    kws = merge(options(DefaultStyle()), kwargs)
+    return fmt(s; kws..., i = i, m = m, style = YASStyle())
```
nickrobinson251, line 42: oh i think this isn't actually overwriting YAS options in the way i intended, and that's why tests fail for YAS -- will fix

domluna:
This is great just 1 question

The default options are overwritten by any kwargs right, so that works the same way it currently does?

domluna (commented):
I don't think you need to add additional tests. 

For the blue style tests I would prefer they also used the original options the same way you setup YAS. The reason being the purpose of those tests is to isolate the changes made by only the style and not any of the other options now on by default.

Adding a repo in test and then adding a config for blue style is a better way to do a full fledged integration test. Basically the same purpose Onda.jl has but for blue style.

nickrobinson251:
Makes sense -- I'll update the tests :)

nickrobinson251:
> For the blue style tests I would prefer they also used the original options the same way you setup YAS

Done 👍 

> Adding a repo in test and then adding a config for blue style is a better way to do a full fledged integration test

For now I've opened an issue about adding something like that for BlueStyle https://github.com/domluna/JuliaFormatter.jl/issues/301 :)

nickrobinson251:
Since this PR adds `format_text(str, style)` (with style a positional arg) as a method, how'd you feel about adding that as an interface for all format methods? i.e. so that `format(str, BlueStyle())` is a valid API, as well as `format(str, style = BlueStyle())`? Or do you prefer to keep the keyword-only methods for the other `format` functions (`format`, `format_file`, i guess `format_md`)?

domluna:
would format(str, BlueStyle()) and format(str, style=BlueStyle()) be the same thing?

nickrobinson251:
yeah, since removing `format(str, style=BlueStyle())`would be breaking, but also i kinda prefer `format(str, BlueStyle())` as the API, so we could allow both. Or we can stick with just `format(str, style=BlueStyle())` if you prefer :)

domluna:
I'm fine for having both - multiple dispatch ftw!

domluna (approved):
:tada: I'm good with merging this if you are

nickrobinson251:
yeah, i think this is good to go :) 

should i bump the patch version in the Project.toml?

thanks!

domluna:
sure or I could do it in a separate commit, either way

nickrobinson251:
Cool - thanks for your help. Merge whenever :)

PR #309, opened 2020-09-30
Extends `remove_extra_newlines` to remove empty newlines at the start or end of a code block

domluna:
**The exemption(s) from this rule are `module` code blocks which are permitted a single newline**

### Example

```julia
module M



a = 1

function foo()


    return nothing

end


b = 2


end
```

is rewritten as

```julia
module M

a = 1

function foo()
    return nothing
end

b = 2

end
```

On master this would currently be

```julia
module M

a = 1

function foo()

    return nothing

end

b = 2

end
```


nickrobinson251:
LGTM

PR #310, opened 2020-10-01
Inner styles for Blue and YAS styles

domluna:
This allows all styles to be interleaved which allows dispatch on the most appropriate style


nickrobinson251:
thanks!

Issue #338, opened 2020-11-28
Formatter turns good code into bad and violates margin


odow:
Here's an example from JuMP: https://github.com/jump-dev/JuMP.jl/pull/2387/files#r531813190
```Julia
s = """
mock = MOI.Utilities.MockOptimizer(
    MOI.Utilities.UniversalFallback(MOI.Utilities.Model{Float64}())
)
"""
format_text(s; margin = 80)

# yields

mock =
    MOI.Utilities.MockOptimizer(MOI.Utilities.UniversalFallback(MOI.Utilities.Model{Float64}()))
```

domluna:
this is the behavior for when there's a singular arg in (...), {...}, [...]

this could be revisited or tinkered with but there are instances where it does result in a nicer format

domluna:
hmmm looking more at this and I'm leaning towards making this change now. It also is more consistent behavior on the formatter's end.

odow:
I would expect the rule is: 
 - Does the line fit on one line?
 - If no, drop the outer argument. Does it fit on one line? Recurse.

That would result in things like this:
```Julia
mock = MOI.Utilities.MockOptimizer(
    MOI.Utilities.UniversalFallback(MOI.Utilities.Model{Float64}())
)

# or, if margin even shorter

mock = MOI.Utilities.MockOptimizer(
    MOI.Utilities.UniversalFallback(
        MOI.Utilities.Model{Float64}()
    )
)
```

nickrobinson251:
i _think_ `BlueStyle` has these line break and indent rules

domluna:
This is going to require creating an `unnest` type function similar to the way `pretty` or `nest` currently work. The current un-nesting mechanism makes some big assumptions for simplicity and is not extendable. This would be a good change overall, it's just a bit more involved than I originally hoped.

domluna:
@odow as of 0.12 this now formats as

```
julia> format_text(s; margin = 80) |> println
mock = MOI.Utilities.MockOptimizer(
    MOI.Utilities.UniversalFallback(MOI.Utilities.Model{Float64}()),
)
```

PR #345, opened 2021-01-01
Nest on single arguments in iterable types unless it's a single argument of the syntactic form {...}, [...], (...)

domluna:
ref #338

The gist is

1. Single arguments are now nested. Before we used a heuristic where if (using a Call as an example) there was a single argument `foo("hello")`. Then `"hello"` would not be nested, now if it exceeds the length it will become:

```
foo(
    "hello",
)
```

The main benefit of this being the formatter is more consistent when it comes to nesting, the second being the previous behavior caused subpar formats in certain cases, as shown in the above issue.

A downside of this (and why it was the way was) is we can get deeply nested structure in cases where joining multiple lines
would be desirable.

```
var = foo1(
  foo2(
    foo3(
      foo4(
        "hellooooooooooooooooooooooooooooooooooooooooooooooooooooo",
      ),
    ),
  ),
)
```

In this example we could join all the `foo*` calls on a single line

2. More sophisticated unnesting

This allows the `foo*` calls in the above example to be joined onto a single line so the upside of the previous behavior is maintained. 

A current caveat is this only works inside a nested binary op call and is the reason why this is a draft. I'm going to extend this to a top-level unnest function next.

That being said it may be worth merging this as is

> As an aside other formatter's don't make this attempt to unnest in these cases and it may only make contrived test cases format better so pursuing this may not be worthwhile. Instead we could just go with 1) and call it :thinking: 

## UPDATE

2) is being scrapped. Instead we'll keep the current behavior there is a single argument and it's of the syntactic form (...), [...], or {...}

Applying this to the above foo* example it will format as

```
var = foo1(
  foo2(
    foo3(
      foo4(
        "hellooooooooooooooooooooooooooooooooooooooooooooooooooooo",
      ),
    ),
  ),
)
```

but if the nested foo* were not present it would format as

```
var = foo1((((
  "hellooooooooooooooooooooooooooooooooooooooooooooooooooooo",
))))
```

To me this is a reasonable middle ground because the parenthesis without a prefix don't add any context when nested, it's syntactically superfluous.


test/issues.jl
```diff
@@ -251,26 +251,15 @@
...
         function f(args...)
 
             next!(
-                s.progress;
+                s.progress,
                 # comment
             )
             nothing
         end"""
-        @test_broken fmt(str_) == str
```
domluna, line 273: this is a nice side effect

PR #353, opened 2021-01-19
Update CompatHelper.yml

devmotion:
This PR updates the CompatHelper settings to the updated defaults. The main difference is that it is not required anymore to install Julia. In fact, I am wondering if CompatHelper fails (it does currently) because Julia 1.2 is too old.

### DIFF
--- a/.github/workflows/CompatHelper.yml
+++ b/.github/workflows/CompatHelper.yml
@@ -1,19 +1,16 @@
 name: CompatHelper
-
 on:
   schedule:
     - cron: '00 00 * * *'
-
+  workflow_dispatch:
 jobs:
   CompatHelper:
     runs-on: ubuntu-latest
     steps:
-      - uses: julia-actions/setup-julia@latest
-        with:
-          version: 1.2
       - name: Pkg.add("CompatHelper")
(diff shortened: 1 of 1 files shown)


.github/workflows/CompatHelper.yml
```diff
@@ -1,19 +1,16 @@
...
     steps:
-      - uses: julia-actions/setup-julia@latest
-        with:
-          version: 1.2
       - name: Pkg.add("CompatHelper")
         run: julia -e 'using Pkg; Pkg.add("CompatHelper")'
       - name: CompatHelper.main()
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          COMPATHELPER_PRIV: ${{ secrets.COMPATHELPER_PRIV }}
```
devmotion, line 15: CompatHelper requires a private key and a deploy token to be able to trigger Github CI. Maybe you use different names for the keys or just want to use the (possibly existing) keys for Documenter? See https://github.com/JuliaRegistries/CompatHelper.jl#12-set-up-the-ssh-deploy-key-optional.

devmotion:
BTW I noticed that the CompatHelper action fails currently since JuliaFormatter is not compatible with Documenter 0.26.

PR #354, opened 2021-01-19
Add compatibility with Documenter 0.26

devmotion:
This PR tries to add compatibility with Documenter 0.26 (there was no CompatHelper PR since the action fails currently, see also https://github.com/domluna/JuliaFormatter.jl/pull/353).


devmotion:
Thank you for merging this so quickly! Is it possible to make a new release with the updated compatibility bounds?

PR #364, opened 2021-02-10
do not write formatted string to file at all if overwrite==false

### LINKED ISSUE #363: Add feature to check if code is formatted correctly
Users may want to check if code is formatted correctly as part of a CI pipeline.  The ```check_format()``` function should only return true/false whether the code is correctly formatted, but *not* change anything if incorrectly formatted.  It should be the user's responsibility to fix the code formatting how he/she sees fit.

baumgold:
Closes #363

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "JuliaFormatter"
 uuid = "98e50ef6-434e-11e9-1051-2b60c6c9e899"
 authors = ["Dominique Luna <dluna132@gmail.com>"]
-version = "0.13.1"
+version = "0.13.2"
 
 [deps]
 CSTParser = "00ebfdb7-1f24-5e51-bd34-a7502290713f"
--- a/bin/format.jl
+++ b/bin/format.jl
@@ -29,9 +29,9 @@ Flags:
     -h, --help
         Print this message.
 
-    -o, --overwrite
-        Writes the formatted source to a new file where the original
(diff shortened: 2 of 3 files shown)


domluna:
bin/format.jl hasn't been updated in a while, IMO it should be deleted.

Doesn't https://github.com/julia-actions/julia-format/blob/master/workflows/format_check.yml#L34-L42 do the same thing?

baumgold:
> bin/format.jl hasn't been updated in a while, IMO it should be deleted.

Sounds good to me.

> Doesn't https://github.com/julia-actions/julia-format/blob/master/workflows/format_check.yml#L34-L42 do the same thing?

Yes but it only works for CI systems that use Github Actions.  I think a generic tool that can be called from any CI system would be useful.

domluna:
ic - well that snippet could be copied over, GHA is not a requirement for it.

baumgold:
> ic - well that snippet could be copied over, GHA is not a requirement for it.

That snippet depends on calling ```format(".", verbose=true)``` on L32 which modifies the files in-place.  My goal here is to be able to call ```format(".", overwrite=false)``` to check whether the files are formatted correctly without modifying the files.

domluna:
Ok that makes sense - the current overwrite functionality is not useful anyway so this looks good.

Can you update the docs here https://github.com/domluna/JuliaFormatter.jl/blob/acf788c3c9ebfc6532912534486325ca74748cf4/src/JuliaFormatter.jl#L386-L388

PR #365, opened 2021-02-17
configurable line ending normalization

pfitzseb:
Fixes https://github.com/domluna/JuliaFormatter.jl/issues/98.

### DIFF
--- a/src/JuliaFormatter.jl
+++ b/src/JuliaFormatter.jl
@@ -79,8 +79,14 @@ include("print.jl")
 
 include("markdown.jl")
 
-# on Windows lines can end in "\r\n"
-normalize_line_ending(s::AbstractString) = replace(s, "\r\n" => "\n")
+const UNIX_TO_WINDOWS = r"\r?\n" => "\r\n"
+const WINDOWS_TO_UNIX = "\r\n" => "\n"
+function line_ending_replacer(text)
+    rn = count("\r\n", text)
+    n = count(r"(?<!\r)\n", text)
+    n >= rn ? WINDOWS_TO_UNIX : UNIX_TO_WINDOWS
+end
+normalize_line_ending(s::AbstractString, replacer = WINDOWS_TO_UNIX) = replace(s, replacer)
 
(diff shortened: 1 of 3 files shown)


domluna:
LGTM only question I have is if the option is in a config file do you need to do any extra for will the symbol be read correctly 

As a reference https://github.com/domluna/JuliaFormatter.jl/pull/365/files#diff-8712d022b5e49f56d5b2d2b3ff1d7f072ae2e818741ef43d4eacc9084c4c2a3fL566-L579 here the style string is converted to the appropriate object.

pfitzseb:
Yeah, dunno. Switched to strings instead, which should definitely work :)

src/JuliaFormatter.jl
```diff
@@ -358,7 +370,15 @@ function format_text(cst::CSTParser.EXPR, style::AbstractStyle, s::State)
...
 
     text = String(take!(io))
-    text = normalize_line_ending(text)
+
+    replacer = if s.opts.normalize_line_endings === "unix"
+        WINDOWS_TO_UNIX
+    elseif s.opts.normalize_line_endings === "windows"
+        UNIX_TO_WINDOWS
+    else
+        choose_line_ending_replacer(s.doc.text)
```
pfitzseb, line 379: This is the original text, right? Is there a better way to get that?

domluna: not atm I think it's only referred to in one other place.

I think using the text arg in the function should work too - you would just have to not override it like we do right now on line 372

pfitzseb: This function only takes `cst::CSTParser.EXPR, style::AbstractStyle, s::State`, so we don't have that in here :)

domluna: oops misread that, then yeah s.doc.text is the original text :)

domluna (approved):
this seems g2g, anything else to do?

pfitzseb:
No this is good from my end.

PR #367, opened 2021-02-17
Remove dependency on Documenter.jl

davidanthoff:


domluna:
Wondering why the test CI didn't run

To fix the format CI call format(".") at the top directory, the new file and it's formatted version don't match up

PR #380, opened 2021-03-07
Solve #316 - Add option to replace `in` used in `always_for_in`

domluna:
Add an option `for_in_replacement`, that works alongside `always_for_in` whereby you can specify an alternative to `in`.

Can be used when `always_for_in` is `true` to replace `in` with ∈ (\\in), or `=` instead:
```julia
for a = 1:10
end
# formatted with always_for_in = true, for_in_replacement = "∈"
for a ∈ 1:10
end
```

### DIFF
--- a/src/options.jl
+++ b/src/options.jl
@@ -18,10 +18,10 @@ Base.@kwdef struct Options
     align_pair_arrow::Bool = false
     conditional_to_if::Bool = false
     normalize_line_endings::String = "auto"
-    in_replacement::String = "in"
+    for_in_replacement::String = "in"
 end
 
-function valid_in_op(s::String)
+function valid_for_in_op(s::String)
     s == "in" && return true
     s == "=" && return true
     s == "∈" && return true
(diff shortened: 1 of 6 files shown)


src/options.jl
```diff
@@ -18,4 +18,12 @@ Base.@kwdef struct Options
     align_pair_arrow::Bool = false
     conditional_to_if::Bool = false
     normalize_line_endings::String = "auto"
+    for_in_replacement::String = "in"
+end
+
+function valid_for_in_op(s::String)
```
iblis17, line 24:
```suggestion
valid_for_in_op(s::String) = s ∈ ("in", "=", "∈")
```

And... maybe we can run this function in constructor of `Options`?

domluna: that makes a lot of sense

domluna:
> Hi all, I'd really love to have this merged&released, is there anything I could potentially help with to get this through?

Probably just has to be rebased. And also do this https://github.com/domluna/JuliaFormatter.jl/pull/380#discussion_r650462183

exaexa:
OK, I'll push it to my fork ASAP. The rebase will be slightly challenging it seems.

PR #412, opened 2021-05-10
Adding revamped bin/format.jl

colinxs:
I had a script I've been using as a CLI version of JuliaFormatter.jl that I figured you might be interested in. 

Advantages over the existing format.jl are:
- Full support for all the arguments of `JuliaFormatter.format_text`
- Can handle stdin and paths
- Fancy shebang so that script can be directly executable

Downsides:
- Slightly slower startup because of ArgParse

Examples:
```
./bin/jlfmt ~/tmp/StaticArrays.jl/src/StaticArrays.jl --style blue -o indent 2 -o import_to_using true
cat ~/tmp/StaticArrays.jl/src/StaticArrays.jl | ./bin/jlfmt --style yas
```

It's only user has been me so it's seen minimal testing but it's fairly simple. Feel free to change however.


domluna:
@colinxs how do you use this from a fresh setup? Copy the contents in the bin folder locally and use that?

colinxs:
I personally use NixOS/Nix which makes bundling `jlfmt` and it's dependencies without fully compiling it like PackageCompiler.jl pretty easy. Outside of the ecosystem I'm not sure. Julia could really benefit by providing a mechanism for publishing executable scripts like Python does. I think PackageCompiler.jl is the only real choice, and has the benefit of fixing the slow startup which I know is a painpoint for many. I compiled a package ton too long ago and the new, revamped PackageCompiler.jl is fairly pain free.

PR #435, opened 2021-06-30
fix #429

### LINKED ISSUE #429: `short_to_long_function_def` no indent on first format
```
julia> str |> print
find_derivatives!(vars, expr::Equation, f=identity) = (find_derivatives!(vars, expr.lhs, f); find_derivatives!(vars, expr.rhs, f); vars)

julia> format_text(str, short_to_long_function_def=true) |> print
function find_derivatives!(vars, expr::Equation, f = identity)
(find_derivatives!(vars, expr.lhs, f); find_derivatives!(vars, expr.rhs, f); vars)
end
```

domluna:
create a block node for the new long function definition
if the original short form function definition was not wrapped in one.

### DIFF
--- a/src/passes.jl
+++ b/src/passes.jl
@@ -202,6 +202,8 @@ end
 ```
 """
 function short_to_long_function_def!(fst::FST, s::State)
+    (fst[1].typ !== Call && fst[1].typ !== Where) && return
+
     # 3 cases
     #
     # case 1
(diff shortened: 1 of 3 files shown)


src/passes.jl
```diff
@@ -212,61 +214,44 @@ function short_to_long_function_def!(fst::FST, s::State)
...
+        # ```
+        # function
+        #     body
         # end
-        kw = FST(KEYWORD, -1, fst[end].startline, fst[end].endline, "end")
-        add_node!(funcdef, kw, s)
-
-        fst.typ = funcdef.typ
-        fst.nodes = funcdef.nodes
-        fst.len = funcdef.len
```
domluna, line 268: this ensure case is not longer needed since CSTParser v3

Issue #405, opened 2021-05-04
Bug in weird edge case with docstrings


ericphanson:
For some reason, I was doing this:
```julia
module MyPkg

str = "a"
function f end
function __init__()
    @doc """
    Here's a docstring for `f` that includes dynamic content: $(str).
    """
    f
end
end
```
(where `str` is actually `PyObject` populated at init time.... The idea is to autogenerate parts of Julia docstrings from Python ones for a wrapper package).

This does seem to work, but JuliaFormatter (with `style=YASStyle()`) converts it to
```julia
module MyPkg

str = "a"
function f end
function __init__()
    @doc """
    Here's a docstring for `f` that includes dynamic content: $(str).
    """
    return f
end
end
```
which then errors because it can't document `return f`. Somehow it should turn this into
```julia
module MyPkg

str = "a"
function f end
function __init__()
    @doc """
    Here's a docstring for `f` that includes dynamic content: $(str).
    """
    f
    return nothing
end
end
```
instead.

Edit: it’s open source now, so for context the case where this came up is https://github.com/beacon-biosignals/CatBoost.jl/blob/a548b0bd58ca2a1c2f17cf65c0565420ae0f1b47/src/CatBoost.jl#L58

domluna:
edge case related to the always adding returns option

ericphanson:
Awesome, thanks!

PR #448, opened 2021-08-02
Restore 1.0 compatibility

pfitzseb:
I'd love to get JuliaFormatter into the next language server release, but we need Julia 1.0 compat for that. 

The `count` stuff is a bit ugly, but should be fine. All other test failures happened because of strict string comparisons, but Julia 1.3 and older returned `SubString`s instead of `String`s in certain cases. In general, `==` comparisons are more appropriate for comparing to string literals, imho.


test/issues.jl
```diff
@@ -708,7 +708,15 @@
...
-        @test fmt(str, always_use_return = true) == str
+        str_ret = """
+        function __init__()
+            return @doc raw\"""
+                   Doc string.
+                   \"""
+            f
+        end
+        """
+        @test fmt(str, always_use_return = true) == str_ret
```
pfitzseb, line 719: @domluna Your call on what this should do, but I think this is the more correct interpretation of `always_use_return` (and yes, it does the right thing).

domluna: https://github.com/domluna/JuliaFormatter.jl/issues/405

did the parsing of this change with v3.2.4?

domluna:
Looks good. Wonder why doc CI is failing

test/issues.jl
pfitzseb: Yes. I added a special case for doc parsing, so this parses `@doc "asd" f` as one expression, which is consistent with `Meta.parse`:
```
julia> str = "a"
"a"

julia> function f end
f (generic function with 0 methods)

julia> function __init__()
           @doc """
           Here's a docstring for `f` that includes dynamic content: $(str).
           """
           f
       end
__init__ (generic function with 1 method)

julia> __init__()
f

julia> function __init__()
           return @doc """
           Here's a docstring for `f` that includes dynamic content: $(str).
           """
           f
       end
__init__ (generic function with 1 method)

julia> __init__()
f
```

pfitzseb:
Dunno, maybe some secret got messed up (would explain the ` fatal: unable to access 'https://github.com/domluna/JuliaFormatter.jl.git/': The requested URL returned error: 403`).

Project.toml
```diff
@@ -11,11 +11,11 @@ Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
 Tokenize = "0796e94c-ce3b-5d07-9a54-7f471281c624"
 
 [compat]
-CSTParser = "^3.2.3"
+CSTParser = "^3.2.4"
```
domluna, line 14: Do we need to bump the package version too?

domluna:
> Dunno, maybe some secret got messed up (would explain the ` fatal: unable to access 'https://github.com/domluna/JuliaFormatter.jl.git/': The requested URL returned error: 403`).

yeah might just be a temporary thing, either way not gonna block this based on that.

Not sure if we need to bump the version of JuliaFormatter or not but this looks good to me when you're ready ill merge it in

domluna (approved):

Project.toml
pfitzseb: https://github.com/domluna/JuliaFormatter.jl/pull/448/commits/8c274fff8661055df3a6aa83f2c2802938ea2808

pfitzseb:
This is good to go on my end.

.github/workflows/ci.yml
```diff
@@ -9,6 +9,9 @@ jobs:
     strategy:
       matrix:
         version:
+          - '1.0'
+          - '1.1'
+          - '1.2'
```
davidanthoff, line 14: We also need to run CI on windows and mac and x86. Should we maybe just put the entire package under pkg butler management, so that it is aligned with how we handle all the other packages?

domluna: sounds alright to me

domluna:
@pfitzseb you'll have to bump it to 0.15.4 doing to quick patch release to fix an issue before this goes in

domluna:
@pfitzseb @davidanthoff is this ok now?

sidie note I'm not sure how to run all the github actions jobs, any ideas? I don't think this breaks anything though

pfitzseb:
Not sure what was going on with the macOS tests there (`realpath` failed even when only called if `ispath(root) == true`).

Tests run fine now though, so this is good to go from my side.

Issue #453, opened 2021-08-08
∞ changes using triple quotes in pairs context


t-bltg:
Hi, consider the following:

**bug.jl**
```julia
const _arg_desc = Dict(
    :lims =>
        """NTuple{2,Number} or Symbol. Force axis limits. Only finite values are used (you can set only the right limit with `xlims = (-Inf, 2)` for example).
           `:round` widens the limit to the nearest round number ie. [0.1,3.6]=>[0.0,4.0]
           `:symmetric` sets the limits to be symmetric around zero.""",
)
```

After first pass `$ julia -e 'using JuliaFormatter; format("bug.jl")'`:
```julia
const _arg_desc = Dict(
    :lims =>
        """NTuple{2,Number} or Symbol. Force axis limits. Only finite values are used (you can set only the right limit with `xlims = (-Inf, 2)` for example).
                   `:round` widens the limit to the nearest round number ie. [0.1,3.6]=>[0.0,4.0]
                   `:symmetric` sets the limits to be symmetric around zero.""",
)
```

After second pass $ julia -e 'using JuliaFormatter; format("bug.jl")':
```julia
const _arg_desc = Dict(
    :lims =>
        """NTuple{2,Number} or Symbol. Force axis limits. Only finite values are used (you can set only the right limit with `xlims = (-Inf, 2)` for example).
                           `:round` widens the limit to the nearest round number ie. [0.1,3.6]=>[0.0,4.0]
                           `:symmetric` sets the limits to be symmetric around zero.""",
)
```

And so on ... the second line of the multiline string keeps being indented with one more level.

domluna:
looks similar to an example Roger showed me a little while ago. I'm assuming something is going on with the nesting/unnesting mechanism here - related to `=>`

t-bltg:
Thanks for pointing that out.
Note that this bug happens with `BlueStyle` or `DefaultStyle` but **not** with `YASStyle`.
I'm trying to dig into the code to fix this.

domluna:
This looks good I expanded it to

```
λ JuliaFormatter.jl: git diff
diff --git a/src/fst.jl b/src/fst.jl
index 79e3ccc..9494b63 100644
--- a/src/fst.jl
+++ b/src/fst.jl
@@ -814,6 +814,13 @@ function is_assignment(cst::CSTParser.EXPR)
 end
 is_assignment(::Nothing) = false

+function is_pairarrow(cst::CSTParser.EXPR)
[...]

PR #454, opened 2021-08-08
Fix nesting in pairarrow context

t-bltg:
Fix https://github.com/domluna/JuliaFormatter.jl/issues/453.

### DIFF
--- a/test/default_style.jl
+++ b/test/default_style.jl
@@ -2878,6 +2878,15 @@
             end
         end"""
         @test fmt(str, 4, 1) == str
+
+        # https://github.com/domluna/JuliaFormatter.jl/issues/453
+        str = """
+        bar = Dict(
+            :foo => \"""A triple quoted literal string
+                    with some words in it.\""",
+        )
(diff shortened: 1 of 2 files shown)


test/default_style.jl
```diff
@@ -2878,6 +2878,15 @@
...
         @test fmt(str, 4, 1) == str
+
+        # https://github.com/domluna/JuliaFormatter.jl/issues/453
+        str = """
+        bar = Dict(
+            :foo => \"""A triple quoted literal string
+                    with some words in it.\""",
+        )
+        """
+        @test fmt(fmt(str, 4, 20), 4, 20) == str
```
domluna, line 2889: how come this calls `fmt` twice?

domluna: fmt already formats things twice btw, fmt1 is equivalent to calling format_text once

t-bltg: ok, fixed

t-bltg:
Thanks :+1:

Issue #423, opened 2021-06-04
do not remove style-compliant linebreaks


jrevels:
I'd like to expand the line length limit in my project via setting `margin` fairly high. However, reformatting after doing this will cause JuliaFormatter to try to fit everything on one line:

```diff
@@ -63,8 +63,7 @@ end
 function upsert(target::HasuraHTTPTarget, table_name, row; returning=["__typename"])
     mutation_target_name = string("upsert_", table_name)
     mutation_argument_name = :objects
-    mutation = (query=hasura_insert_mutation_string(mutation_target_name, mutation_argument_name, returning),
-                variables=to_hasura_gql((; mutation_argument_name => [row])))
+    mutation = (query=hasura_insert_mutation_string(mutation_target_name, mutation_argument_name, returning), variables=to_hasura_gql((; mutation_argument_name => [row])))
     response = HTTP.post(target.url, target.headers, JSON3.write(mutation))
     response_json = JSON3.read(String(response.body))
     if haskey(response_json, :data)
```

Both the "before" and "after" representations are YAS-compliant, but I as the author prefer the former representation. My naive opinion is that if code is already compliant with the chosen style guide, it should generally be considered to be in the purview of "author's choice" and left alone.

Is there a way for me to disable this behavior?

jrevels:
xref https://github.com/domluna/JuliaFormatter.jl/issues/359 ?

domluna:
yeah this should format how you would want once #359 is implemented

domluna:
@jrevels I have a prototype of this https://github.com/domluna/JuliaFormatter.jl/tree/dl/423 which repo + options would be good to run this on?

jrevels:
One of the repos where this comes up often for me is private, but I can run things with that branch and report back! Will hopefully time to do so later this week

kleinschmidt:
Looks like that ref was deleted...is there any path forward for this feature?

ericphanson:
cool! Could we make this the default for YASStyle? I guess as a breaking change?

domluna:
yeah that's why i didn't make it default :-)

ericphanson:
I think it's a bit of a BDFL situation (@jrevels) although I'm not sure that's ever been formalized

jrevels:
AFAICT the decision to make this a default is kind of orthogonal to the YASGuide itself, I guess, but FWIW I vote we make it a default :)

PR #470, opened 2021-09-06
Minimal Style

domluna:
So instead of creating a new style I think it's better to generalize the option of joining lines based on how they're present in the source file.

This is toggled with `ignore_maximum_width` option (gonna change the name since it's not quite accurate).

However the `margin` is still viable even if this option is turned on, so if you never want the margin to come into play then you would just set the maximum margin to a really high number. So this would also make the intention of #423 viable as well - where you want to keep the source intact unless it exceeds the maximum margin.

### TODO

- [ ] move all indenting out of nesting
- [ ] remove trailing whitespace
- [ ] tests


davidanthoff:
> However the `margin` is still viable even if this option is turned on

Could you explain that a bit more? I just don't understand it :)

domluna:
> > However the `margin` is still viable even if this option is turned on
> 
> Could you explain that a bit more? I just don't understand it :)

Yeah it was a bad explanation :( 

So it means that the formatter won't combine lines that are separate in the source file but it could still split them up if they are over the margin.

```
function foo(arg1,
                       arg2, arg3
                       )
```

Right now this would move all of this onto 1 line if it's < the maximum margin

With this option the above would remain intact but `arg2, arg3` could still be split if that line exceeds the maximum margin.

pfitzseb:
Which is totally reasonable, I think.  VSCode tends to choke on really long lines anyways, so setting `margin=1000` should be ok (although I guess we can always use some stupid high value instead just to be safe).

davidanthoff:
Ah, thanks for the explanation, that makes sense!

Wouldn't it make most sense that the `margin` value is just `Union{Nothing,Int}`?

pfitzseb:
OK, just ran this on our codebase and it looks pretty good. I'd like to add a config argument for trailing commas in multi-line lists and the `in`/`=` transform.

domluna:
I'm still working on it - can you link the diff you get?

maybe the stuff I'm working on locally is not worth it, especially if people use this option with a large margin then it's kind of pointless but if I can get it working it'll be really dope though

pfitzseb:
```
julia> JuliaFormatter.format(".";
           indent = 4,
           ignore_maximum_width = true,
           margin = 10_000
       )
```
results in https://github.com/julia-vscode/julia-vscode/pull/2334/commits/719ac4d247339161a2e10458489648b651e112c6.

davidanthoff:
So looking through that, here are some comments:
- we probably don't want no-indent module content if the module isn't the only one in the file? All the indents in modules in VSCodeServer.jl really do make sense, I think.
- It seems to add spaces around `=` in function signatures, I also don't think we want that, currently DocumentFormat.jl does the opposite.
- I think we don't want character replacements like `in` -> `=` by default, right?
- Same for trailing commas in lists, I don't think that should be on by default.
- It seems to add `::Any` to fields without a type declaration, again I don't think we want that.

Have to run now and haven't looked through the whole files, that is what I found so far. Once this is merged I also want to try this on the various code-bases we have in my lab.

pfitzseb:
> we probably don't want no-indent module content if the module isn't the only one in the file? All the indents in modules in VSCodeServer.jl really do make sense, I think.

Yeah, but not that big of a deal imho.

> It seems to add spaces around = in function signatures, I also don't think we want that, currently DocumentFormat.jl does the opposite.

That's configurable. I for one like having the spaces better.

> I think we don't want character replacements like in -> = by default, right?
> Same for trailing commas in lists, I don't think that should be on by default.
> It seems to add ::Any to fields without a type declaration, again I don't think we want that.

Agreed.

davidanthoff:
On the `=` in function signatures, that strikes me as something where tastes will just differ (I'm actually quite indifferent), but I think we should just stick with what DocumentFormat.jl does so that this whole transition is less disruptive for existing users.

domluna:
> It seems to add ::Any to fields without a type declaration, again I don't think we want that.

that's an option - set `annotate_untyped_fields_with_any` to false

domluna:
Updated compare `https://github.com/julia-vscode/julia-vscode/compare/master...domluna:dl/minimalstyle-fmt?expand=1`

domluna:
```
julia> format(".", ignore_maximum_width=true)
true
```

That's really nice :relieved:

domluna:
`remove_trailing_comma` is now an option 

defaults to `false`

davidanthoff:
I don't fully understand the meaning of `remove_trailing_comma` :) The problem previously was that it _added_ a comma at the end, right? Don't we need an option like `trailing_comma_handling` which can take values like `:off` (in the sense that no changes are made to the original code), `:add` and `:remove`? And `:off` would be the default?

domluna:
> I don't fully understand the meaning of `remove_trailing_comma` :) The problem previously was that it _added_ a comma at the end, right? Don't we need an option like `trailing_comma_handling` which can take values like `:off` (in the sense that no changes are made to the original code), `:add` and `:remove`? And `:off` would be the default?

My objection to this is it's not consistent. From the perspective of a style guide it makes more sense to be all in on something or not want to dabble with that at all. Rather than "well sometimes I want to put a trailing comma and sometimes I don't" which leads to confusion.

davidanthoff:
I don't follow, to be honest. It seems to me that the style could a) just leave trailing commas alone and not make any changes to them (this would correspond to `:off` in my proposal), or b) always add them (`:add`) or c) always remove them (`:remove`).

I don't understand how those choices correspond to one flag `remove_trailing_comma` that is a `bool`.

Are you saying that you think the formatter should _always_ either remove or add a trailing comma, but not have the option to just not do anything about them?

domluna:
> Are you saying that you think the formatter should always either remove or add a trailing comma, but not have the option to just not do anything about them?

I believe that's a better way to go and other formatters do the same. If you feel really strongly about the do nothing option we can do Union{Bool,Nothing} and then `nothing` would signify that. The actual implementation for that is really straightforward.

davidanthoff:
Yeah, I think our minimal default style really should not do anything other than indent and whitespace stuff, and then anything else should be in specific styles that one can opt-in to. I think our eventual goal should be to have the formatter on-by-default in VS Code with the format on type feature, and if that is our goal, I think we just need to have the default style be minimal, not opiniated and not invasive. Of course if a user then opts-into a specific style, these are great features to have.

In terms of naming the option, `remove_trailing_comma` does not strike me as ideal, if `remove_trailing_comma=false` then actually starts to _add_ trailing commas, that is not what I would expect from the name of the option.

domluna:
@pfitzseb @davidanthoff this is good to go now from my perspective

domluna:
https://github.com/domluna/julia-vscode/commit/5f7fbcc4f262f334afc8cd7f993396592b055432

domluna:
Any last comments before I merge this?

PR #515, opened 2021-12-02
Allow `always_for_in=nothing`

pfitzseb:
to disable `=`/`in` normalization.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "JuliaFormatter"
 uuid = "98e50ef6-434e-11e9-1051-2b60c6c9e899"
 authors = ["Dominique Luna <dluna132@gmail.com>"]
-version = "0.19.2"
+version = "0.20.0"
 
 [deps]
 CSTParser = "00ebfdb7-1f24-5e51-bd34-a7502290713f"
--- a/README.md
+++ b/README.md
@@ -53,7 +53,8 @@ be formatted across multiple lines.
 > default: `false`
 
 If true, `=` is always replaced with `in` if part of a `for` loop condition.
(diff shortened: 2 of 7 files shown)


domluna (approved):
nice! just wondering what the motivation was for this?

how does the vscode suite handle versioning btw? I generally to do a minor version bump whenever the change is more than a bug fix.

would bumping this to v0.20 mess with vscode at all?

davidanthoff:
Is this changing this globally? I thought we had a set of defaults for VS Code, and had assumed we would change it there?

domluna:
The default would still `false`, so there won't be any changes unless it's explicitly done, which could be where vscode comes in and changes it to nothing by default when using vscode

davidanthoff:
Ah, I misread this PR, ignore! So we'll need another accompanying PR on the LS side to fix this for VS Code.

pfitzseb:
> nice! just wondering what the motivation was for this?

Eh, we just try to keep the formatter as un-opinionated as possible for the extension, and `=` vs `in` seems like a fairly opinion based decision.

> how does the vscode suite handle versioning btw? I generally to do a minor version bump whenever the change is more than a bug fix.
>
> would bumping this to v0.20 mess with vscode at all?

Currently we're just pulling in `JuliaFormatter#master` for VS Code, so it doesn't really matter :)

> So we'll need another accompanying PR on the LS side to fix this for VS Code.

Yup. I'll prepare that once this is in.

domluna:
In that case can you bump the version to 0.20?

domluna:
@pfitzseb ready to merge?

pfitzseb:
Yup.

PR #534, opened 2022-01-10
change default for YAS to set `join_lines_based_on_source=true` by default

### LINKED ISSUE #423: do not remove style-compliant linebreaks

kleinschmidt:
WIP, not sure about tests but want to see what CI thinks

fixes #423 for YAS users at least, of which we can certainly count @jrevels who opened the issue :)

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "JuliaFormatter"
 uuid = "98e50ef6-434e-11e9-1051-2b60c6c9e899"
 authors = ["Dominique Luna <dluna132@gmail.com>"]
-version = "0.21.1"
+version = "0.22.0"
 
 [deps]
 CSTParser = "00ebfdb7-1f24-5e51-bd34-a7502290713f"
--- a/src/styles/yas/pretty.jl
+++ b/src/styles/yas/pretty.jl
@@ -13,6 +13,7 @@ Configurable options with different defaults to [`DefaultStyle`](@ref) are:
 - `short_to_long_function_def` = true
 - `always_use_return` = true
 - `whitespace_in_kwargs` = false
+- `join_lines_based_on_source` = true
 """
 struct YASStyle <: AbstractStyle
     innerstyle::Union{Nothing,AbstractStyle}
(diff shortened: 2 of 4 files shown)


domluna:
CI will probably fail

domluna:
Only a handful fail, for those I think it's fine to manually set  `join_lines_based_on_source` to false.

domluna:
I think you need to resolve the conflict for the tests to run

kleinschmidt:
looks like it needs an approval @domluna

domluna:
can you run format on the repo, probably some files in the YAS subdir are a little different

kleinschmidt:
I think it was that i used `, key=val` in all my tests kwargs :joy:

domluna (approved):
haha nice

domluna:
this LGTM any last things or we g2g?

kleinschmidt:
Looks good as far as I'm concerned!

PR #540, opened 2022-02-03
Support `trailing_comma = "nothing"` in TOML

### LINKED ISSUE #539: Cannot set trailing_comma=nothing in .JuliaFormatter.toml
As far as I know, TOML does not have a `nothing` value, so there is no way to set `trailing_comma = nothing` in `.JuliaFormatter.toml`.

The easiest solution would probably be to translate the string `"nothing"` in the TOML to `nothing`.

goerz:
Since TOML does not have a `nothing` type, this translates the string
`"nothing"` to the value `nothing` for the `trailing_comma` options.

Closes #539

### DIFF
--- a/src/JuliaFormatter.jl
+++ b/src/JuliaFormatter.jl
@@ -417,6 +417,13 @@ funccall(
 * When set to `false`, the trailing comma is always removed during nesting.
 * When set to `nothing`, the trailing comma appears as it does in the original source.
 
+In the [Configuration File](@ref), a `nothing` value can be set as the string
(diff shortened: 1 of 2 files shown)


domluna:
Looks good, can you confirm it works in a repo? Could be a good idea to add a test here as well

https://github.com/domluna/JuliaFormatter.jl/blob/master/test/config.jl

goerz:
I've verified in manually on my local workstation, but I'll still add a test

goerz:
Ok, Done

PR #549, opened 2022-02-17
CLI: Speed up via --compile=min -O0 flags

rht:
The -S feature has existed since Linux coreutils 8.30.
See https://stackoverflow.com/questions/4303128/how-to-use-multiple-arguments-for-awk-with-a-shebang-i-e/52979955#52979955.
The comment also says that it should work on FreeBSD (not tested).

This speeds up from ~1:16.69 s to ~31 s on my laptop (Linux).

TODO:
- [ ] I haven't tested this on macOS. I can't test this by myself.

### DIFF
--- a/bin/format.jl
+++ b/bin/format.jl
@@ -1,4 +1,4 @@
-#!/usr/bin/env julia
+#!/usr/bin/env -S julia --compile=min -O0
 using JuliaFormatter
 
 help = """


domluna:
i have a macbook air now so I can test this later.

I'm honestly surprised anyone uses this script still it's really, really, outdated; with respect to option selection.

rht:
> I'm honestly surprised anyone uses this script still it's really, really, outdated; with respect to option selection.

I suppose that no one uses the script because it's slow regardless of the speedup?
Do you think it is better to delete the script, and recommend people to use other means, like via vim integration?
Though I see it could be used in CI.

This is the original discussion that led to this PR: https://julialang.zulipchat.com/#narrow/stream/137791-general/topic/cli.20not.20included.

penelopeysm:
This file doesn't exist anymore and there are now faster ways to run JuliaFormatter I think :)

PR #551, opened 2022-03-02
Improve three-valued logic handling for config toml

pfitzseb:
The only functional change here is allowing `always_for_in = "nothing"` in .JuliaFormatter.toml, but the implementation is future proof in the sense that it inspects the `Options` type and automatically replaces all `"nothing"`s for fields that are declared as `Union{Bool, Nothing}`.

I've also added a bit of additional structure to the config tests via additional testsets.

### DIFF
--- a/src/JuliaFormatter.jl
+++ b/src/JuliaFormatter.jl
@@ -762,9 +762,11 @@ function kwargs(dict)
     return pairs(NamedTuple{ns}(vs))
 end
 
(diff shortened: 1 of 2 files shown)


domluna (approved):
looks great!

thanks for adding the @testsets. Makes that a file easier to read now :slightly_smiling_face:

domluna:
@pfitzseb There's a conflict since recent markdown config PR also touched this test file

pfitzseb:
Done.

PR #552, opened 2022-03-02
Improve jmd support

devmotion:
I just tried to format a jmd file, and it almost worked but there were two issues:
- The check of the file extension in `format_file` is too strict (hence I tried to call `format_md` directly)
- The YAML headers are not preserved when calling `format_md`

This PR fixes both issues.


domluna (approved):

domluna:
@devmotion anything else or this g2g?

devmotion:
This is good to go from my side 🙂

PR #554, opened 2022-03-02
Support ".jmd" in `format`

devmotion:
I noticed that unfortunately I forgot to add support for ".jmd" files in `format` in #552.


domluna (approved):
:+1:

PR #566, opened 2022-04-08
Update

Heptazhou:
+ Update CI workflow
+ Reduce redirection for links in README
+ Replace `!= nothing` with `!== nothing`
+ Fix typos

### DIFF
--- a/src/nest_utils.jl
+++ b/src/nest_utils.jl
@@ -35,7 +35,7 @@ should return a value other than `nothing`.
 """
 function walk(f, fst::FST, s::State)
     stop = f(fst, s)
-    (stop != nothing || is_leaf(fst)) && return
+    (stop !== nothing || is_leaf(fst)) && return
     walk(f, fst.nodes, s, fst.indent)
 end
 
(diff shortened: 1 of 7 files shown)


Heptazhou:
actually `!= nothing` is several times slower than `!== nothing` or `!isnothing()` ...

src/nest_utils.jl
```diff
@@ -35,7 +35,7 @@ should return a value other than `nothing`.
 """
 function walk(f, fst::FST, s::State)
     stop = f(fst, s)
-    (stop != nothing || is_leaf(fst)) && return
+    (!isnothing(stop) || is_leaf(fst)) && return
```
domluna, line 38: I don't think this is compatible with earlier versions of Julia so we need to use !== or ===

domluna (commented):
Can you add a description for the PR, it looks like it updates the CI workflow and typos, etc

domluna:
@Heptazhou looks good, let me know if this is done

Heptazhou:
@domluna It's good to go.

Issue #569, opened 2022-04-26
Allow format of .qmd files in addition to .jmd


dmbates:
https://quarto.org/ provides for technical writing with evaluation of code blocks in Python, R, Julia and Observable.  It is a comprehensive and flexible system.  The source files carry a .qmd extension.  I can format the Julia code by moving the file to a .jmd extension, formatting that file and moving back to a .qmd extension but it would be cleaner to just allow the .qmd extension.  I am happy to create a PR if this seems like a good idea.

dmbates:
I am trying to find out where a Julia code block in the text is detected.  In a Quarto .qmd file an executable Julia code blocks starts with three backticks and `{julia}` (not sure how to write that in a markdown message).  It appears that if I remove the braces around `julia` the code block is recognized, but the braces are necessary for the Quarto file to be processed correctly.

domluna:
it's pretty well hidden, should be in the markdown.jl file but this line https://github.com/domluna/JuliaFormatter.jl/blob/master/src/styles/default/pretty.jl#L241 is what you're looking for.

Now the markdown parser is based on https://github.com/MichaelHatherly/CommonMark.jl and I'm unsure how well it supports .qmd files

PR #570, opened 2022-04-27
Allow for .qmd extension and {julia} code blocks.

dmbates:
- a `.qmd` file extension, used by [Quarto](https://quarto.org), is similar to `.jmd` except that Julia code blocks are flagged as `{julia}`, not just `julia`
- this PR allows for both these extensions - #569
- added test

### DIFF
--- a/src/JuliaFormatter.jl
+++ b/src/JuliaFormatter.jl
@@ -643,7 +643,7 @@ function format_file(
 )::Bool
     path, ext = splitext(filename)
     shebang_pattern = r"^#!\s*/.*\bjulia[0-9.-]*\b"
-    formatted_str = if !isnothing(match(r"^\.[jq]*md$", ext))
+    formatted_str = if match(r"^\.[jq]*md$", ext) ≠ nothing
         format_markdown || return true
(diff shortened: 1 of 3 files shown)


domluna:
LGTM - running tests

anything else to add @dmbates  ?

dmbates:
I should add a test or @palday will feel that his training of me has failed.

src/JuliaFormatter.jl
```diff
@@ -643,7 +643,7 @@ function format_file(
 )::Bool
     path, ext = splitext(filename)
     shebang_pattern = r"^#!\s*/.*\bjulia[0-9.-]*\b"
-    formatted_str = if ext == ".md" || ext == ".jmd"
+    formatted_str = if !isnothing(match(r"^\.[jq]*md$", ext))
```
domluna, line 646: oops forgot about this initially but we can't use `isnothing` since it's not compatible with earlier Julia versions, need `===` or `!==` instead

dmbates:
I think this is ready to go new.

domluna:
alright looks good!

PR #583, opened 2022-05-21
The last semicolon cannot be drop in vcat if that's the only semicolon

YingboMa:
- Fix 582
- Add tests

### DIFF
--- a/test/yas_style.jl
+++ b/test/yas_style.jl
@@ -578,8 +578,8 @@
     end
 
     @testset "issue 582 - vcat" begin
-        @test format_text("[sts...;]", YASStyle()) == "[sts...;]"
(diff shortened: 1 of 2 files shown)


test/yas_style.jl
```diff
@@ -576,4 +576,10 @@
         str = "using A"
         @test yasfmt(str) == str
     end
+
+    @testset "issue 582 - vcat" begin
+        @test format_text("[sts...;]", YASStyle()) == "[sts...;]"
+        @test format_text("[a;b;]", YASStyle()) == "[a; b]"
+        @test format_text("[a;b;;]", YASStyle()) == "[a; b;;]"
```
domluna, line 583: can you use yasfmt instead of format_text, yasfmt does a double format to ensure the formatted output remains static

PR #588, opened 2022-05-23
SciMLStyle

YingboMa:


domluna (approved):

src/JuliaFormatter.jl
```diff
@@ -549,6 +552,23 @@ function format_text(text::AbstractString, style::AbstractStyle; kwargs...)
...
+    iter = 1
+    while formatted_text != text
+        iter > maxiters &&
+            error("formatted_text hasn't reached to a fixpoint in $iter iterations")
+        text = formatted_text
+        formatted_text = format_text(text, style, opts)
+        iter += 1
+    end
+    return formatted_text
+end
```
davidanthoff, line 570: That feels quite hacky? Also, not clear how this would interact with a potential partial formatting in an IDE? Wouldn't it make sense to require that a second call to `format_text` is always a no-op?

src/JuliaFormatter.jl
```diff
@@ -549,6 +552,23 @@ function format_text(text::AbstractString, style::AbstractStyle; kwargs...)
     return format_text(text, style, opts)
 end
 
+function format_text(text::AbstractString, style::SciMLStyle; maxiters = 3, kwargs...)
+    isempty(text) && return text
+    opts = Options(; merge(options(style), kwargs)...)
+    # We need to iterate to a fixpoint because the result of short to long
+    # form isn't properly formatted
```
domluna, line 559: maybe we should just make this part work properly then we wouldn't have to do this

domluna: @YingboMa can you give a minimal example of this?

YingboMa: It's just https://github.com/domluna/JuliaFormatter.jl/pull/588/files#diff-60c9ecaf817f34a35b04e83d428e16cd44a2bbf06b642a1fffc8f67d39e2aa9dR2-R12

domluna: ok made an issue to track it, since this is function is only applicable to SciMLStyle this is fine for now.

PR #605, opened 2022-05-31
Set all options in each style

davidanthoff:


domluna:
format check fails hmmm

davidanthoff:
I think also ready to merge?

PR #629, opened 2022-08-03
implement `long_to_short_function_def`

t-bltg:
This is a proposition in order to transform long function definitions (with a single statement) to short defs, if the margin allows it.

With this option enabled, declarations such as 
```julia
function foo(a)
    return a + 1
end
```
are transformed to
```julia
foo(a) = a + 1
```


test/options.jl
```diff
@@ -284,6 +284,63 @@
...
+        end"""
+        str = "foo(a) = a + 1"
+        @test fmt(str_, 4, length(str), long_to_short_function_def = true) == str
+
+        str = """
+        function foo()
+            expr1
+            expr2
+        end"""
+        @test fmt(str, 4, length(str), long_to_short_function_def = true) == str
```
domluna, line 324: length(str) would be the length of the entire string not just the longest line so I'm not sure what this test checks for

test/options.jl
```diff
@@ -284,6 +284,63 @@
...
+            end
+        end"""
+        str = """
+        foo(a) =
+            if a > 1
+                2
+            else
+                nothing
+            end"""
+        @test fmt(str_, 4, length(str), long_to_short_function_def = true) == str
```
domluna, line 341: same here, is it just that if the margin is big it will go to the short def kind of thing?

domluna (commented):
looks good, a few comments

does this conflict with `short_to_long_function_def` at all? For example what happens if they're both set to `true`?

src/options.jl
```diff
@@ -8,6 +8,7 @@ Base.@kwdef struct Options
     import_to_using::Bool = false
     pipe_to_function_call::Bool = false
     short_to_long_function_def::Bool = false
+    long_to_short_function_def::Bool = false
```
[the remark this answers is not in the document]
t-bltg: Sorry for missing that indeed, fixed.

test/options.jl
t-bltg: So this test, regardless of the margin, has two statements: `expr1` and `expr2` so this tests that the transform `long_to_short_function_def` is a noop, since we allow transformation only for single statements.

test/options.jl
t-bltg: So this one is a bit more tricky. I wished it could be transformed into:
```julia
foo(a)  = if a > 1
    2
else
    nothing
end
```
in order to same another extra line, but I couldn't find a fix regarding the addition of newlines in https://github.com/domluna/JuliaFormatter.jl/blob/598a2485bbd36a118c66b3774e73684c4a4e9697/src/fst.jl#L647.

However, the current (saving one line over the long def) is imo satisfactory enough.

t-bltg:
> does this conflict with short_to_long_function_def at all? For example what happens if they're both set to true?

I knew you'd ask :). Currently, `short_to_long_function_def` preempts `long_to_short_function_def` since it is parsed after in https://github.com/domluna/JuliaFormatter.jl/blob/598a2485bbd36a118c66b3774e73684c4a4e9697/src/styles/default/nest.jl#L81, but if you wish, we could issue a warning if both conflicting options are set ...

domluna:
> > does this conflict with short_to_long_function_def at all? For example what happens if they're both set to true?
> 
> I knew you'd ask :). Currently, `short_to_long_function_def` preempts `long_to_short_function_def` since it is parsed after in
> 
> https://github.com/domluna/JuliaFormatter.jl/blob/598a2485bbd36a118c66b3774e73684c4a4e9697/src/styles/default/nest.jl#L81
> 
> , but if you wish, we could issue a warning if both conflicting options are set ...

Yeah a warning would be good since it might be confusing

src/options.jl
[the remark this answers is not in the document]
domluna: no worries we've just started enforcing these recently since v2 will default to MinimalStyle instead of DefaultStyle and so the base options will be different. This way if  you are using another style nothing weird will happen

src/JuliaFormatter.jl
```diff
@@ -633,6 +633,10 @@ function format_text(cst::CSTParser.EXPR, style::AbstractStyle, s::State)
 
     needs_alignment(s.opts) && align_fst!(fst, s.opts)
 
+    if s.opts.long_to_short_function_def && s.opts.short_to_long_function_def
+        @warn "Both `long_to_short_function_def` and `short_to_long_function_def` are set"
+    end
+
```
domluna, line 639: this should be done as earlier as possible in the process so ideally before we we initially parse the code

t-bltg: Is the new position in https://github.com/domluna/JuliaFormatter.jl/pull/629/commits/d091dd1690b4ca755c25d0881112687f22ff5f16 ok ?

domluna: yeah that looks good

test/options.jl
domluna: yeah I guess it would be the inverse of `conditional_to_if`

t-bltg: Maybe I'll have some time to implement `if_to_conditional` later. I'll open an issue if someone wants to help.

domluna (approved):

test/options.jl
domluna: hmm  could you just replace that Newline node with Whitespace(0) ?

domluna: either way this is not a big deal, it's a subjective addition

domluna:
This looks good, lemme know if we are g2g

test/options.jl
t-bltg: > hmm could you just replace that Newline node with Whitespace(0) 

I tried that, but it didn't work :/

t-bltg:
> This looks good, lemme know if we are g2g

Yes, ready to merge :+1:

PR #641, opened 2022-09-22
Log file being formatted when failure occurs

### LINKED ISSUE #640: Log which file is being formatted when failure occurs
`format(filename; kwargs...)` often gives unhelpful error messages, and that very well may be the best and safest thing to do because we perhaps can't reason about code without trying to compile it.

However, `format(path; kwargs...)` gives unhelpful results, and 99% of the time I spend a few minutes applying `format` to all the Julia files I see in `git diff main` on a development branch. Or worse (e.g., for the very first time in a new repo) file by file. I think many people, including myself, would find it useful if we were to wrap `format` in a try-catch so that we can still throw the same error but also print out which file is failing to format.

I'm not super familiar with JuliaFormatter.jl but, concretely, I want a function like this:

```julia
function logged_format(paths; kwargs...)
    jl_files = map(paths) do path
        [
            joinpath(root, f) for
            (root, dirs, files) in Base.Filesystem.walkdir(path) for
            f in files if endswith(f, ".jl")
        ]
    end
    for f in Iterators.flatten(jl_files)
        try
            format(f; kwargs...)
        catch # still error, but tell us which file this happened in
            @error "Error in formatting file $f" # could also be `@warn` or `@info`
            format(f; kwargs...)
        end
    end
end
```

charleskawczynski:
Closes #640.

### DIFF
--- a/src/JuliaFormatter.jl
+++ b/src/JuliaFormatter.jl
@@ -844,7 +844,13 @@ function format(paths; options...)::Bool
             else
                 options
             end
-            _format_file(path; opts...)
+            try # log which file fails
+                # https://github.com/domluna/JuliaFormatter.jl/issues/640
+                _format_file(path; opts...)
+            catch
+                @info "Error in formatting file $path"
+                _format_file(path; opts...)
+            end
         else
             reduce(walkdir(path), init = true) do formatted_path, dir_branch
                 root, dirs, files = dir_branch
@@ -861,7 +867,13 @@ function format(paths; options...)::Bool
                         else
                             options
                         end
-                        _format_file(full_path; opts...)
+                        try # log which file fails
+                            # https://github.com/domluna/JuliaFormatter.jl/issues/640
+                            _format_file(full_path; opts...)
+                        catch
+                            @info "Error in formatting file $full_path"
+                            _format_file(full_path; opts...)
+                        end
                     else
                         true
                     end
--- a/src/JuliaFormatter.jl
+++ b/src/JuliaFormatter.jl
@@ -873,6 +873,7 @@ function format(paths; options...)::Bool
                             _format_file(full_path; opts...)
                         catch
                             @info "Error in formatting file $full_path"
+                            @debug "formatting failed due to" exception=(err, catch_backtrace()
                             _format_file(full_path; opts...)
                         end
                     else
--- a/src/JuliaFormatter.jl
+++ b/src/JuliaFormatter.jl
@@ -873,7 +873,7 @@ function format(paths; options...)::Bool
                             _format_file(full_path; opts...)
                         catch
                             @info "Error in formatting file $full_path"
-                            @debug "formatting failed due to" exception=(err, catch_backtrace()
+                            @debug "formatting failed due to" exception=(err, catch_backtrace())
                             _format_file(full_path; opts...)
                         end
                     else
--- a/src/JuliaFormatter.jl
+++ b/src/JuliaFormatter.jl
@@ -849,7 +849,7 @@ function format(paths; options...)::Bool
                 _format_file(path; opts...)
             catch
                 @info "Error in formatting file $path"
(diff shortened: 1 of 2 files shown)


domluna:
nice

src/JuliaFormatter.jl
```diff
@@ -844,7 +844,13 @@ function format(paths; options...)::Bool
...
                 options
             end
-            _format_file(path; opts...)
+            try # log which file fails
+                # https://github.com/domluna/JuliaFormatter.jl/issues/640
+                _format_file(path; opts...)
+            catch
+                @info "Error in formatting file $path"
+                _format_file(path; opts...)
+            end
```
pfitzseb, line 853: Thoughts on (optionally) showing the original exception? E.g.
```suggestion
            catch
                @info "Error in formatting file $path"
                @debug "formatting failed due to" exception=(err, catch_backtrace()
                _format_file(path; opts...)
            end
```

src/JuliaFormatter.jl
```diff
@@ -861,7 +868,13 @@ function format(paths; options...)::Bool
                         else
                             options
                         end
-                        _format_file(full_path; opts...)
+                        try # log which file fails
+                            # https://github.com/domluna/JuliaFormatter.jl/issues/640
+                            _format_file(full_path; opts...)
+                        catch
+                            @info "Error in formatting file $full_path"
```
domluna, line 875:
```suggestion
                            @info "Error in formatting file $full_path"
                            @debug "formatting failed due to" exception=(err, catch_backtrace()
```

src/JuliaFormatter.jl
```diff
@@ -861,7 +868,14 @@ function format(paths; options...)::Bool
...
+                            @debug "formatting failed due to" exception=(err, catch_backtrace()
```
domluna, line 876:
```suggestion
                            @debug "formatting failed due to" exception=(err, catch_backtrace())
```

src/JuliaFormatter.jl
```diff
@@ -844,7 +844,14 @@ function format(paths; options...)::Bool
             else
                 options
             end
-            _format_file(path; opts...)
+            try # log which file fails
+                # https://github.com/domluna/JuliaFormatter.jl/issues/640
+                _format_file(path; opts...)
+            catch
+                @info "Error in formatting file $path"
+                @debug "formatting failed due to" exception=(err, catch_backtrace()
```
domluna, line 852:
```suggestion
                @debug "formatting failed due to" exception=(err, catch_backtrace())
```

charleskawczynski:
I think that's all for me, and including the original backtrace seems like a good idea

src/JuliaFormatter.jl
```diff
@@ -844,7 +844,14 @@ function format(paths; options...)::Bool
             else
                 options
             end
-            _format_file(path; opts...)
+            try # log which file fails
+                # https://github.com/domluna/JuliaFormatter.jl/issues/640
+                _format_file(path; opts...)
+            catch
+                @info "Error in formatting file $path"
+                @debug "formatting failed due to" exception=(err, catch_backtrace())
```
domluna, line 852:
```suggestion
                @debug "formatting failed due to" exception = (err, catch_backtrace())
```

src/JuliaFormatter.jl
```diff
@@ -861,7 +868,14 @@ function format(paths; options...)::Bool
                         else
                             options
                         end
-                        _format_file(full_path; opts...)
+                        try # log which file fails
+                            # https://github.com/domluna/JuliaFormatter.jl/issues/640
+                            _format_file(full_path; opts...)
+                        catch
+                            @info "Error in formatting file $full_path"
+                            @debug "formatting failed due to" exception=(err, catch_backtrace())
```
domluna, line 876:
```suggestion
                            @debug "formatting failed due to" exception = (err, catch_backtrace())
```

Issue #639, opened 2022-09-13
Feature request: Options as a function argument have higher priority compared to config file


senhalil:
At the moment the options are merged in a way that if the same option is supplied as a function argument and via config file then the one coming from the config file takes the priority; however, this prevents the user implementing a logic of "default" vs programmatically "conditional" settings. 

My observation over the years is that generally command line (overrides everything) -> environment variable -> config file -> defaults. In the case of a library "command line" represents the function call.

If you like the idea, I would be happy to open a PR.

domluna:
@senhalil This makes sense to me and tbh I thought this was already how it worked but I haven't used the package in this way myself so I might be wrong. A PR would be dope!

penelopeysm:
This would be breaking, but I think it makes sense to put in v3.

penelopeysm:
Looking at it carefully now, this does already work correctly

PR #643, opened 2022-10-06
Slight refactor + Ignore files and directories

gustaphe:
I was toying with the idea of allowing an `ignored` section in `.JuliaFormatter.toml`. Ironically, I had some trouble decoding what the `format` function was actually doing (mostly because of `a = if ...` constructions and missing `return`s), so I did a light refactor to help me bone out the `ignore` part. Since this way of doing it seems more intuitive to me, and comes with a very slitght positive difference in benchmark (see below), I'm leaving it in. If you like the `ignored` thing, but prefer the unrefactored version, I might go back and try to incorporate it.

At current master:
* `rm -r test/files docs` (so they don't affect the timing), `@btime format(JuliaFormatter)`: `199.845 ms (2184853 allocations: 129.94 MiB)`

With PR:
* `@btime format(JuliaFormatter)`: `197.540 ms (2176641 allocations: 129.47 MiB)` (this is barely a noticeable difference, but at least there is no penalty for the increased recursivity)
* Can specify files and directories to be ignored in the `.JuliaFormatter.toml` file, which will be skipped.


.JuliaFormatter.toml
```diff
@@ -1,3 +1,4 @@
 remove_extra_newlines = true
 always_for_in = true
 conditional_to_if = true
+ignored = ["bin", "test/files", "docs"]
```
domluna, line 4: why is test/files ignored? This is part of the test suite

gustaphe: They are unformatted on purpose, right? So `format(JuliaFormatter)` shouldn't be allowed to format them. Note that there's a config file in `test/files`, so `format("test/files")` doesn't read the top level config.

gustaphe:
So one thing that's kind of strange right now is this situation:
```
.
├── a
│   ├── a.jl
│   ├── b.jl
│   └── .JuliaFormatter.toml
├── b
│   ├── a.jl
│   └── b.jl
└── .JuliaFormatter.toml
```
If `./.JuliaFormatter.toml` contains `ignored = ["a.jl"]`, then `format(".")` will not touch either `a.jl` (unless `a/.JuliaFormatter.toml` has an `ignored` section and overwrites it), and `format("b")` won't either. But `format("a")` *will*, because it finds a config file in `a` and so doesn't look further up the tree. In that sense, the existence of the *file* in the current directory matters more than the existence of the *option*.
This is true for other options as well, of course, but it might be a bit more jarring here.

Another thing to consider is that `format("b/a.jl")` will report that the file is already formatted since it's supposed to be ignored. If one wants to manually make sure to format an ignored file, one way would be to specify `ignored=String[]`, but that requires #639 . If `.JuliaFormatter.toml` instead says `ignored = "b"`, then `format(".")` and `format("b")` do not touch anything in `b`, but `format("b/a.jl")` does. One would hope that package authors who want to use this option take the time to make sure that whatever patterns they choose actually do the right thing.

.JuliaFormatter.toml
domluna: They are formatted though, they're part of the test suite when format(".") is called at the top level.

domluna: just to clarify - does the `ignored` functionality stop if the subdirectory has it's own format file? So if at the top level I have ignored=["subdir"] but then subdir has it's own .JuliaFormatter.toml. Does subdir still get formatted according to its own rules?

gustaphe: So that's `test/format_repo.jl`? I guess one could overwrite the `.JuliaFormatter.toml` for that test specifically.

gustaphe: Strangely, if I add
```
open(joinpath(sandbox_dir, ".JuliaFormatter.toml"), "w") do io
    write(io, """
          remove_extra_newlines = true
          always_for_in = true
          conditional_to_if = true
          """)
end
```
to that test, the first two `format` calls return `false`, because some comments in `test/files/Docs.jl` are indented, and `test/files/reverse.jl` has a couple of lines being split into several. I don't quite understand why this happens.

domluna: hmmm I think that makes sense because test/files has it's own .JuliaFormatter.toml file

https://github.com/domluna/JuliaFormatter.jl/blob/master/test/files/.JuliaFormatter.toml

that only has remove_extra_newlines in it so when format(".") is called at the top level only that option will be overwritten (from the defaults)

where as the top level has https://github.com/domluna/JuliaFormatter.jl/blob/master/.JuliaFormatter.toml

```
remove_extra_newlines = true
always_for_in = true
conditional_to_if = true
```

So if you call format on test/files with the options of the top-level then it makes sense some files would turn out differently.

domluna: The most specific .JuliaFormatter.toml file is chosen. If the structure is

A
  B
  C
   D

and A, C, and D have their own .JuliaFormatter.toml files then each is used for that repo. Files in D uses D's .toml file, same with files in C.

B however since it doesn't have its own file will default to using A's .toml file

domluna:
> So one thing that's kind of strange right now is this situation:
> 
> ```
> .
> ├── a
> │   ├── a.jl
> │   ├── b.jl
> │   └── .JuliaFormatter.toml
> ├── b
> │   ├── a.jl
> │   └── b.jl
> └── .JuliaFormatter.toml
> ```
> 
> If `./.JuliaFormatter.toml` contains `ignored = ["a.jl"]`, then `format(".")` will not touch either `a.jl` (unless `a/.JuliaFormatter.toml` has an `ignored` section and overwrites it), and `format("b")` won't either. But `format("a")` _will_, because it finds a config file in `a` and so doesn't look further up the tree. In that sense, the existence of the _file_ in the current directory matters more than the existence of the _option_. This is true for other options as well, of course, but it might be a bit more jarring here.
> 
> Another thing to consider is that `format("b/a.jl")` will report that the file is already formatted since it's supposed to be ignored. If one wants to manually make sure to format an ignored file, one way would be to specify `ignored=String[]`, but that requires #639 . If `.JuliaFormatter.toml` instead says `ignored = "b"`, then `format(".")` and `format("b")` do not touch anything in `b`, but `format("b/a.jl")` does. One would hope that package authors who want to use this option take the time to make sure that whatever patterns they choose actually do the right thing.

I think it's ok if `format(".")` and `format("b")` do different things, assuming the format file at "." has `ignored=["b"]`. Although I believe b/ would need to have it's own format file otherwise it would use the file in a/ and then the format would do nothing? Or is the format relative based on the path?

The trickiest part of the API from my perspective is that if you have multiple format files in a repo then you need to specify which files you want to ignore in the file that most relevant.

So if the structure is

```
A
  B
  C
    D
```

and you want to ignore files in D, then you should do that from D's format file (assuming it has one) rather than A's format file for example.

.JuliaFormatter.toml
gustaphe: But then I would expect the second `format` to return `true`, because I'm running the same call twice?

And is the intention for nested configs to overwrite entirely? Because in the current implementation of this PR, if you call `format("A")`, only the options mentioned in `C/.JuliaFormatter.toml` are overwritten - for instance, say `A/.JuliaFormatter.toml` has 
```toml
remove_extra_newlines = true
always_for_in = true
```
 and `C/.JuliaFormatter.toml` has
```toml
always_for_in = false
conditional_to_if = true
```
then the effective config for files in `C` is
```toml
remove_extra_newlines = true
always_for_in = false
conditional_to_if = true
```

I think that's how it worked before the PR too.

gustaphe: (This, btw, only applies for `format("A")`, if we run `format("C")`, `A/.JuliaFormatter.toml` is never read, so the effective config is
```
always_for_in = false
conditional_to_if = true
```
)

gustaphe:
Well, you need to do it in D *if* you want it to apply even when running `format("D")` and you have a `.JuliaFormatter.toml` file in the tree downstream from A. For most package development I believe it will be perfectly fine to supply a top level `.JuliaFormatter.toml` that specifies which files and directories should be ignored, and then have developers run `format(PackageName)`. Hopefully this means there's very little reason to run `format` on specific subdirectories.

gustaphe:
Okay, I'll need an opinion on this, becuase I've figured out the issue.

I would have expected a configfile in a subdirectory to only overwrite the settings it mentions, but before this PR it would overwrite the entire config file. This is definitely intentional?

domluna:
> Okay, I'll need an opinion on this, becuase I've figured out the issue.
> 
> I would have expected a configfile in a subdirectory to only overwrite the settings it mentions, but before this PR it would overwrite the entire config file. This is definitely intentional?

Overwrite the settings relative to a config file in a higher directory? Every config file starts from the same base of defaults. I don't think assuming it should inherit overrides from higher up is a good idea, it's better to be explicit about the options.

gustaphe:
Okay, I had just misunderstood. Then the latest commit does it. Now there's a struct keeping track of which arguments came from the function call and which arguments are from a config file, so we can replace those when we encounter a new config file.

I'm sorry this turned into a bigger overhaul than I initially intended. Lmk if you have any more questions.

README.md
```diff
@@ -453,6 +453,10 @@ function func(...) where {TPARAM}
 end
 ```
 
+### `ignored`
+
+An array of paths to files and directories (with possible Glob wildcards)
+which should not be processed.
```
domluna, line 459: should be under File Options no?

domluna, line 459:
```suggestion
### `ignored`

An array of paths to files and directories (with possible Glob wildcards)
which will not be formatted.
```

gustaphe:
Bikeshedding: `ignored` sounds most like the name of a list to me, but an argument can be made for `skip`, `ignore`, `pass`, `pesah` ... 
Especially `ignore` seems common, it's what codecov and git use.

.JuliaFormatter.toml
```diff
@@ -1,3 +1,4 @@
 remove_extra_newlines = true
 always_for_in = true
 conditional_to_if = true
+ignored = ["bin", "docs"]
```
domluna, line 4: I'm totally fine with these being ignored but I'm just wondering things are still fine if they are not ignored right?

gustaphe: I guess. Can't remember why I picked those specifically. They should probably not be ignored.

test/config.jl
```diff
@@ -421,4 +421,38 @@
...
+        nottobeignored = (
+            "a.jl",
+            "other_directory/a.jl",
+            "other_directory/directory/a.jl",
+            "third_directory/b.jl",
+            "third_directory/ignored_directory/b.jl",
+        )
+        try
+            cp("files/ignore", sandbox_dir)
+            @test format(sandbox_dir) == false
```
domluna, line 445: jw if this is run again the output the true correct?

gustaphe: Yes. I should add a second run.

domluna: other than that this looks really good, appreciate you working on this

gustaphe: I'm having fun. Thank you for writing a good package!

domluna:
let's go with `ignore` and then this looks g2g

gustaphe:
Hold up, I did not intend for the confused history, I'll try to clean it up

gustaphe:
I thought of another place where this option should be mentioned in the documentation. Now I think this is pretty complete @domluna

domluna:
looks good! it says it can't be rebased due to conflicts though

gustaphe:
Still? It reports as "no conflicts" here:

image

domluna:
all good now looks like. Will merge when tests pass unless you want to add something else?

gustaphe:
I think I'm happy there. Will probably think of something else, but then I'll make new PRs. This one goes far enough IMO :smile:

Project.toml
```diff
@@ -1,19 +1,21 @@
 name = "JuliaFormatter"
 uuid = "98e50ef6-434e-11e9-1051-2b60c6c9e899"
 authors = ["Dominique Luna <dluna132@gmail.com>"]
-version = "1.0.12"
```
domluna, line 4: oh make this 1.0.13 instead for vscode. This doesn't change any prior behaviour so we don't need to increment the version to 1.1

domluna: I'm fine to do 1.1 but it helps the vscode folks to not bump the versions like this unless it's required.

gustaphe: Oh, ok. Normally only the leftmost non-zero number signifies breaking changes, and this is a feature add, but I don't know VSCode so... One moment.

PR #648, opened 2022-10-22
Parallelize over subpaths

MilesCranmer:
This adds basic parallelization over subpaths, which is useful when you need to repeatedly format a large package. Using my 8 core M1, my SymbolicRegression.jl package (~5000 SLOC) now only takes 80 ms seconds to format, down from 260 ms before this change.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "JuliaFormatter"
 uuid = "98e50ef6-434e-11e9-1051-2b60c6c9e899"
 authors = ["Dominique Luna <dluna132@gmail.com>"]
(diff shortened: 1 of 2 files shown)


domluna:
awesome! thanks so much for this

domluna:
can you increment the patch number? otherwise this is g2g

MilesCranmer:
Done

MilesCranmer:
Btw you might want to update this workflow https://github.com/julia-actions/julia-format/blob/a8502d9a6b40ef5da6e88721e253e79981aa26d1/workflows/format_check.yml#L32 with `--threads=auto`.

PR #650, opened 2022-11-02
Fix `join_lines_based_on_source` description

mroavi:
Swaps the descriptions for the `true` and `false` values for this option.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "JuliaFormatter"
 uuid = "98e50ef6-434e-11e9-1051-2b60c6c9e899"
 authors = ["Dominique Luna <dluna132@gmail.com>"]
-version = "1.0.13"
+version = "1.0.14"
 
 [deps]
 CSTParser = "00ebfdb7-1f24-5e51-bd34-a7502290713f"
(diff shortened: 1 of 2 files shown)


domluna:
The current description looks correct to me.

What's the confusion here?

mroavi:
It currently says:

> When `false` and the maximum margin is > than the length of "function foo(arg1, arg2, arg3)" this is formatted to
> ```julia
> function foo(arg1, arg2, arg3)
>    body
> end
> ```

According to me, It should be the opposite: when `true` the lines are joined, resulting in the formatted code above. I'm I wrong?

mroavi:
I'm using the default configuration, which means that this option is set to `false` (as stated in the docs).
If I format:
```julia
function foo(arg1,
                       arg2, arg3
                       )
       body
end
```
I obtain:
```julia
function foo(arg1,
  arg2, arg3
)
  body
end
```
which is not joining the lines, unlike in the example posted in the previous comment.

domluna:
what style are you using ? the option may be set as part of that style

mroavi:
Hmm, perhaps I misunderstood what is meant by `source`. What exactly does it mean in this context?

mroavi:
This PR doesn't make sense indeed. I'm closing it. But I still think that the intended meaning of the word "join" in this description is unclear. If you agree, I can open an issue.

PR #653, opened 2022-11-03
rebased fix of Issue316

### LINKED ISSUE #316: Replace `in` by `∈`

exaexa:
this should close #316 and is a newer version of #380.

I hope the rebase went well, still catching some minor other issues. Let me know in case I should clean up the history.

### DIFF
--- a/src/JuliaFormatter.jl
+++ b/src/JuliaFormatter.jl
@@ -607,6 +607,17 @@ end
 ->
 
 function func(...) where {TPARAM}
+
+### `for_in_replacement`
+
+Can be used when `always_for_in` is `true` to replace `in` with ∈ (\\in), or `=` instead:
+
+```julia
+for a = 1:10
+end
+
+# formatted with always_for_in = true, for_in_replacement = "∈"
+for a ∈ 1:10
 end
 ```
 
--- a/src/JuliaFormatter.jl
+++ b/src/JuliaFormatter.jl
@@ -613,7 +613,7 @@ end
 ### `for_in_replacement`
 
 Can be used when `always_for_in` is `true` to replace the default `in` with `∈` (`\\in`),
-or `=` instead:
+or `=` instead. The replacement options are `("in", "=", "∈")`.
 
 ```julia
 for a = 1:10
--- a/src/JuliaFormatter.jl
+++ b/src/JuliaFormatter.jl
@@ -156,6 +156,7 @@ normalize_line_ending(s::AbstractString, replacer = WINDOWS_TO_UNIX) = replace(s
         indent::Int = 4,
         margin::Int = 92,
         always_for_in::Union{Bool,Nothing} = false,
+        for_in_replacement::String = "in",
         whitespace_typedefs::Bool = false,
         whitespace_ops_in_indices::Bool = false,
         remove_extra_newlines::Bool = false,
@@ -611,7 +612,8 @@ end
 
 ### `for_in_replacement`
 
-Can be used when `always_for_in` is `true` to replace `in` with ∈ (\\in), or `=` instead:
+Can be used when `always_for_in` is `true` to replace the default `in` with `∈` (`\\in`),
+or `=` instead:
 
 ```julia
 for a = 1:10
--- a/src/JuliaFormatter.jl
+++ b/src/JuliaFormatter.jl
@@ -607,6 +607,7 @@ end
 ->
 
 function func(...) where {TPARAM}
+end
 
 ### `for_in_replacement`
 
--- a/src/JuliaFormatter.jl
+++ b/src/JuliaFormatter.jl
@@ -657,7 +657,7 @@ function format_text(text::AbstractString, style::AbstractStyle, opts::Options)
         @warn "Both `long_to_short_function_def` and `short_to_long_function_def` are set"
     end
     if opts.always_for_in && !valid_for_in_op(opts.for_in_replacement)
-        @warn "`for_in_replacement` is set to an invalid operator \"$(opts.for_in_replacement)\", valid operators are $(VALID_FOR_IN_OPERATORS). Change it to one of the valid operators then reformat."
+        @warn "`for_in_replacement` is set to an invalid operator \"$(opts.for_in_replacement)\", valid operators are $(VALID_FOR_IN_OPERATORS). Change it to one of the valid operators and then reformat."
         return text
     end
     cst, ps = CSTParser.parse(CSTParser.ParseState(text), true)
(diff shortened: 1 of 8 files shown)


exaexa:
@domluna I still have 4 tests failing for some reason I can't comprehend, the problem is with code like `[i for i = 1:10 if i == 2]` (array comprehension with `if`). 

One example is this:
```
always convert `=` to `in` (for loops): Test Failed at /home/exa/work/JuliaFormatter.jl/test/options.jl:253
  Expression: fmt(str_, always_for_in = true) == str
   Evaluated: "[i for i = 1:10 if i == 2]" == "[i for i in 1:10 if i == 2]"
Stacktrace:
 [1] macro expansion
   @ ~/work/julia-1.8.1/share/julia/stdlib/v1.8/Test/src/Test.jl:464 [inlined]
 [2] macro expansion
   @ ~/work/JuliaFormatter.jl/test/options.jl:253 [inlined]
 [3] macro expansion
   @ ~/work/julia-1.8.1/share/julia/stdlib/v1.8/Test/src/Test.jl:1357 [inlined]
 [4] macro expansion
   @ ~/work/JuliaFormatter.jl/test/options.jl:201 [inlined]
 [5] macro expansion
   @ ~/work/julia-1.8.1/share/julia/stdlib/v1.8/Test/src/Test.jl:1357 [inlined]
 [6] top-level scope
   @ ~/work/JuliaFormatter.jl/test/options.jl:2
```

I can't really spot anything that would have broken these tests; hints very welcome.

src/JuliaFormatter.jl
```diff
@@ -607,6 +607,17 @@ end
...
+### `for_in_replacement`
+
+Can be used when `always_for_in` is `true` to replace `in` with ∈ (\\in), or `=` instead:
+
+```julia
+for a = 1:10
+end
+
+# formatted with always_for_in = true, for_in_replacement = "∈"
+for a ∈ 1:10
```
domluna, line 620: looks like you're using the `end` for the function as the one for the for loop

`function func(...) where {TPARAM}`

needs an end keyword

src/fst.jl
```diff
@@ -1139,11 +1139,11 @@ function eq_to_in_normalization!(fst::FST, always_for_in::Bool)
             op.val = "="
             op.len = length(op.val)
         end
-    elseif fst.typ === Block || fst.typ === Brackets
+    elseif !is_leaf(fst)
```
domluna, line 1142: what's the reasoning for changing this?

src/fst.jl
```diff
@@ -1120,14 +1120,14 @@ for i = 1:10 body end
...
-function eq_to_in_normalization!(fst::FST, always_for_in::Bool)
+function eq_to_in_normalization!(fst::FST, always_for_in::Bool, for_in_replacement::String)
     if fst.typ === Binary
         idx = findfirst(n -> n.typ === OPERATOR, fst.nodes)
         idx === nothing && return
         op = fst[idx]
 
-        if always_for_in
-            op.val = "in"
+        if always_for_in && valid_for_in_op(op.val)
```
domluna, line 1129: the validation should occur when the options are initially set. And if the op is invalid it should display a message and not format.

domluna (changes requested):
Overall very nice, just a couple comments.

src/JuliaFormatter.jl
exaexa: ah, another rebase artifact

src/fst.jl
exaexa: Without this I couldn't make the `[for ... if ...]` syntax working. Also a content of the original PR:
https://github.com/domluna/JuliaFormatter.jl/pull/380/files#diff-fdbcedac9cc0bafcd2fa355f83f9558522f56e8e432e9a3775f59612dce9fe8aR1029

src/fst.jl
exaexa: Hm, this is taken from:
https://github.com/domluna/JuliaFormatter.jl/pull/380/files#diff-fdbcedac9cc0bafcd2fa355f83f9558522f56e8e432e9a3775f59612dce9fe8aR1016

I assumed that it validates the "incoming" operator (that we're not rewriting something that's not a cycle), but actually that doesn't make sense because that wouldn't ever get parsed. This condition doesn't make sense here then, will change shortly.

exaexa: OK this is gonna be harder, the validation was there because of this:
```
always convert `=` to `in` (for loops): Test Failed at /home/exa/work/JuliaFormatter.jl/test/options.jl:253
  Expression: fmt(str_, always_for_in = true) == str
   Evaluated: "[i for i in 1:10 if i in 2]" == "[i for i in 1:10 if i == 2]"
```
(i.e., it force-replaces the == operator even in the `if` part)

Any best way to dodge this? (I pushed the failing version now)

domluna: you shouldn't need this check the `valid_for_in_op(op.val)` part of `if always_for_in && valid_for_in_op(op.val)` anymore since it's being checked when the options are created.

exaexa: the `valid_for_in_op` check in the condition is gone (see the new commits); the problem is that it now mistakenly fires on the `if` part as described above.

src/options.jl
```diff
@@ -36,3 +37,10 @@ function needs_alignment(opts::Options)
         opts.align_pair_arrow ||
         opts.align_matrix
 end
+
+valid_for_in_op(s::String) = s in ("in", "=", "∈")
+
+function validate_options(opts::Options)
+    @assert valid_for_in_op(opts.for_in_replacement)
+    opts
+end
```
exaexa, line 46: @domluna btw I didn't find a good "central" place to validate the options (except perhaps for the place where styles are expanded, but that's very raw), so hooking this into the `State` constructor seemed like the best way. Please lmk in case there's a better place.

src/fst.jl
```diff
@@ -1120,30 +1120,29 @@ for i = 1:10 body end
...
             op.val = "in"
             op.len = length(op.val)
         elseif op.val == "in" && op_kind(fst[end]) === Tokens.COLON
             op.val = "="
             op.len = length(op.val)
         end
-    elseif fst.typ === Block || fst.typ === Brackets
-        for n in fst.nodes
-            eq_to_in_normalization!(n, always_for_in)
+    elseif fst.typ === Block || fst.typ === Brackets || fst.typ === Filter
```
domluna, line 1141: made this just Filter of !is_leaf since it's a more minimal change

domluna:
this should get the tests to pass now, I'll see if the initial validation check can be placed elsewhere but that might be ok how it is.

domluna:
ok I moved it to be a warning I'm happy with this now.

domluna (approved):

domluna (approved):

src/fst.jl
exaexa: ah great, yes, that should work pretty well

exaexa:
@domluna thanks a lot!

PR #674, opened 2023-01-10
Add `pre-commit` support

### LINKED ISSUE #498: Add support for pre-commit
Would be great to have julia format automatically run before commit https://pre-commit.com/#plugins.

jmuchovej:
Hacks together `pre-commit` support using the `./bin/format.jl` script in the repository root. Closes #498. (Hackish because `pre-commit` lacks direct support for Julia (WIP by me, but will take a bit).)

Some major quirks that seem to deviate (particularly from the `black` inspiration, and seemingly the `pre-commit` execution flow (of passing individual files)) is that passing individual files appears to:

1. ignore the local `.JuliaFormatter.toml`
2. require using the Julia REPL, like so: `julia -e "import JuliaFormatter: format_file; format_file(path)"`

(2) can probably be remedied with it's own script that mirrors `./bin/format.jl`, but I'm not quite sure how to remedy (1) without guidance on desired execution of `JuliaFormatter`.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,7 +1,7 @@
 - id: julia-formatter
   name: "Julia Formatter"
-  entry: "./bin/hook.jl"
-  pass_filenames: true 
-  # always_run: true
+  entry: "julia -e 'import JuliaFormatter: format; format(\".\")'"
+  pass_filenames: false
+  always_run: true
   language: "system"
   description: "An opinionated code formatter for Julia. Plot twist - the opinion is your own."
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,7 +1,7 @@
 - id: julia-formatter
   name: "Julia Formatter"
-  entry: "./bin/format.jl ."
-  pass_filenames: false
-  always_run: true
+  entry: "./bin/hook.jl"
+  pass_filenames: true 
+  # always_run: true
   language: "system"
   description: "An opinionated code formatter for Julia. Plot twist - the opinion is your own."
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,0 +1,7 @@
+- id: julia-formatter
+  name: "Julia Formatter"
+  entry: "./bin/format.jl ."
+  pass_filenames: false
+  always_run: true
+  language: "system"
+  description: "An opinionated code formatter for Julia. Plot twist - the opinion is your own."
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,6 @@
+# NOTE: Do NOT use this config in your own repositories. Instead, see 
+#   docs/integrations.md
+repos:
+- repo: local
+  hooks:
+  - id: "julia-formatter"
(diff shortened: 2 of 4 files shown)


jmuchovej:
Hmm. I may have spoken too soon about totally ignoring the local `.JuliaFormatter.toml`, but I'm still working out how to pass individual files to enable `JuliaFormatter` to run in "default way" `pre-commit` uses.

Looking through `tests`, `format_file` (and `format`) work correctly when `.JuliaFormatter.toml` is colocated with the passed file (e.g., `test/files/ignore/b.jl`), but not when the `.JuliaFormatter.toml` is higher up the directory tree (e.g., `test/files/ignore/ignored_directory/a.jl`).

The current behavior mostly touches far more files than it needs to (e.g., say you have a Julia and Python project colocated (let's ignore why for the moment), this `pre-commit` hook would traverse all of the Python files, too. Particularly because we don't have a way to filter on `julia` (and `jmd`, `qmd`, and `md`) files then pass them to the hook.

Though running the tests `cd tests/; julia --project=.. runtests.jl` returns all passing (minus the two broken cases), if I run `julia --project=. -e 'import JuliaFormatter: format; @assert format("test/files/ignore/ignored_directory/a.jl"; overwrite=false) == true'`, I end up with an `AssertionError`.

domluna:
not very familiar with pre-commits. if you call format it should pick up the config file correctly. So in your example above

julia -e "import JuliaFormatter: format; format(path)"

even if path is just 1 file it'll do what you expect

.pre-commit-hooks.yaml
```diff
@@ -0,0 +1,7 @@
+- id: julia-formatter
+  name: "Julia Formatter"
+  entry: "./bin/format.jl ."
```
domluna, line 3: this file shouldn't be used it's super old

jmuchovej: Ah, okay. Reverted to `julia -e ...`. 🙂

jmuchovej:
All that `pre-commit` does is provide a hook library that ties itself to the `pre-commit` step of `git commit [args...]` so that certain things can be ensured prior to actually committing (e.g., linting, formatting, notify when secrets are committed, etc.). The ability to ensure code formatting prior to commits was the main reason I wanted to open this PR, so that Julia code can make use of `pre-commit`s, particularly on larger projects.

Running `format(".")` from the `pre-commit` works as expected. The only reason explicitly passing a file matters is because `pre-commit` can find all relevant files _and_ passes them individually, but in the current condition `format` _must_ traverse the entire tree (which could become excessive in larger projects where Julia may only be a small part of it – if that's the desired mode for using `JuliaFormatter`, then I think it there should be a caveat noting that (e.g., in `docs/src/integrations.md` or similar).

domluna:
format_file just calls format anyway https://github.com/domluna/JuliaFormatter.jl/blob/master/src/JuliaFormatter.jl#L821-L835
If the argument is a path to a file, i.e. src/file1.jl it will only format that one file and not search any further

jmuchovej:
As far as I know it works. I tried it against the `JuliaFormatter` repo and my own (private) repo, both of which seemed to have expected results. 🙂

(I also saw, elsewhere, that there are a few other `pre-commit` hooks that traverse the whole tree, so it seems that it's not as divergent from the norm as I thought.)

domluna:
this is good to merge then yeah?

jmuchovej:
Just a heads-up, this would require a version bump for it to actually be usable by others. Up to you how to do said bumping. 🙂

domluna (approved):

Issue #609, opened 2022-06-18
inline comments in function calls removed for SciMLStyle


visr:
```julia
t = """
# https://github.com/SciML/SciMLStyle#comments
# In general, comments above a line of code or function are preferred to inline comments.
f(; q) = q
f(
    q = 2  # this comment will be removed
)
"""

format_text(t, style=SciMLStyle()) |> Text
```

Returns
```julia
# https://github.com/SciML/SciMLStyle#comments
# In general, comments above a line of code or function are preferred to inline comments.
f(; q) = q
f(q = 2)
```

As noted the style guide favors commands above, but I assume the formatter is not supposed to remove them?
#513 and #562 seem similar, but not the same.

For the example of #513 only the last comment gets removed:

```julia
f(10 # i got removed!
  *
  10 # me too!
  +
  10)
```

domluna:
interesting corner case since SciMLStyle uses YASStyle for a lot of defaults and YASStyle doesn't allow opening/closing parenthesis to be on their own line.

The simplest fix which should still adhere to the style is to put the comment after the closing parenthesis

visr:
This seems to be fixed, now the input string is kept as-is.

PR #676, opened 2023-01-17
Add option `variable_dict_indent`

### LINKED ISSUE #675: Format `Dict` definitions like `Vector` definitions
We're using the SciML style guide since we found that it's very consistent with our "in-house" style.
However, we would like to format `Dict` definitions similar to `Vector` definitions like this:
```julia
x = [
    "some arbitrary string bla bla bla bla bla bla bla bla bla",
    "another longer arbitrary string bla bla bla bla bla bla bla bla",
]

SomeLongerTypeThanJustString = String
y = Dict{Int, SomeLongerTypeThanJustString}(
    1 => "some arbitrary string bla bla bla bla bla bla bla bla bla",
    2 => "another longer arbitrary string bla bla bla bla bla bla bla bla"
)
```
After auto-formatting, however, it looks like this:
```julia
y = Dict{Int, SomeLongerTypeThanJustString}(1 => "some arbitrary string bla bla bla bla bla bla bla bla bla",
                                            2 => "another longer arbitrary string bla bla bla bla bla bla bla bla")
```
This also goes way beyond the 92-character limit.
Is there a way to redefine this behavior without changing the rest of the formatting rules?

efaulhaber:
Closes #675.

Right now, the `SciMLStyle` only allows `Dict` definitions of the form
```julia
Dict{Int, Int}(1 => 2,
               3 => 4)
```
Especially for longer definitions, this produces IMHO undesirable results like
```julia
SomeLongerTypeThanJustString = String
y = Dict{Int, SomeLongerTypeThanJustString}(1 => "some arbitrary string bla bla b  sdafasdfasdf dsa",
                                            2 => "another longer arbitrary string bla bla bla bla bla bla bla bla")
```
and it doesn't allow a line break after the opening parenthesis.

I added a new option `variable-dict-indent` (working title, suggestions welcome) to allow all these versions:
```julia
Dict{Int, Int}(1 => 2,
               3 => 4)

Dict{Int, Int}(
    1 => 2,
    3 => 4)

Dict{Int, Int}(
    1 => 2,
    3 => 4,
)
```
The long definition above is automatically changed to
```julia
SomeLongerTypeThanJustString = String
y = Dict{Int, SomeLongerTypeThanJustString}(
    1 => "some arbitrary string bla bla b  sdafasdfasdf dsa",
    2 => "another longer arbitrary string bla bla bla bla bla bla bla bla",
)
```
It also allows comments:
```julia
Dict{Int, Int}(
    # Comment
    1 => 2,
    3 => 4)

Dict{Int, Int}(# Comment
    1 => 2,
    3 => 4)

Dict{Int, Int}( # Comment
    # Comment
    1 => 2,
    # Another comment
    3 => 4)
```
Note that in the second example, the formatter would eat the comment with `variable_dict_indent = false` (#609).


docs/src/sciml_style.md
```diff
@@ -30,3 +30,28 @@ Or to use `SciMLStyle` except change one of the settings:
 ```julia
 format("file.jl", SciMLStyle(), remove_extra_newlines=false)
 ```
+
+## Additional Options
+
+The `SciMLStyle` supports the additional option `variable_dict_indent`,
+which is set to `false` by default.
+It allows `Dict` definitions without aligning to the opening parenthesis:
```
domluna, line 38: since you've added this functionality to the YAS functions, which SciML fallbacks too we should mention this works for YAS as well

domluna (commented):
Overall I'm ok with this. I'm wondering whether this can be generalized. Maybe be able to list the names you want to be indented in this way?

So in the toml file this would be

```
variable_call_indent = ["Dict"]
```

not sure about the option naming

efaulhaber:
Done. I couldn't come up with a better name for the option.

efaulhaber:
Any news on this?

domluna:
Can you run JET on this branch

https://github.com/domluna/JuliaFormatter.jl/pull/685

domluna (approved):

src/fst.jl
```diff
@@ -1158,3 +1158,15 @@ function eq_to_in_normalization!(fst::FST, always_for_in::Bool, for_in_replaceme
...
+# Note that this also works for CSTParser.EXPR
+function caller_in_list(fst, list)
+    if is_leaf(fst[1]) && fst[1].val in list
+        return true
+    elseif !is_leaf(fst[1]) && is_leaf(fst[1][1]) && fst[1][1].val in list
+        return true
+    end
+
+    return false
+end
```
efaulhaber, line 1172:
```suggestion
function caller_in_list(fst, list::AbstractArray{String})
    if is_leaf(fst[1]) && fst[1].val::String in list
        return true
    elseif !is_leaf(fst[1]) && is_leaf(fst[1][1]) && fst[1][1].val::String in list
        return true
    end

    return false
end
```
`list` might contain `missing` (at least that's what the compiler thinks). Explicitly requiring a vector of a non-`Missing` type solves this warning. Same for the other two annotations telling the compiler that this must not be `missing`. However, now the user can't pass a tuple or other iterable to `variable_dict_indent`. Is this the correct fix for this warning? Should we allow `NTuple{N, String} where N` here as well?

src/styles/yas/nest.jl
```diff
@@ -1,6 +1,24 @@
 function n_call!(ys::YASStyle, fst::FST, s::State)
     style = getstyle(ys)
 
+    # With `variable_call_indent`, check if the caller is in the list
+    if caller_in_list(fst, s.opts.variable_call_indent) && length(fst.nodes) > 5
```
efaulhaber, line 5:
```suggestion
    if caller_in_list(fst, s.opts.variable_call_indent) && length(fst.nodes::AbstractArray) > 5
```
This solves the last remaining warning.

efaulhaber (commented):
I ran JET and got these possible errors (plus the 17 that I get on master):
```
┌ @ /mnt/ssd/home/efaulhab/.julia/dev/JuliaFormatter/src/styles/yas/pretty.jl:255 JuliaFormatter.caller_in_list(cst, s.opts.variable_call_indent)
│┌ @ /mnt/ssd/home/efaulhab/.julia/dev/JuliaFormatter/src/fst.jl:1165 goto %9 if not JuliaFormatter.in(fst[1].val, list)
││ non-boolean `Missing` found in boolean context (1/2 union split): goto %9 if not JuliaFormatter.in(((fst::CSTParser.EXPR)[1]::Any).val::Any, list::Vector{String})::Union{Missing, Bool}
│└────────────────────────────────────────────────────────────────────
│┌ @ /mnt/ssd/home/efaulhab/.julia/dev/JuliaFormatter/src/fst.jl:1167 goto %23 if not JuliaFormatter.in(fst[1][1].val, list)
││ non-boolean `Missing` found in boolean context (1/2 union split): goto %23 if not JuliaFormatter.in((((fst::CSTParser.EXPR)[1]::Any)[1]::Any).val::Any, list::Vector{String})::Union{Missing, Bool}
│└────────────────────────────────────────────────────────────────────
┌ @ /mnt/ssd/home/efaulhab/.julia/dev/JuliaFormatter/src/styles/yas/nest.jl:5 JuliaFormatter.caller_in_list(fst, s.opts.variable_call_indent)
│┌ @ /mnt/ssd/home/efaulhab/.julia/dev/JuliaFormatter/src/fst.jl:1165 goto %9 if not JuliaFormatter.in(fst[1].val, list)
││ non-boolean `Missing` found in boolean context (1/2 union split): goto %9 if not JuliaFormatter.in(((fst::JuliaFormatter.FST)[1]::JuliaFormatter.FST).val::Union{Nothing, AbstractString}, list::Vector{String})::Union{Missing, Bool}
│└────────────────────────────────────────────────────────────────────
│┌ @ /mnt/ssd/home/efaulhab/.julia/dev/JuliaFormatter/src/fst.jl:1167 goto %23 if not JuliaFormatter.in(fst[1][1].val, list)
││ non-boolean `Missing` found in boolean context (1/2 union split): goto %23 if not JuliaFormatter.in((((fst::JuliaFormatter.FST)[1]::JuliaFormatter.FST)[1]::JuliaFormatter.FST).val::Union{Nothing, AbstractString}, list::Vector{String})::Union{Missing, Bool}
│└────────────────────────────────────────────────────────────────────
┌ @ /mnt/ssd/home/efaulhab/.julia/dev/JuliaFormatter/src/styles/yas/nest.jl:5 JuliaFormatter.length(fst.nodes)
│ no matching method found `length(::Nothing)` (1/2 union split): JuliaFormatter.length((fst::JuliaFormatter.FST).nodes::Union{Nothing, Vector{JuliaFormatter.FST}})
└─────────────────────────────────────────────────────────────────────────────
```
I am not sure how to fix these (and if we should fix them at all). I've added comments to the relevant code, which solve the warnings. Should I leave the code as it is or implement these changes?

PR #677, opened 2023-01-20
Use SnoopPrecompile.

chriselrod:
With this PR:
```julia
julia> @time using JuliaFormatter
[ Info: Precompiling JuliaFormatter [98e50ef6-434e-11e9-1051-2b60c6c9e899]
 33.716529 seconds (2.11 M allocations: 124.493 MiB, 0.13% gc time, 0.11% compilation time)

julia> @time @eval JuliaFormatter.format("/home/chriselrod/.julia/dev/Octavian/")
  1.131611 seconds (1.63 M allocations: 103.127 MiB, 887.14% compilation time)
true
```

Versus the master branch:
```julia
julia> @time using JuliaFormatter
[ Info: Precompiling JuliaFormatter [98e50ef6-434e-11e9-1051-2b60c6c9e899]
  9.866693 seconds (1.75 M allocations: 102.886 MiB, 0.42% gc time, 0.37% compilation time)

julia> @time @eval JuliaFormatter.format("/home/chriselrod/.julia/dev/Octavian/")
  8.763866 seconds (8.31 M allocations: 543.450 MiB, 1.22% gc time, 1102.95% compilation time)
true
```

This is on Julia master.

For me, the increase in precompile time is well worth the benefit to time-to-first-format, but I understand how some may disagree.
I use JuliaFormatter a lot -- I set my text editor to run the formatter every time I save a Julia file.
So this would save a decent bit of time for me.
Of course, some people may use `LanguageServer.jl`, which depends on JuliaFormatter, without ever running format...

### DIFF
--- a/src/other/precompile.jl
+++ b/src/other/precompile.jl
@@ -7,3 +7,10 @@ function _precompile_()
     Base.precompile(Tuple{typeof(pretty),BlueStyle,CSTParser.EXPR,State})
     Base.precompile(Tuple{typeof(pretty),MinimalStyle,CSTParser.EXPR,State})
 end
+using SnoopPrecompile
+@precompile_setup begin
+    dir = joinpath(@__DIR__,"../..")
+    @precompile_all_calls begin
+        format(dir)
+    end
+end
--- a/src/other/precompile.jl
+++ b/src/other/precompile.jl
@@ -8,7 +8,7 @@ using SnoopPrecompile
     """
     @precompile_all_calls begin
         format(dir)
-        for style = [DefaultStyle(), BlueStyle(), SciMLStyle(), YASStyle()]
+        for style = [DefaultStyle(), BlueStyle(), SciMLStyle(), YASStyle(), MinimalStyle()]
           format_text(str, style)
         end
     end
--- a/src/other/precompile.jl
+++ b/src/other/precompile.jl
@@ -1,16 +1,15 @@
 #! format: off
-function _precompile_()
-    ccall(:jl_generating_output, Cint, ()) == 1 || return nothing
-    # pretty
-    Base.precompile(Tuple{typeof(pretty),DefaultStyle,CSTParser.EXPR,State})
-    Base.precompile(Tuple{typeof(pretty),YASStyle,CSTParser.EXPR,State})
-    Base.precompile(Tuple{typeof(pretty),BlueStyle,CSTParser.EXPR,State})
-    Base.precompile(Tuple{typeof(pretty),MinimalStyle,CSTParser.EXPR,State})
-end
 using SnoopPrecompile
 @precompile_setup begin
     dir = joinpath(@__DIR__,"..", "..")
+    str = raw"""
+       @noinline require_complete(m::Matching) =
+           m.inv_match === nothing && throw(ArgumentError("Backwards matching not defined. `complete` the matching first."))
+    """
     @precompile_all_calls begin
         format(dir)
+        for style = [DefaultStyle(), BlueStyle(), SciMLStyle(), YASStyle()]
+          format_text(str, style)
+        end
     end
 end
--- a/src/other/precompile.jl
+++ b/src/other/precompile.jl
@@ -9,7 +9,7 @@ function _precompile_()
 end
 using SnoopPrecompile
 @precompile_setup begin
-    dir = joinpath(@__DIR__,"../..")
+    dir = joinpath(@__DIR__,"..", "..")
     @precompile_all_calls begin
         format(dir)
     end
--- a/Project.toml
+++ b/Project.toml
@@ -9,6 +9,7 @@ CommonMark = "a80b9123-70ca-4bc0-993e-6e3bcb318db6"
 DataStructures = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
 Glob = "c27321d9-0574-5035-807b-f59d2c89b15c"
 Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
+SnoopPrecompile = "66db9d55-30c0-4569-8b51-7e840670fc0c"
 Tokenize = "0796e94c-ce3b-5d07-9a54-7f471281c624"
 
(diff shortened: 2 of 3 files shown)


src/other/precompile.jl
```diff
@@ -7,3 +7,10 @@ function _precompile_()
     Base.precompile(Tuple{typeof(pretty),BlueStyle,CSTParser.EXPR,State})
     Base.precompile(Tuple{typeof(pretty),MinimalStyle,CSTParser.EXPR,State})
 end
+using SnoopPrecompile
+@precompile_setup begin
+    dir = joinpath(@__DIR__,"../..")
```
oscardssmith, line 12:
```suggestion
    dir = joinpath(@__DIR__,"..", "..")
```

src/other/precompile.jl
```diff
@@ -7,3 +7,10 @@ function _precompile_()
     Base.precompile(Tuple{typeof(pretty),BlueStyle,CSTParser.EXPR,State})
     Base.precompile(Tuple{typeof(pretty),MinimalStyle,CSTParser.EXPR,State})
 end
+using SnoopPrecompile
+@precompile_setup begin
+    dir = joinpath(@__DIR__,"..", "..")
+    @precompile_all_calls begin
+        format(dir)
```
domluna, line 14: would it be more beneficial to run the entire test suite in here? since `format(dir)` would be a subset of the calls in the test suite

domluna (commented):
I think this is awesome. I've been looking to invalidations and things of that nature. From what I can tell no invalidations stem from the package itself but there are some calls which are not inferred during precompilation https://timholy.github.io/SnoopCompile.jl/dev/snoopi_deep_analysis/#Trigger-trees but I'm not entirely sure how to fix those.

@davidanthoff any big reservations about the time increase in `using JuliaFormatter` here for does that not matter since the vscode extension is compiled anyway

chriselrod:
```julia
julia> @time using JuliaFormatter
[ Info: Precompiling JuliaFormatter [98e50ef6-434e-11e9-1051-2b60c6c9e899]
 30.143423 seconds (1.91 M allocations: 108.742 MiB, 0.06% gc time, 0.10% compilation time)

julia> @time format("/home/chriselrod/.julia/dev/Octavian/")
  0.452070 seconds (785.68 k allocations: 47.339 MiB, 699.16% compilation time)
false
```
This is a slightly faster computer in terms of single threaded performance han before.
For the latest release, I have
```julia
julia> @time format("/home/chriselrod/.julia/dev/Octavian/")
  6.105013 seconds (3.36 M allocations: 213.737 MiB, 0.36% gc time, 769.80% compilation time)
true
```
This is still a rather large improvement vs before.

src/other/precompile.jl
```diff
@@ -7,3 +7,10 @@ function _precompile_()
     Base.precompile(Tuple{typeof(pretty),BlueStyle,CSTParser.EXPR,State})
     Base.precompile(Tuple{typeof(pretty),MinimalStyle,CSTParser.EXPR,State})
 end
```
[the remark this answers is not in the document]
chriselrod: I got rid of these.

src/other/precompile.jl
chriselrod: I added a `format_text` with a few different styles.

I haven't looked through the code extensively, but I wonder if it'd be better to handle styles via [Unityper](https://github.com/YingboMa/Unityper.jl), [Expronicon](https://discourse.julialang.org/t/ann-expronicon-v0-8-type-stable-rust-like-algebra-data-type-pretty-printing-julia-expr-more/92647), or manual `@enum`s?
Manual `@enum`s might be the easiest if you aren't doing many branches or holding information in the structs.

chriselrod:
I added `max_methods = 1` which doesn't impact type stable code, but stops excessive inference when union splitting.

src/other/precompile.jl
```diff
@@ -1,9 +1,15 @@
...
+using SnoopPrecompile
+@precompile_setup begin
+    dir = joinpath(@__DIR__,"..", "..")
+    str = raw"""
+       @noinline require_complete(m::Matching) =
+           m.inv_match === nothing && throw(ArgumentError("Backwards matching not defined. `complete` the matching first."))
+    """
+    @precompile_all_calls begin
+        format(dir)
+        for style = [DefaultStyle(), BlueStyle(), SciMLStyle(), YASStyle()]
```
domluna, line 11:
```suggestion
        for style = [DefaultStyle(), BlueStyle(), SciMLStyle(), YASStyle(), MinimalStyle()]
```

domluna (approved):
Awesome!

One last thing regarding development with Revise. Would this increase load times when editing? If so is there a way I can disable this besides commenting out the file?

src/other/precompile.jl
domluna: information is held in the structs so I don't enum is viable but perhaps one of the of the other options

chriselrod:
> One last thing regarding development with Revise. Would this increase load times when editing? If so is there a way I can disable this besides commenting out the file?

Precompile times will increase.
Revise won't get any slower.

Aside from commenting out the file or commenting out the include, we could copy OrdinaryDiffEq and use `Preferences` to optionally turn off precompilation:
https://github.com/SciML/OrdinaryDiffEq.jl/blob/2a14472aef321ca0cd599f56e6ed87053f8c9386/src/OrdinaryDiffEq.jl#L306

domluna:
ok sounds good ready to merge this then

davidanthoff:
This is now preventing us from shipping the latest version of JuliaFormatter with the VS Code extension. First, adding any new dependency to the LS is always painful, and second, we don't support shipping packages that are located in a mono-repo at the moment, like SnoopPrecompile. So, for now, we'll have to pin JuliaFormatter (and CommonMark) at old versions that don't have SnoopPrecompile as deps.

For JuliaFormatter that is 1.0.18. There is a 1.0.19 in the registry, but it doesn't have a tag here in the repo, and we can only ship versions with a tag.

Issue #707, opened 2023-04-19
"no indent rule" also for other blocks than modules


ranocha:
Would it be possible to have a no-indent-rule also for other top level blocks besides `module`s?

Our use case in Trixi.jl is something like this

```julia
@muladd begin

# lots of stuff

end # @muladd
```

A lot of our files look like this. Thus, JuliaFormatter.jl basically indents all of our files - which we would like to avoid.

CC @sloede 
Xref https://github.com/trixi-framework/Trixi.jl/pull/1403#issuecomment-1514697345

domluna:
https://github.com/domluna/JuliaFormatter.jl/pull/724

Issue #711, opened 2023-04-27
Avoid indentation within `@testitem` blocks


gdalle:
The new [test item framework](https://discourse.julialang.org/t/prerelease-of-new-testing-framework-and-test-run-ui-in-vs-code/86355) in VSCode allows you to specify standalone tests that can be executed independently. However, since they are typically pretty long, it would be nice to avoid indenting them (as is done eg. for `module` blocks). What do you think? Is that a difficult change?

domluna:
Probably similar effort to https://github.com/domluna/JuliaFormatter.jl/pull/489/files

domluna:
similar to https://github.com/domluna/JuliaFormatter.jl/issues/707

generalizing the module indent option to user patterns might be worth it

domluna:
maybe adding a comment so something along the lines of

```
@muladd begin
# fmt: dedent
code here is not indented
end
```

i think this should be fairly easy to add since in print.jl you can check if the comment has that before printing the indent

sloede:
> i think this should be fairly easy to add since in print.jl you can check if the comment has that before printing the indent

This sounds like an interesting and comparably portable approach. However, I am not sure into which function I would have to look - `print_inlinecomment` does sound correct, but it is not clear to me how and where the indent size is modified.

domluna:
https://github.com/domluna/JuliaFormatter.jl/blob/master/src/print.jl#L49-L97

the check part is

     ws = repeat(" ", max(indent, 0))

i think adding another keyword argument would be sufficient. defaults to false and would be true if the NOTCODE node contains the string we're looking for

format: deindent

or "noindent" ?

ehh we can bikeshed that later

domluna:
occursin(r"^#!\s*format\s*:\s*deindent\s*$", n.val) would capture it

domluna:
I'll tackle this tonight

sloede:
> I'll tackle this tonight

That would be fantastic, thanks a lot 🙏

FWIW, I think from a *user* perspective, `noindent` is clearer, since what the user (presumably) thinks is "I put this here to get no indentation, ie, noindent". But any keyword is fine, really.

domluna:
@sloede this branch should do the trick https://github.com/domluna/JuliaFormatter.jl/pull/724

domluna:
this is possible now with #724

gdalle:
thanks!

PR #724, opened 2023-05-21
Functionality to unindent or noindent a section of code

domluna:
Done with

```
            #! format: noindent

```

Example

```

        begin
            @muladd begin
            #! format: noindent
            # dawdawdaw comment
            a = 10
            b = 20
            begin
                # another inent
                z = 33
            end

            a * b
            end
        end
```

solves https://github.com/domluna/JuliaFormatter.jl/issues/711, #707

### DIFF
--- a/src/document.jl
+++ b/src/document.jl
@@ -48,6 +48,10 @@ struct Document
     # the start and end lines of regions in the
     # file formatting should be skipped.
     format_skips::Vector{Tuple{Int,Int,String}}
+
+    # Line where there's a noindent comment. "#! format: noindent". This is checked during
+    # the print stage to know if the contents of the block (recursive) should not be indented.
+    noindent_blocks::Vector{Int}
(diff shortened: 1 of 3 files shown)


sloede:
Thanks a lot @domluna! Currently testing this and then we'll get back to you, but first trials look promising!

ranocha:
Thanks a lot! This seems to work fine - at least in the cases I have looked at so far.

domluna:
@ranocha let me if you think this is good to merge

ranocha:
This is good to merge from my tests. Thanks a lot!

sloede:
@domluna Thanks a lot for responding to this feature request, this looks very good and can imho be merged! Also, in addition to #711, I think this will also resolve #707 (at least for the use case that we have).

gdalle:
Awesome!

Issue #753, opened 2023-09-06
[SciML]: Long variable definitions turn into functions


ven-k:
With SciML formatting style, `x(t) = value, [...]` turns into `function x(t)` with value as its body whenever the definition is longer than 94 chars.

MWE: 
```julia
using ModelingToolkit
@mtkmodel A begin
	@variables begin
		i(t) = 0.0, [description = "Line longer than the char limit of 94 characters", unit = u"A"]
	end
end
```

Turns into 
```julia
@mtkmodel A begin
    @variables begin
        function i(t)
            0.0,
            [description = "Line longer than the char limit of 94 characters", unit = u"A"]
        end
    end
end
```

domluna:
is this problematic because it's not semantically the same?

domluna:
this occurs because YASStyle has short_to_long_function_def as true and it's used quite extensively in SciMLStyle

https://github.com/domluna/JuliaFormatter.jl/blob/master/src/styles/sciml/pretty.jl#L84-L91

@YingboMa would you guys want to have different defaults than what comes with https://github.com/domluna/JuliaFormatter.jl/blob/master/src/styles/yas/pretty.jl#L24? Would it make sense to override it with the options set by SciML?

ven-k:
> is this problematic because it's not semantically the same?

Yes. `i(t) = 0` is parsed as `i` variable dependent on `t` with default value 0.

ChrisRackauckas:
> is this problematic because it's not semantically the same?

It's unintended in the style and not something we've wanted. I don't know how to revert it. We have disabled using the formatter until it's fixed.

domluna:
actually the options are consistent already since they are only checked once when format_text is initially called and then saved.

this is what's causing this particular transform

```
function n_binaryopcall!(ss::SciMLStyle, fst::FST, s::State; indent::Int = -1)
    style = getstyle(ss)
    line_margin = s.line_offset + length(fst) + fst.extra_margin
    if line_margin > s.opts.margin &&
       fst.ref !== nothing &&
       CSTParser.defines_function(fst.ref[])
        transformed = short_to_long_function_def!(fst, s)
        transformed && nest!(style, fst, s)
    end

    if findfirst(n -> n.typ === PLACEHOLDER, fst.nodes) !== nothing
        n_binaryopcall!(DefaultStyle(style), fst, s; indent = indent)
        return
    end

    start_line_offset = s.line_offset
[...]

PR #754, opened 2023-09-08
fix for #753

domluna:


src/styles/sciml/nest.jl
```diff
@@ -38,7 +38,14 @@ function n_binaryopcall!(ss::SciMLStyle, fst::FST, s::State; indent::Int = -1)
     line_margin = s.line_offset + length(fst) + fst.extra_margin
     if line_margin > s.opts.margin &&
        fst.ref !== nothing &&
-       CSTParser.defines_function(fst.ref[])
+       (
```
domluna, line 41: @ChrisRackauckas should another check for `short_to_long_function_def` be added here or is it intentional to not check?

domluna: the check is not in the pass itself so all that's left is verify that we can to make try transform always, even if `short_to_long_function_def` is set to `false`

ChrisRackauckas: > should another check for short_to_long_function_def be added here or is it intentional to not check?

That I don't know, @YingboMa wrote that I think?

Issue #759, opened 2023-09-21
Short circuiting idiom to `if` conversion


chriselrod:
It would be great if we could have an option to automatically convert
```julia
b0 && foo()
b1 || bar()
```
into
```julia
if b0
    foo()
end
if !b1
    bar()
end
```
This is a style that is often preferred for being able to tell whether a branch is taken in code coverage, so it'd be great if it can be enforced.

domluna:
i think this would be a good option to have

ericphanson:
xref https://github.com/domluna/JuliaFormatter.jl/issues/382

PR #765, opened 2023-10-05
Option to transform short circuit to if condition

domluna:
implements what's discussed in #759 

cc @chriselrod if you could try this out that would be great


Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "JuliaFormatter"
 uuid = "98e50ef6-434e-11e9-1051-2b60c6c9e899"
 authors = ["Dominique Luna <dluna132@gmail.com>"]
-version = "1.0.39"
+version = "1.0.40"
```
hyrodium, line 4: This PR adds a new option `short_circuit_to_if`, so it would be better to bump the minor version.

```suggestion
version = "1.1.0"
```

hyrodium (commented):
Could you please update `docs/src/index.md` to list the new option?

Project.toml
domluna: by default it's off so from the user perspective there's no change implying there's not really a need to bump it. Also there's this compatibility thing with VSCode where bumping the minor version frequently leads to things being out of sync so it's easier to just bump the patch version.

hyrodium: >by default it's off so from the user perspective there's no change implying there's not really a need to bump it.

Hmm, Julia packages should respect semantic versioning, so I believe bumping the patch version should imply a bug fix. (cf. [Pkg.jl docs](https://pkgdocs.julialang.org/v1/compatibility/#Version-specifier-format), [semver docs](https://semver.org/#spec-item-6))

>Also there's this compatibility thing with VSCode where bumping the minor version frequently leads to things being out of sync so it's easier to just bump the patch version.

What is the reason behind this? I couldn't find the `[compat]` table in `Project.toml` files in the [julia-vscode](https://github.com/julia-vscode/julia-vscode) repository.

chriselrod:
Just tried it, and works great!
```julia
454                 (1 <= a <= j) || error("oops")    488                 if !((1 <= a <= j))
...                                                   489                     error("oops")
...                                                   490                 end
455
```

Only possible suggestion (other than documenting it), is in case of `||`, maybe check if the condition is already wrapped in parenthesis so we avoid double-wrapping?
Not that big of a deal, though.
Happy to see this.

Project.toml
hyrodium: >Also there's this compatibility thing with VSCode where bumping the minor version frequently leads to things being out of sync so it's easier to just bump the patch version.

Is it because of this (https://github.com/julia-vscode/julia-vscode/issues/3226#issuecomment-1451803980)? If so, the problem about the JuliaFormatter's version in vscode is not related to bumping patch or minor version.

PR #785, opened 2023-11-28
Fix ugly line breaks in curly braces outside of page margin when using `SciMLStyle` with `yas_style_nesting=true`

efaulhaber:
On master, this
```julia
xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx::SomePrettyLongTypeName{Foo}
```
will be reformatted to this:
```julia
xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx::SomePrettyLongTypeName{
                                                                                            Foo
                                                                                            }
```

I modified `p_curly` of `YASStyle` to not always insert a `Placeholder(0)` but add a `Placeholder(1)` when `whitespace_typedefs=true`. This is identical to what is done in `DefaultStyle`. With this change, I can use `p_curly` of `YASStyle` for `SciMLStyle` as well when `yas_style_nesting=true`.

Here is a demonstration of what happens in a large project when reformatted with this PR:
https://github.com/trixi-framework/Trixi.jl/pull/1759

IMHO, many things get prettier, and everything else gets, well, at least not uglier `¯\_(ツ)_/¯`
The only thing that gets slightly worse IMO is the file `src/solvers/dgmulti/sbp.jl`, but that wasn't pretty to begin with.

PR #786, opened 2023-11-29
Hotfix for #785

efaulhaber:
I don't know why this wasn't failing the tests in #785, but I think this is what's failing https://github.com/JuliaRegistries/General/pull/96136.

### DIFF
--- a/test/sciml_style.jl
+++ b/test/sciml_style.jl
@@ -372,5 +372,5 @@
     # xxxxx::SomePrettyLongTypeName{
     #                               Foo
     #                              }
-    @test format_text(str, SciMLStyle(), yas_style_nesting=true) == str
+    @test format_text(str, SciMLStyle(), yas_style_nesting = true) == str
 end


domluna:
there is no format check in the CI so I don't know why that would cause registration to fail. either way we'll see if this works

efaulhaber:
```
Failed to precompile JuliaFormatter [98e50ef6-434e-11e9-1051-2b60c6c9e899] to "/tmp/jl_dBrbRZ/compiled/v1.9/JuliaFormatter/jl_q3BnuY".
[ Info: Error in formatting file /tmp/jl_dBrbRZ/packages/JuliaFormatter/O48b1/test/sciml_style.jl
ERROR: LoadError: TaskFailedException

    nested task error: TaskFailedException
    
        nested task error: SystemError: opening file "/tmp/jl_dBrbRZ/packages/JuliaFormatter/O48b1/test/sciml_style.jl": Permission denied
```
I have absolutely no idea why it can't read this file, so my only guess is that it might be related to
```
[ Info: Error in formatting file /tmp/jl_dBrbRZ/packages/JuliaFormatter/O48b1/test/sciml_style.jl
```

domluna:
maybe it's that the file gets modified during the test due to the format call. but the format test should have failed then so that's a bit weird

domluna:
running locally it said

julia> format(".")
true

which is incorrect since there was a diff

λ ~/.julia/dev/JuliaFormatter: git diff
diff --git a/test/sciml_style.jl b/test/sciml_style.jl
index 3fe09c60..3d89595b 100644
--- a/test/sciml_style.jl
+++ b/test/sciml_style.jl
@@ -372,5 +372,5 @@
     # xxxxx::SomePrettyLongTypeName{
     #                               Foo
     #                              }
-    @test format_text(str, SciMLStyle(), yas_style_nesting=true) == str
+    @test format_text(str, SciMLStyle(), yas_style_nesting = true) == str
 end

efaulhaber:
That explains why I didn't catch it. But shouldn't `format(".")` return `true` if and only if there is no diff? I rely on this return value in the CI checks in my code.

domluna:
yeah this is the only time I've seen it fail

domluna:
with -t auto (threads) it seems to work as well

PR #792, opened 2023-12-16
DRAFT: Optimal-ish nesting for multilines for SciML

domluna:
Update: ready for merging!

What this PR does:

Uses the same style for pretty/nesting. You can use different styles for different parts of the format tree but that style has to be consistent for the pretty and nesting stage for that part.

For nesting for iterable-like types (tuples, function calls, etc) it splits the line based on a "best-fit" heuristic. This is a iteration which will end up at nesting all lines if no suitable nest solution prior to that is found. But it goes something like this: Can i split this into two lines? If so which of these splits minimizes the length between those lines. So for example.

arg1, arg2, arg3, arg4

then

arg1, arg2
arg3, arg4

would be chosen over

arg1,
arg2,arg3,arg4

If no split is found for two lines which fits into the margin, then we try got 3 lines and so on.

Use `unnest!` for cleanup.

Sample string:

```julia
function (p::CoevolveJumpAggregation{T,S,F1,F2})(integrator::AbstractSSAIntegrator) where {T, S, F1, F2 <: Union{Tuple, Nothing}}
    body
end

```

this PR

```julia
function (p::CoevolveJumpAggregation{T, S, F1, F2})(integrator::AbstractSSAIntegrator,
) where {T, S, F1, F2 <: Union{Tuple, Nothing}}
    body
end

```

where cannot be a separate from ) which is why we get this kind of split. The trailing comma there is not needed so that's a bug in un-nesting.

cc @YingboMa @isaacsas


domluna:
[diff.txt](https://github.com/domluna/JuliaFormatter.jl/files/13691883/diff.txt)

diff for https://github.com/isaacsas/DiffEqJump.jl/blob/master/src/aggregators/coevolve.jl

isaacsas:
@domluna thanks for taking a pass. I know this is WIP. If it helps, I left some examples of issues I'm seeing on a new JumpProcesses PR (I wanted to see where things are at with this PR). https://github.com/SciML/JumpProcesses.jl/pull/383

domluna:
> @domluna thanks for taking a pass. I know this is WIP. If it helps, I left some examples of issues I'm seeing on a new JumpProcesses PR (I wanted to see where things are at with this PR). [SciML/JumpProcesses.jl#383](https://github.com/SciML/JumpProcesses.jl/pull/383)

thanks I'll take a look. I expect there are kinks in this still so yeah WIP but I think the idea of it is good.

domluna:
cc @ericphanson the style of nesting might be preferable for YAS as well since we currently do a greedy approach for that atm.

domluna:
```julia
┌ Info:
│   segments =
│    1-element Vector{Tuple{Int64, Int64}}:
└     (1, 2)
function (p::CoevolveJumpAggregation{T, S, F1, F2})(
        integrator::AbstractSSAIntegrator,
) where {T, S, F1, F2 <: Union{Tuple, Nothing}}
    body
end
julia> format_text(s, SciMLStyle(); join_lines_based_on_source=true) |> print
┌ Info: groups
│   placeholder_groups =
│    2-element Vector{Vector{Int64}}:
│     [2, 5, 8, 11, 14, 17, 20]
└     [24, 27, 30, 33, 36]
┌ Info:
│   segments =
│    1-element Vector{Tuple{Int64, Int64}}:
└     (1, 6)
┌ Info:
│   segments =
│    1-element Vector{Tuple{Int64, Int64}}:
└     (1, 4)
const JUMP_AGGREGATORS = (Direct(), DirectFW(), DirectCR(), SortingDirect(), RSSA(), FRM(),
    FRMFW(), NRM(), RSSACR(), RDirect(), Coevolve())
```

ok it's getting better

src/styles/sciml/nest.jl
```diff
@@ -1,35 +1,23 @@
...
     :n_generator!,
     :n_filter!,
     :n_flatten!,
-    :n_using!,
-    :n_export!,
-    :n_import!,
-    :n_chainopcall!,
-    :n_comparison!,
-    :n_for!,
-    #:n_vect!
```
efaulhaber, line 24: This was commented out.

src/styles/sciml/nest.jl
```diff
@@ -64,32 +52,160 @@ function n_functiondef!(ss::SciMLStyle, fst::FST, s::State)
...
+    return
+end
+
+for f in [
+    :n_tuple!,
+    :n_call!,
+    :n_curly!,
+    :n_macrocall!,
+    :n_ref!,
+    :n_vect!,
```
efaulhaber, line 186: And now it's added here.

src/styles/sciml/pretty.jl
```diff
@@ -64,32 +65,59 @@ for f in [
...
-    :p_ref,
     :p_row,
     :p_nrow,
     :p_hcat,
     :p_comprehension,
-    :p_typedcomprehension, #:p_whereopcall,
+    :p_typedcomprehension,
     :p_generator,
     :p_filter,
-    :p_flatten, #:p_vect
```
efaulhaber, line 82: Same here with the pretty functions. Commented out here.

src/styles/sciml/pretty.jl
```diff
@@ -64,32 +65,59 @@ for f in [
...
+        $f(YASStyle(style), cst, s; kwargs...)
+    end
+end
+
+for f in [
+    :p_call,
+    :p_curly,
+    :p_ref,
+    :p_braces,
+    :p_vect,
```
efaulhaber, line 92: But now not commented out anymore.

efaulhaber (commented):
I did some digging, and found that these two changes are responsible for #806.
Is this intended?

PR #801, opened 2024-01-13
Fix indented docstrings with Unicode characters

devmotion:
Fixes the Unicode problem described in https://github.com/domluna/JuliaFormatter.jl/issues/65#issuecomment-1889518359 (the result is still a bit off/unintended but the same problem seems to happen on the master branch without Unicode characters).


src/styles/default/pretty.jl
```diff
@@ -283,27 +283,26 @@ end
 
 function format_docstring(style::AbstractStyle, state::State, text::AbstractString)
     state_indent = state.indent
-    boundaries = findall(text) do character
-        character != '"'
-    end
+    start_boundary = findfirst(!=('"'), text)
+    end_boundary = findlast(!=('"'), text)
```
devmotion, line 287: Just a minor optimization while working on this part of the code (I'll revert if desired): Only the indices of the first and last non-`"` character are needed.

src/styles/default/pretty.jl
```diff
@@ -283,27 +283,26 @@ end
...
     # first, we need to remove any user indent
     # only some lines will "count" towards increasing the user indent
     # start at a very big guess
     user_indent = typemax(Int)
-    user_indented = text[boundaries[1]:boundaries[end]]
+    user_indented = text[start_boundary:end_boundary]
```
devmotion, line 292: `findfirst` and `findlast` return valid indices of `text` (given that there is a non-`"` character, which would be good to check...)

src/styles/default/pretty.jl
```diff
@@ -283,27 +283,26 @@ end
...
     deindented = IOBuffer()
     user_lines = split(user_indented, '\n')
     for (index, line) in enumerate(user_lines)
         # the first line doesn't count
         if index != 1
-            first_character = findfirst(character -> !isspace(character), line)
```
devmotion, line 299: This returns a valid index of the string `line` (or `nothing`), but generally `first_character - 1` might neither be a valid index and nor be equal to the number of spaces before the first non-space character (in case there is a Unicode space character with > 1 codepoint?).

src/styles/default/pretty.jl
```diff
@@ -320,7 +319,7 @@ function format_docstring(style::AbstractStyle, state::State, text::AbstractStri
                     write(deindented, line)
                 else
                     write(deindented, '\n')
-                    write(deindented, SubString(line, user_indent + 1, length(line)))
```
devmotion, line 323: Similarly, here generally neither `user_indent + 1` nor `length(line)` (which is the number of characters, not necessarily the last index!) might be a valid index.

src/styles/default/pretty.jl
```diff
@@ -320,7 +319,7 @@ function format_docstring(style::AbstractStyle, state::State, text::AbstractStri
...
+                    write(deindented, chop(line; head = user_indent, tail = 0))
```
devmotion, line 322: It seemed a bit easier to use `chop` here (which returns a `SubString`) but if preferred this could also be written using `SubString`:
```suggestion
                    write(deindented, SubString(line, nextind(line, firstindex(line), user_indent), lastindex(line))
```

domluna:
this looks good to me. it's too bad commonmark doesn't do this by default though (or by an option afaik)

devmotion:
With the latest changes of the master branch, tests seem to pass 🙂 The diff is viewed best with "hide whitespace" option enabled: https://github.com/domluna/JuliaFormatter.jl/pull/801/files?w=1

domluna:
@devmotion are we good to merge this then?

devmotion:
Yes, good from my side.

PR #807, opened 2024-02-12
Fix vector nesting with `SciMLStyle` and `yas_style_nesting = true`

### LINKED ISSUE #806: #792 breaks vector definitions with `SciMLStyle` and `yas_style_nesting`
Before #792, this was possible with `SciMLStyle` and `yas_style_nesting`:
```julia
some_very_loooooooooooooooooong_variable_name = [
    "some_pretty_long_string",
    "another_pretty_long_string",
    "yet_another_pretty_loooooooooooooooooong_string"
]
```
Now, it's reformatted to
```julia
some_very_loooooooooooooooooong_variable_name = ["some_pretty_long_string",
                                                 "another_pretty_long_string",
                                                 "yet_another_pretty_loooooooooooooooooong_string"]
```
@domluna Is this intended or a bug?

The former style is very common, which is why I added the `variable_call_indent` option to allow the same for `Dict`s.

efaulhaber:
Before #792, default style `p_vect` and `n_vect!` have always been used in `SciMLStyle`, even with `yas_style_nesting = true`. I restored this behavior.

Closes #806.

### DIFF
--- a/src/styles/sciml/nest.jl
+++ b/src/styles/sciml/nest.jl
@@ -183,7 +183,7 @@ for f in [
     :n_curly!,
     :n_macrocall!,
     :n_ref!,
-    :n_vect!,
+    # :n_vect!, don't use YAS style vector formatting with `yas_style_nesting = true`
     :n_braces!,
     :n_parameters!,
     :n_invisbrackets!,
--- a/src/styles/sciml/nest.jl
+++ b/src/styles/sciml/nest.jl
@@ -81,10 +81,6 @@ function _n_tuple!(ss::SciMLStyle, fst::FST, s::State)
     has_closer = is_closer(fst[end])
     start_line_offset = s.line_offset
 
-    if s.opts.yas_style_nesting && length(nodes) > 0 && is_opener(fst[1])
-        fst.indent += 1
-    end
-
     if has_closer
         fst[end].indent = fst.indent
     end
@@ -182,7 +178,7 @@ function _n_tuple!(ss::SciMLStyle, fst::FST, s::State)
 end
 
 for f in [
-    # :n_tuple!,
+    :n_tuple!,
     :n_call!,
     :n_curly!,
     :n_macrocall!,
@@ -201,11 +197,6 @@ for f in [
             _n_tuple!(style, fst, s)
         end
     end
-end
-
-function n_tuple!(ss::SciMLStyle, fst::FST, s::State)
-    style = getstyle(ss)
-    _n_tuple!(style, fst, s)
 end
 
 for f in [:n_chainopcall!, :n_comparison!, :n_for!]
--- a/src/styles/sciml/nest.jl
+++ b/src/styles/sciml/nest.jl
@@ -183,7 +183,7 @@ for f in [
     :n_curly!,
     :n_macrocall!,
     :n_ref!,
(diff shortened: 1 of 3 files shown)


src/styles/sciml/nest.jl
```diff
@@ -183,7 +183,7 @@ for f in [
     :n_curly!,
     :n_macrocall!,
     :n_ref!,
-    :n_vect!,
+    # :n_vect!, don't use YAS style vector formatting with `yas_style_nesting = true`
```
efaulhaber, line 186: With `yas_style_nesting = true`, this just restores the original version before #792.

With `yas_style_nesting = false`, #792 forwarded `n_vect!` to `_n_tuple!` below. Now it is forwarded to the default style `n_vect!`. Is this correct?

domluna: This would change the nesting behaviour imposed in #792 so that it would put all arguments on one line instead of trying to best fit the arguments among the fewest lines while trying to fit within the maximum margin.

The entire point of #792 was that the former behaviour was unwanted amongst people in the SciML organization.

domluna: for vect specifically if this is fine then I'm ok with the PR but I'm not sure if there's a concensus on it.

efaulhaber: I see. I guess we want to best fit, but still allow the line break with vector definitions. I'll have another look tomorrow. Or can you point me in the right direction how to do that?

domluna: > but still allow the line break with vector definitions

can you give an example of what this means?

efaulhaber: Sorry, I meant that I would like to combine the vector definition style that I restored in this PR with the best fit feature you implemented.

efaulhaber:
#808

domluna:
It's called here https://github.com/domluna/JuliaFormatter.jl/blob/master/src/styles/sciml/nest.jl#L100-L121 which is dispatched as SciMLStyle. YAS style hasn't changed.

efaulhaber:
Then I suggest merging this PR for now, and I'll see if I can find time to make the best fit feature work with YAS in a new PR.

efaulhaber:
Update: This PR is messing up `SciMLStyle`.
master:
```julia
julia> print(format_text(str3, SciMLStyle(), margin = 32))
test = [arg1, arg2, arg3,
    arg4, arg5]
```
This PR:
```julia
julia> print(format_text(str3, SciMLStyle(), margin = 32))
test = [
    arg1,
    arg2,
    arg3,
    arg4,
    arg5
]
```

test/sciml_style.jl
```diff
@@ -430,61 +430,153 @@
...
 
-        fstr = """
-        function foo(
+            fstr = """
+            test = [arg1,
+                arg2,
+                arg3,
+                arg4,
+                arg5]
+            """
```
efaulhaber, line 558: These look a bit weird. But this is how it's formatted on master. I only added these tests here to check that best optimal nesting still works with vector definitions because my previous attempt broke this.

domluna:
looks good to me the nightly test failure is because of the new Memory type. I'm not sure what that's about.

efaulhaber:
Do we have to wait for the nightly tests, or can we merge and release before that?

domluna:
we can merge it

PR #831, opened 2024-04-30
Put quotes around nothing

araujoms:
The documentation says that we need `always_for_in = nothing` to disable it, but this is not true, what we need is `always_for_in = "nothing"`. I've changed the documentation to match the code, and also the style file `pretty.jl` which was broken.


domluna:
in what situation does this come up? Config from a .JuliaFormatter.toml file?

araujoms:
Indeed, put that into `.JuliaFormatter.toml`, call `format`, and see it fail.

domluna:
ahh ic. In that case I'm rather see if we can get this worse as is with some additional parsing handling of the TOML file

araujoms:
My syntax highlighting really doesn't like "nothing" without the quotation marks in a TOML file, so maybe this violates the format specification?

pfitzseb:
Yeah, that's invalid TOML. Should really be a string imho.

domluna:
https://github.com/domluna/JuliaFormatter.jl/blob/master/src/JuliaFormatter.jl#L487-L493

ok so this would already work if the value is "nothing" so we don't need to update anything besides documentation can we change the PR to do that instead.

araujoms:
That's what the PR is doing.

domluna:
> That's what the PR is doing.

it's changing it to "nothing" but during parsing when "nothing" is written in the TOML file it will be converted to `nothing`, so we don't need to change the code implementation. What I was referring to would be a note that says the TOML file should be

always_for_in = "nothing"

and not

always_for_in = nothing

araujoms:
I know, I'm not changing the code implementation, I'm only changing the documentation (and the style file `pretty.jl`).

araujoms:
Nevermind then.

src/styles/minimal/pretty.jl
```diff
@@ -11,10 +11,10 @@ function options(style::MinimalStyle)
         indent = 4,
         annotate_untyped_fields_with_any = false,
         join_lines_based_on_source = true,
-        trailing_comma = nothing,
+        trailing_comma = "nothing",
```
domluna, line 14: these not stay as is, the issue is documenting the TOML file. In that file `nothing` needs to be surrounded with quotes but everything else remains as is.

araujoms: I'm no longer interested. I'll also unsubscribe from this PR.

PR #840, opened 2024-06-06
Change pre-commit hook to run on specific files only

abelsiqueira:
The pre-commit hook should run only on jl, md, jmq, and qmd files, since it only modifies those files.

I'm running the pre-commit a lot on [COPIERTemplate.jl](https://github.com/abelsiqueira/COPIERTemplate.jl), and I've noticed that it runs for any modifications, not just relevant ones. This PR changes that.

@jmuchovej, you created the file, so let me know if you agree with the changes.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -2,6 +2,8 @@
   name: "Julia Formatter"
   entry: "julia -e 'import JuliaFormatter: format; format(\".\")'"
   pass_filenames: false
-  always_run: true
+  always_run: false
+  types: [file]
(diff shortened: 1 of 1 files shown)


abelsiqueira:
I guess all the tests were running and some cancelled them manually? It could be useful to have something like https://github.com/abelsiqueira/COPIERTemplate.jl/blob/7e3060531bfac1dc05892fe45fc3eaf4aebb2aac/.github/workflows/TestOnPRs.yml#L7-L10 to only run the tests if the files changed match the path.

domluna:
i dont think the failures have anything to do with the PR

PR #848, opened 2024-06-20
Initial stage of JuliaSyntax

domluna:
Initial stage for integrating JuliaSyntax. In this part I want to generate the Document using JuliaSyntax instead of the Tokenize package.

Current improvements:

- If the format fails it will pinpoint the exact location. Right now this is hit or miss.
- Format skips work better.
- Remove DataStructures package

I might actually just remove CSTParser completely since that might be easier but we'll see


src/print.jl
```diff
@@ -103,10 +131,10 @@ function print_tree(
 
         if n.typ === NEWLINE && s.on && i < length(nodes)
             if is_closer(nodes[i+1]) || nodes[i+1].typ === Block || nodes[i+1].typ === Begin
-                write(io, repeat(" ", max(nodes[i+1].indent, 0)))
+                s.on && write(io, repeat(" ", max(nodes[i+1].indent, 0)))
                 s.line_offset = nodes[i+1].indent
             elseif !skip_indent(nodes[i+1])
-                write(io, ws)
+                s.on && write(io, ws)
```
domluna, line 137: this was a bug but I'm not sure it ever was an issue.

src/print.jl
```diff
@@ -4,32 +4,60 @@ function format_check(io::IOBuffer, fst::FST, s::State)
...
+        end
+
+        output = JuliaSyntax.sourcetext(s.doc.srcfile)[skip.startoffset:skip.endoffset]
+        l1 = skip.endline + 1
+        l2 = fst.endline
+        nlines = numlines(s.doc)
+
+        if l1 <= nlines && output[end] == '\n'
+            output = output[1:prevind(output, end)]
+        end
```
domluna, line 26: NotCode nodes will add newlines so we need to remove these for now

src/other/precompile.jl
```diff
@@ -1,15 +1,15 @@
...
+#        @noinline require_complete(m::Matching) =
+#            m.inv_match === nothing && throw(ArgumentError("Backwards matching not defined. `complete` the matching first."))
+#     """
+#     @compile_workload begin
+#         format(dir)
+#         for style = [DefaultStyle(), BlueStyle(), SciMLStyle(), YASStyle(), MinimalStyle()]
+#           format_text(str, style)
+#         end
+#     end
+# end
```
domluna, line 15: disabled until this works

src/document.jl
```diff
@@ -150,84 +79,92 @@ function Document(text::AbstractString)
...
-    for (l, r) in enumerate(ranges)
-        insert!(range_to_line, r, l)
-        line_to_range[l] = r
     end
 
     # If there is a SINGLE "#! format: off" tag
     # do not format from the "off" tag onwards.
     if length(stack) == 1 && length(format_skips) == 0
         # -1 signifies everything afterwards "#! format: off"
         # will not formatted.
```
domluna, line 154: this comment is no longer applicable

PR #849, opened 2024-06-21
BioStructures.jl compat

jgreener64:
I don't know if this file is still in use, but the recent release of BioStructures.jl v4.0.0 means that this line needs changing.


domluna:
thanks but this file is only used for testing formatting. The code in it is not actually run, so it doesn't matter if the code is outdated.

PR #851, opened 2024-06-22
Format only files selected by pre-commit

m-fila:
The hook will be more useful if it would work on a list of files given to it by pre-commit. For instance `pre-commit run` will check only files in the staging area, `pre-commit run --all-files` will run on all the files. Additionally the hook should respect the `exclude` directive of pre-commit.

This PR adds support to working on the lists of files provided by pre-commit


domluna:
If you pass no files would be default to "." ?

m-fila:
Pre-commit will skip a hook if there are no eligible files

domluna:
ok this looks good to me

abelsiqueira:
Can we have a new release so this and #840 are available, please?

PR #861, opened 2024-08-06
Allow expanding all short function definitions

jw3126:
I would like to have a way to banish all short function definitions from a code base. This PR implements one way to make that possible, by setting the max length of short function definitions to zero. But I would be happy to implement a different mechanism if desired. 
I will add docs once it is clear whether we want an option like this.


domluna:
I generally like this since it's backwards compatible. The downside that comes to mind is that it opens up the door for doing this in all sorts of circumstances for other parts of the code.

jw3126:
So what should I do:
* Polish this PR so it can be merged?
* Implement a different approach to banish all short functions?
* Close this PR?

jw3126:
Any suggestion for a name for the new boolean option?

domluna:
i like `force_long_function_def`. sorry I didn't reply sooner totally forgot about this

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "JuliaFormatter"
 uuid = "98e50ef6-434e-11e9-1051-2b60c6c9e899"
 authors = ["Dominique Luna <dluna132@gmail.com>"]
-version = "1.0.59"
```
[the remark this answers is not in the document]
jw3126: I changed it. Although personally, I would interpret this PR as a feature that should increase the minor version instead of a bugfix that should increase the patch version.

jw3126:
@domluna can you approve CI?

domluna:
failure seems unrelated? maybe something to do with the GH action

src/options.jl
```diff
@@ -35,6 +36,22 @@ Base.@kwdef struct Options
...
+        opts = new(args...)
+        if (opts.force_long_function_def === true) &&
+           (opts.short_to_long_function_def === false)
+            msg = """
+            The combination `force_long_function_def = true` and `short_to_long_function_def = false` is invalid.
+            """
+            throw(ArgumentError(msg))
+        end
+        return opts
+    end
```
domluna, line 49: I think this function is causing failures on Julia < v1.6

domluna: ERROR: LoadError: LoadError: syntax: ... is not supported inside "new"

domluna: @jw3126 update this when you get the chance and it should be g2g

PR #863, opened 2024-08-22
use TOML instead of Pkg.TOML to read TOML files

KristofferC:


domluna:
fine with me just curious why TOML instead of Pkg.TOML though. is the latter outdated?

KristofferC:
Pkg is a big dependency. TOML is a small one. They have different load times for example. And one shouldn't in general rely on other packages dependencies.

PR #883, opened 2024-10-19
V2 Fixes

### LINKED ISSUE #878: New formatting turns optional positional argument into keyword argument
Kudos on the full rewrite!
Unfortunately I have encountered a bug when I tried to apply the new formatting in https://github.com/gdalle/DifferentiationInterface.jl/pull/589. An unnamed, `Val`-typed positional argument was "formatted" into a keyword argument (see [this comment](https://github.com/gdalle/DifferentiationInterface.jl/pull/589/files#r1803531574)).

Schematically, the following code
```julia
function f(x, y, ::Val{N}=Val(1)) where {N}
```
turned into (note the semicolon)
```julia
function f(x, y; (::Val{N})=Val(1)) where {N}
```
which changed the semantics.

EDIT: As discussed below, this apparently applied to all _optional_ positional arguments, named or not. It might be due to BlueStyle.

domluna:
fixes #878 

```
     open(path, "w") do io
-        TOML.print(io, data)
+        return TOML.print(io, data)
     end
```

The `return` is no longer added for `do` expressions.

### DIFF
--- a/src/fst.jl
+++ b/src/fst.jl
@@ -83,6 +83,7 @@
     Const,
     Import,
     Export,
+    Public,
     Using,
     File,
     Quotenode,
@@ -425,7 +426,7 @@ is_opener(t::JuliaSyntax.GreenNode) = kind(t) in KSet"{ ( ["
 function is_iterable(t::JuliaSyntax.GreenNode)
     if !(
         kind(t) in
-        KSet"parens tuple vect vcat braces curly comprehension typed_comprehension macrocall ref typed_vcat import using export"
+        KSet"parens tuple vect vcat braces curly comprehension typed_comprehension macrocall ref typed_vcat import using export public"
     )
         is_func_call(t)
     else
@@ -447,7 +448,7 @@ function is_named_iterable(x::FST)
 end
 
 function is_import_expr(x::FST)
-    return x.typ in (Import, Using, Export)
+    return x.typ in (Import, Using, Export, Public)
 end
 
 """
--- a/src/fst.jl
+++ b/src/fst.jl
@@ -630,7 +630,7 @@ function is_function_or_macro_def(cst::JuliaSyntax.GreenNode)
         return false
     end
     k = kind(cst)
-    if (k == K"function" || k == K"macro")
+    if k in KSet"function macro"
         return true
     end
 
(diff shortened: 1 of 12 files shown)


ararslan:
> The `return` is no longer added for `do` expressions.

Is it not possible to support that? It's a requirement for YASGuide.

domluna:
> > The `return` is no longer added for `do` expressions.
> 
> Is it not possible to support that? It's a requirement for YASGuide.

Is the example above, with the return added, not a dealbreaker for YAS? It's really tricky to only add return in cases where not nothing is returned. I think in v1 this was broken in a bunch of ways but functional enough that it worked for the most cases. In v2 this was fixed but the downside is that you get `return TOML.print(io, data)`.

ararslan:
> It's really tricky to only add return in cases where not nothing is returned.

I'm not sure about the other style guides but at least for YAS the `return` is always required, even if it's `nothing`. In that example, `return TOML.print(io, data)` is correct for YAS. I'd probably personally write it as separate `TOML.print(io, data)` + `return nothing` but that's functionally equivalent and I wouldn't expect a formatter to be able to know that `TOML.print` is used for its side effects (writing to `io`) and not for its return value. Sounds like it's fully correct for YAS in v2 then?

domluna:
> > It's really tricky to only add return in cases where not nothing is returned.
> 
> I'm not sure about the other style guides but at least for YAS the `return` is always required, even if it's `nothing`. In that example, `return TOML.print(io, data)` is correct for YAS. I'd probably personally write it as separate `TOML.print(io, data)` + `return nothing` but that's functionally equivalent and I wouldn't expect a formatter to be able to know that `TOML.print` is used for its side effects (writing to `io`) and not for its return value. Sounds like it's fully correct for YAS in v2 then?

I'll put this back in for YAS style.

> I wouldn't expect a formatter to be able to know that `TOML.print` is used for its side effects (writing to `io`) and not for its return value.

right, we would need to do side effect analysis which is out of scope for a formatter from my PoV.

domluna:
Merging this to master, still some things we can do but this had most of the fixes in.

RoyCCWang:
Thanks for everyone's work on this. May we please get a new release to get all these merged changes to the end user? I think just the #886  change along can encourage more Julia developers to consider writing non-exported public APIs, and It'll address #890 and #892 too.

PR #884, opened 2024-10-23
Disable following links

dpaetzel:
Thanks for developing and maintaining this project! :slightly_smiling_face: 

Following links resulted on my machine (running [NixOS](https://nixos.org/) and using [devenv](https://devenv.sh/)) in precompilation of JuliaFormatter failing because it attempted to format read-only `.jl` files (because `.devenv/profile` links to readonly `/nix/store/…` directories). Since I don't see a case where I want JuliaFormatter to jump to far-away directories by following links and reformat them (looks rather dangerous to me, actually?), I simply disabled following links. In case you want to integrate this tiny change, I created this PR.

Note that this could be improved by checking whether the link points to a file within the currently-to-be-formatted directory or something like that.


domluna:
makes sense thanks!

Issue #928, opened 2025-07-17
Precompilation fails with JuliaSyntax: unknown Kind name "core_@cmd"


Sixzero:
## Problem

JuliaFormatter fails to precompile with the following error when using JuliaSyntax:

```
ERROR: LoadError: unknown Kind name "core_@cmd"
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:44
  [2] (::JuliaSyntax.var"#Kind##0#Kind##1"{SubString{String}})()
    @ JuliaSyntax ~/.julia/packages/JuliaSyntax/J00sR/src/kinds.jl:47
  [3] get
    @ ./dict.jl:526 [inlined]
  [4] Kind
    @ ~/.julia/packages/JuliaSyntax/J00sR/src/kinds.jl:46 [inlined]
  [5] #@KSet_str##0
    @ ./none:-1 [inlined]
  [6] iterate
    @ ./generator.jl:48 [inlined]
  [7] collect_to!
    @ ./array.jl:848 [inlined]
  [8] collect_to_with_first!(dest::Vector{JuliaSyntax.Kind}, v1::JuliaSyntax.Kind, itr::Base.Generator{Vector{SubString{String}}, JuliaSyntax.var"#@KSet_str##0#@KSet_str##1"}, st::Int64)
    @ Base ./array.jl:826
  [9] collect(itr::Base.Generator{Vector{SubString{String}}, JuliaSyntax.var"#@KSet_str##0#@KSet_str##1"})
    @ Base ./array.jl:800
 [10] var"@KSet_str"(__source__::LineNumberNode, __module__::Module, str::Any)
    @ JuliaSyntax ~/.julia/packages/JuliaSyntax/J00sR/src/kinds.jl:170
 [11] include(mapexpr::Function, mod::Module, _path::String)
    @ Base ./Base.jl:304
 [12] top-level scope
    @ ~/.julia/packages/JuliaFormatter/IqoSF/src/JuliaFormatter.jl:116
 [13] include(mod::Module, _path::String)
    @ Base ./Base.jl:303
 [14] include_package_for_output(pkg::Base.PkgId, input::String, depot_path::Vector{String}, dl_load_path::Vector{String}, load_path::Vector{String}, concrete_deps::Vector{Pair{Base.PkgId, UInt128}}, source::Nothing)
    @ Base ./loading.jl:2991
```

## Analysis

[...]

PR #929, opened 2025-07-18
Try to compat JuliaSyntax v1.0

hz-xiaxz:
Disclaimer: I'm in no way familiar with JuliaSyntax, just play with Gemini and try to make it not error.
related to #901 #928 
No error on the kinds part now, I fixed the lost API of `core_@cmd` (macro is now parsed to a tree), `false` (this is `Bool` now), `cartesian_iteration` (maybe it's `iteraction` now? https://github.com/JuliaLang/JuliaSyntax.jl/issues/432)   

changed deprecated `has_children` to `!is_leaf`.

I would admit I don't fully understand the change, but I hope it would be a good starting point?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -13,7 +13,7 @@ TOML = "fa267f1f-6049-4f14-aa54-33bafae1ed76"
 [compat]
 CommonMark = "0.5, 0.6, 0.7, 0.8, 0.9"
 Glob = "1.3"
-JuliaSyntax = "^0.4.10"
+JuliaSyntax = "1"
 PrecompileTools = "1"
 TOML = "1"
 julia = "1.10"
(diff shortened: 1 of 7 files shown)


hz-xiaxz:
tons of test not passed, I can't fix it myself.
For a cleaner diff, maybe revert the [a218a99](https://github.com/domluna/JuliaFormatter.jl/pull/929/commits/a218a99a4225d8f7351b79a149319a437093fbb3) locally?

domluna:
the main problem with v1 is that it's changed how operators are treated. they are now all identifiers. the formatter relied heavily on them being tagged as operators and being able to determine the type of operator so a more involved solution is required that would take some time to figure out.

oscardssmith:
It is worth getting 1.0 support as 2.0 is an intermediary in terms of what changes. Also JuliaSyntax 1.x is likely to be used by Julia 1.10-1.12 for a while. IIUC JuliaSyntax 2.0 is for Julia 1.13+ with JuliaLowering.

Issue #789, opened 2023-12-06
special case for guard pattern match syntax if ... end => ...


Roger-luo:
This is related to #404, consider the following code

```julia
@match node begin
    if is_leaf(node) end => nothing
end
```

currently this will format it into

```julia
@match node begin
    if is_leaf(node)
    end => nothing
end
```

I'm wondering if we can special case this?

domluna:
we probably could. right now the we just split it no matter what but we could add an option where we consider the margin.

penelopeysm:
#937 provides a config option that might fix this

PR #937, opened 2025-08-08
Add support for `preserve_single_line_if::Bool`

### LINKED ISSUE #516: Allow preserving style for single-line `if` expressions
Single-line `if` expressions currently get split into multiple lines. Would it be possible to keep the style to preserve the first version below?

Before formatting:
```julia
function f()
  if rand() < 0.5 return 2 end
  0
end
```

After formatting (`if` gets split):
```julia
function f()
    if rand() < 0.5
        return 2
    end
    0
end
```

While the splitting is probably desired in most cases, the case above is equivalent to a short-circuiting `&&` so it's a bit of a corner case, but still it could be made more readable by keeping it in one line.

alexandergunnarson:
Before formatting (or with `preserve_single_line_if = true`):
```jl
function f()
    if rand() < 0.5 return 2 end
    0
end
```

After formatting with `preserve_single_line_if = false` (default):
```jl
function f()
    if rand() < 0.5
        return 2
    end
    0
end
```

Resolves https://github.com/domluna/JuliaFormatter.jl/issues/516.

### DIFF
--- a/src/options.jl
+++ b/src/options.jl
@@ -36,7 +36,7 @@ Base.@kwdef struct Options
     yas_style_nesting::Bool = false
     short_circuit_to_if::Bool = false
     disallow_single_arg_nesting::Bool = false
-    preserve_single_line_if::Bool = true
+    preserve_single_line_if::Bool = false
 
     function Options(args...)
         opts = new(args...)
--- a/src/options.jl
+++ b/src/options.jl
@@ -36,6 +36,7 @@ Base.@kwdef struct Options
     yas_style_nesting::Bool = false
     short_circuit_to_if::Bool = false
     disallow_single_arg_nesting::Bool = false
+    preserve_single_line_if::Bool = true
 
     function Options(args...)
         opts = new(args...)
--- a/src/styles/default/pretty.jl
+++ b/src/styles/default/pretty.jl
@@ -1622,9 +1622,14 @@ function p_if(
             add_node!(t, pretty(style, c, s, ctx, lineage), s)
         elseif kind(c) === K"block"
             s.indent += s.opts.indent
+            block_ctx = if s.opts.preserve_single_line_if
+                ctx
+            else
+                newctx(ctx; ignore_single_line = true)
+            end
             add_node!(
                 t,
-                pretty(style, c, s, newctx(ctx; ignore_single_line = true), lineage),
+                pretty(style, c, s, block_ctx, lineage),
                 s;
                 max_padding = s.opts.indent,
             )
--- a/src/styles/default/pretty.jl
+++ b/src/styles/default/pretty.jl
@@ -1852,32 +1852,68 @@ function p_if(
         return t
     end
 
+    preserve_single_line =
+        s.opts.preserve_single_line_if &&
+        on_same_line(s, s.offset, s.offset + span(cst) - 1)
+
+    is_first = true
+
     for c in children(cst)
         if kind(c) in KSet"if elseif else"
             if !haschildren(c)
-                add_node!(t, pretty(style, c, s, ctx, lineage), s; max_padding = 0)
+                if preserve_single_line && !is_first
+                    add_node!(t, Whitespace(1), s)
+                end
+                add_node!(
+                    t,
+                    pretty(style, c, s, ctx, lineage),
+                    s;
+                    max_padding = if preserve_single_line
+                        -1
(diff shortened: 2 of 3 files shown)


src/options.jl
```diff
@@ -36,6 +36,7 @@ Base.@kwdef struct Options
     yas_style_nesting::Bool = false
     short_circuit_to_if::Bool = false
     disallow_single_arg_nesting::Bool = false
+    preserve_single_line_if::Bool = true
```
domluna, line 39: should default to `false` to keep current behaviour by default

domluna (changes requested):
please add some test cases

looks good otherwise.

src/options.jl
alexandergunnarson: Good point! Done

penelopeysm:
This patch no longer works on the current version:

```julia
julia> using JuliaFormatter: format_text

julia> s = "if rand() < 0.5 return 2 end"
"if rand() < 0.5 return 2 end"

julia> format_text(s; preserve_single_line_if=true) |> println
if rand() < 0.5
    return 2
end
```

@alexandergunnarson Are you still interested in working on this? No worries if not as I can pick it up (at some point in time....).

alexandergunnarson:
@domluna Sorry — added test cases in the most recent commit.

@penelopeysm — Thanks for alerting me! Patch now works with latest version.

penelopeysm:
Thanks 😊 This looks generally great. I have one question about what the intended behaviour should be when overflowing the max width. Currently we have this:

```julia
julia> using JuliaFormatter: format_text

julia> format_text("if x > 0 return 1 end"; preserve_single_line_if=true, margin=7) |> println
if x >
   0 return 1 end

julia> format_text("if x > 0 return 1 end"; preserve_single_line_if=true, margin=8) |> println
if x > 0 return 1 end
```

The issue is that the `x > 0` gets nested if it goes over the margin. While the result parses fine, I think it's a bit unprincipled for the if-condition to determine whether there's a line break. I think there are two possible options for `preserve_single_line_if = true`:

1. If the _entire_ `if ... end` exceeds the margin, then we split it up.

2. Never split it up.

~~I would lean towards (2)~~ Actually, I'm kind of undecided :/ -- what do you think?

src/styles/default/pretty.jl
```diff
@@ -1852,27 +1852,68 @@ function p_if(
...
+                    t,
+                    pretty(style, c, s, newctx(ctx; ignore_single_line = true), lineage),
+                    s;
+                    max_padding = s.opts.indent,
+                )
+                s.indent -= s.opts.indent
+            end
         elseif !JuliaSyntax.is_whitespace(c)
             add_node!(t, Whitespace(1), s)
             add_node!(t, pretty(style, c, s, ctx, lineage), s; join_lines = true)
```
penelopeysm, line 1919: (If you agree with (2), then I believe this is the appropriate fix:)

```suggestion
            ctx2 = preserve_single_line_if ? newctx(ctx; nonest=true) : ctx
            add_node!(t, pretty(style, c, s, ctx2, lineage), s; join_lines = true)
```

penelopeysm: (1) looks trickier to accomplish to me. I haven't attempted to figure out how that would be done.

penelopeysm:
Found an unusual extra space popping up here as well (example from #789)

```julia
julia> s = """
       @match node begin
           if is_leaf(node) end => nothing
       end
       """
"@match node begin\n    if is_leaf(node) end => nothing\nend\n"

julia> ft(s) |> println
@match node begin
    if is_leaf(node)
    end => nothing
end


julia> ft(s; preserve_single_line_if=true) |> println
@match node begin
    if is_leaf(node)  end => nothing
end
```

penelopeysm:
#1254 did this (it's a bit more general in that it handles all blocks, not just `if`).

This requires JuliaFormatter >= v2.14.0, which should be released in a few minutes

```julia
julia> s = """
       function f()
           if rand() < 0.5 return 2 end
           0
       end"""
"function f()\n    if rand() < 0.5 return 2 end\n    0\nend"

julia> format_text(s) |> println
function f()
    if rand() < 0.5
        return 2
    end
    0
end

julia> format_text(s; preserve_single_line_blocks=true) |> println
function f()
    if rand() < 0.5 return 2 end
    0
end
```

PR #954, opened 2025-11-21
Fix formatting hex float literals

### LINKED ISSUE #921: Formatting float point change its value
- Julia Version 1.11.5 (2025-04-14)
- JuliaFormatter v2.1.2
- Using `blue` style

```diff
-Float32(-0x1p-26)
+Float32(-0x1p-26.0)
```

```jl
julia> Float32(-0x1p-26)
-1.4901161f-8

julia> Float32(-0x1p-26.0)
-0.0f0

julia> Float32(-0x1p-26.0) == Float32(-0x1p-26)
false
```

mhauru:
Closes #921

There's still this case, where the formatter changes the value of the expression:
```julia
julia> format_text("0x1p1.2")
"0x1p10.2"

julia> 0x1p1.2
0.4

julia> 0x1p10.2
204.8
```
However, I think `0x1p1.2` should be a syntax error, so maybe not worth fixing? I've reported it as a Julia bug here https://github.com/JuliaLang/julia/issues/60189. To be clear, what happens there is that Julia parses `0x1p1.2` as a juxtaposition of `0x1p1` and `.2`, i.e. a multiplication. The formatter then expands the .2 to be 0.2, which results in `0x1p10.2`, changing the value of the first literal that is being multiplied.

### DIFF
--- a/test/issues.jl
+++ b/test/issues.jl
@@ -1955,6 +1955,8 @@
     end
 
     @testset "921" begin
+        # These are floating point literals with hexadecimal significands. E.g. 0x1p1 is the
+        # float with with 0x1 as the significant and 1 as the exponent.
         s = "0x1p1"
         @test fmt(s) == s
         s = "+0x1p1"


domluna (approved):

mhauru:
Not from my side, happy to merge.

PR #959, opened 2025-12-16
Switch pre-commit hook to CLI

### LINKED ISSUE #958: Modify pre-commit hook to use CLI?
This would allow it to work even for people who don't have JuliaFormatter in their default env

gdalle:
>[!WARNING]
> This is a breaking change as far as the pre-commit hook is concerned, but I don't think we have any way to version it?

- Switches to the `jlfmt` CLI inside the commit hook (fixes #958)
- Modifies the local hook used for this repo so that it actually works (the YAML apparently [needs more keys](https://pre-commit.com/#repository-local-hooks)) and excludes the `test` folder
- Format an unformatted source file (`app.jl`)

Do we have any way to test the hook, apart from running it on this very repo?

### DIFF
--- a/docs/src/integrations.md
+++ b/docs/src/integrations.md
@@ -18,6 +18,12 @@ repos:
 # ... other repos you may have
 ```
 
+Note that it requires the `jlfmt` Pkg app to be installed with
+
+```julia
+pkg> app add JuliaFormatter
+```
+
 You can find a list of releases [here](https://github.com/domluna/JuliaFormatter.jl/releases).
 **Be sure to use the entire version string!** (You can double-check this by opening the
 release and looking at the part of the URL that follows `.../releases/tag/VERSION`.)


gdalle:
I added a [`pre-commit` run to CI](https://pre-commit.com/#usage-in-continuous-integration), similar to what I have set up for DifferentiationInterface. It will ensure that whatever is in `.pre-commit-config.yaml` actually runs, so if `.pre-commit-hooks.yml` is kept in sync with that, we should be fine. Note that there is a bundled GitHub Action, but it is [not maintained](https://github.com/pre-commit/action).

gdalle:
Could you re-approve CI to check that the new workflow runs properly?

gdalle:
I think I figured it out, the current action [doesn't seem to be aware](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#adding-a-system-path) of path modifications. The last commit should fix it by using the explicit path to `jlfmt` inside CI. Could you approve the workflow again?

PR #978, opened 2026-05-05
Add compat for JuliaSyntax v1

### LINKED ISSUE #948: Update JuliaSyntax.jl to v1
Packages depending on JuliaFormatter.jl (for example, ModelingToolkit.jl) are currently locked to JuliaSyntax.jl v0.4.

### LINKED ISSUE #950: v2 chokes on `@main`
JuliaFormatter@2 fails to format code that uses the `@main` macro.

```jl
$ cat hello.jl 
function @main(ARGS::Vector{String})::Cint
        println("Hello, World")
    return 0
end
```

```jl
$ cat fmt.jl 
using Pkg;
Pkg.add(PackageSpec(name = "JuliaFormatter", version = "2"))

using JuliaFormatter;
format_file(ARGS[1]);

$ julia --threads=auto -O3 --project=@temp  fmt.jl hello.jl 
   Resolving package versions...
    Updating `/tmp/jl_h89kFn/Project.toml`
  [98e50ef6] + JuliaFormatter v2.2.0
    Updating `/tmp/jl_h89kFn/Manifest.toml`
  [a80b9123] + CommonMark v0.9.1
  [c27321d9] + Glob v1.3.1
  [98e50ef6] + JuliaFormatter v2.2.0
⌅ [70703baa] + JuliaSyntax v0.4.10
  [aea7be01] + PrecompileTools v1.3.3
  [21216c6a] + Preferences v1.5.0
  [ade2ca70] + Dates v1.11.0
  [de0858da] + Printf v1.11.0
  [fa267f1f] + TOML v1.0.3
  [4ec0a83e] + Unicode v1.11.0
        Info Packages marked with ⌅ have new versions available but compatibility constraints restrict them from upgrading. To see why use `status --outdated -m`
┌ Warning: Failed to format file /home/kousu/src/hello.jl due to a parsing error, skipping file
│   error =
│    ParseError:
│    # Error @ line 1:10
│    function @main(ARGS::Vector{String})::Cint
│    #        └─────────────────────────┘ ── Invalid signature in function definition
└ @ JuliaFormatter ~/.julia/packages/JuliaFormatter/zLsPw/src/JuliaFormatter.jl:448
```

v1 works:

```
$ cat fmt.jl 
using Pkg;
Pkg.add(PackageSpec(name = "JuliaFormatter", version = "1"))

using JuliaFormatter;
format_file(ARGS[1]);

$ julia --threads=auto -O3 --project=@temp  fmt.jl hello.jl ; cat hello.jl
   Resolving package versions...
    Updating `/tmp/jl_aPdJJX/Project.toml`
⌃ [98e50ef6] + JuliaFormatter v1.0.62
    Updating `/tmp/jl_aPdJJX/Manifest.toml`
  [00ebfdb7] + CSTParser v3.4.3
⌅ [a80b9123] + CommonMark v0.8.16
  [34da2185] + Compat v4.18.1
  [a8cc5b0e] + Crayons v4.1.1
⌅ [864edb3b] + DataStructures v0.18.22
  [c27321d9] + Glob v1.3.1
⌃ [98e50ef6] + JuliaFormatter v1.0.62
  [bac558e1] + OrderedCollections v1.8.1
  [aea7be01] + PrecompileTools v1.3.3
  [21216c6a] + Preferences v1.5.0
  [0796e94c] + Tokenize v0.5.29
  [2a0f44e3] + Base64 v1.11.0
  [ade2ca70] + Dates v1.11.0
  [b77e0a4c] + InteractiveUtils v1.11.0
  [ac6e5ff7] + JuliaSyntaxHighlighting v1.12.0
  [d6f4376e] + Markdown v1.11.0
  [de0858da] + Printf v1.11.0
  [9a3f8284] + Random v1.11.0
  [ea8e919c] + SHA v0.7.0
  [f489334b] + StyledStrings v1.11.0
  [fa267f1f] + TOML v1.0.3
  [cf7118a7] + UUIDs v1.11.0
  [4ec0a83e] + Unicode v1.11.0
        Info Packages marked with ⌃ and ⌅ have new versions available. Those with ⌃ may be upgradable, but those with ⌅ are restricted by compatibility constraints from upgrading. To see why use `status --outdated -m`
function @main(ARGS::Vector{String})::Cint
    println("Hello, World")
    return 0
end
```

```
$ julia --version
julia version 1.12.1
```

I assume if it gets lost on that it will get lost if I build other apps using macros. Will it break on `@ccallable` too?

Still a newbie here so I don't know where I'd even start to go debugging this but hopefully there's something obvious to you syntax experts in here.

Thank you. 🦢

svchb:
I kept the changes as minimal as possible.

### DIFF
--- a/src/fst.jl
+++ b/src/fst.jl
@@ -661,12 +661,8 @@ function is_function_or_macro_def(cst::JuliaSyntax.GreenNode)
     return false
 end
 
-function is_short_function_def(cst::JuliaSyntax.GreenNode)
-    kind(cst) === K"function" && haschildren(cst) || return false
-    idx = findfirst(n -> !JuliaSyntax.is_whitespace(n), children(cst))
-    isnothing(idx) && return false
-    return kind(cst[idx]) !== K"function"
-end
+is_short_function_def(cst::JuliaSyntax.GreenNode) =
+    JuliaSyntax.has_flags(cst, JuliaSyntax.SHORT_FORM_FUNCTION_FLAG)
 
 function is_function_like_lhs(node::JuliaSyntax.GreenNode)
     k = kind(node)
--- a/src/fst.jl
+++ b/src/fst.jl
@@ -631,9 +631,10 @@ function is_function_or_macro_def(cst::JuliaSyntax.GreenNode)
     return false
 end
 
-is_short_function_def(cst::JuliaSyntax.GreenNode) =
+function is_short_function_def(cst::JuliaSyntax.GreenNode)
     kind(cst) === K"function" &&
-    JuliaSyntax.has_flags(cst, JuliaSyntax.SHORT_FORM_FUNCTION_FLAG)
+        JuliaSyntax.has_flags(cst, JuliaSyntax.SHORT_FORM_FUNCTION_FLAG)
+end
 
 function is_function_like_lhs(node::JuliaSyntax.GreenNode)
     k = kind(node)
--- a/src/fst.jl
+++ b/src/fst.jl
@@ -629,10 +629,7 @@ function is_assignment(x::FST)
 end
 
 function is_assignment(t::JuliaSyntax.GreenNode)
-    if JuliaSyntax.is_prec_assignment(kind(t)) && haschildren(t)
-        return !any(c -> kind(c) in KSet"in ∈", children(t))
-    end
-    return false
+    return JuliaSyntax.is_syntactic_assignment(t) && haschildren(t)
 end
 is_assignment(::Nothing) = false
 
@@ -662,6 +659,7 @@ function is_function_or_macro_def(cst::JuliaSyntax.GreenNode)
 end
 
 is_short_function_def(cst::JuliaSyntax.GreenNode) =
+    kind(cst) === K"function" &&
     JuliaSyntax.has_flags(cst, JuliaSyntax.SHORT_FORM_FUNCTION_FLAG)
 
 function is_function_like_lhs(node::JuliaSyntax.GreenNode)
--- a/src/fst.jl
+++ b/src/fst.jl
@@ -278,7 +278,7 @@ function is_macrostr(t::JuliaSyntax.GreenNode)::Bool
 end
 
 function contains_macrostr(t::JuliaSyntax.GreenNode)::Bool
-    if kind(t) in KSet"StringMacroName CmdMacroName core_@cmd"
+    if kind(t) in KSet"StringMacroName CmdMacroName"
         return true
     elseif kind(t) === "quote" && haschildren(t)
         return contains_macrostr(t[1])
@@ -547,6 +547,21 @@ function _callinfo(x::JuliaSyntax.GreenNode)
         return 0, 0
     end
     k = kind(x)
+    if k === K"call" && JuliaSyntax.is_infix_op_call(x)
+        args = count(n -> !JuliaSyntax.is_whitespace(n), children(x))
+        return div(args - 1, 2), div(args + 1, 2)
+    elseif k === K"dotcall" && JuliaSyntax.is_infix_op_call(x)
+        args = count(n -> !JuliaSyntax.is_whitespace(n), children(x))
+        nops = div(args - 1, 3)
+        return nops, nops + 1
+    elseif k === K"op="
+        return 1, 2
+    elseif JuliaSyntax.is_operator(x) && haschildren(x)
+        args = count(n -> !JuliaSyntax.is_whitespace(n), children(x))
+        if args >= 3
+            return div(args - 1, 2), div(args + 1, 2)
+        end
+    end
     n_operators = 0
     n_args = 0
 
@@ -565,7 +580,7 @@ function _callinfo(x::JuliaSyntax.GreenNode)
 end
 
 function is_unary(x::JuliaSyntax.GreenNode)
-    if JuliaSyntax.is_unary_op(x)
+    if JuliaSyntax.is_unary_op(x) && !haschildren(x)
         return true
     end
     if kind(x) in KSet"call dotcall" || (JuliaSyntax.is_operator(x) && haschildren(x))
@@ -644,6 +659,13 @@ function is_function_or_macro_def(cst::JuliaSyntax.GreenNode)
     end
 
     return false
+end
+
+function is_short_function_def(cst::JuliaSyntax.GreenNode)
+    kind(cst) === K"function" && haschildren(cst) || return false
+    idx = findfirst(n -> !JuliaSyntax.is_whitespace(n), children(cst))
+    isnothing(idx) && return false
+    return kind(cst[idx]) !== K"function"
 end
 
 function is_function_like_lhs(node::JuliaSyntax.GreenNode)
@@ -812,6 +834,23 @@ function eq_to_in_normalization!(fst::FST, always_for_in::Bool, for_in_replaceme
         elseif op.val == "in" && op_kind(fst[end]) === K":"
             op.val = "="
             op.len = length(op.val)
+        end
+        if !isnothing(fst.metadata)
+            metadata = fst.metadata::Metadata
+            opkind = try
+                JuliaSyntax.Kind(op.val)
+            catch
+                metadata.op_kind
+            end
+            fst.metadata = Metadata(
+                opkind,
+                metadata.op_dotted,
+                metadata.is_standalone_shortcircuit,
+                metadata.is_short_form_function,
+                op.val == "=",
+                metadata.is_long_form_function,
+                metadata.has_multiline_argument,
+            )
         end
     elseif fst.typ === Block || fst.typ === Brackets || fst.typ === Filter
         past_if = false
--- a/src/fst.jl
+++ b/src/fst.jl
@@ -804,17 +804,13 @@ function eq_to_in_normalization!(fst::FST, always_for_in::Bool, for_in_replaceme
         end
         if !isnothing(fst.metadata)
             metadata = fst.metadata::Metadata
-            opkind = try
-                JuliaSyntax.Kind(op.val)
-            catch
-                metadata.op_kind
-            end
+            opkind = JuliaSyntax.Kind(op.val)
             fst.metadata = Metadata(
                 opkind,
                 metadata.op_dotted,
                 metadata.is_standalone_shortcircuit,
                 metadata.is_short_form_function,
-                op.val == "=",
+                opkind === K"=",
                 metadata.is_long_form_function,
                 metadata.has_multiline_argument,
             )
--- a/src/fst.jl
+++ b/src/fst.jl
@@ -259,7 +259,7 @@ is_comment(fst::FST) = fst.typ in (INLINECOMMENT, NOTCODE, HASHEQCOMMENT)
 
 is_identifier(x) = kind(x) === K"Identifier" && !haschildren(x)
 
-is_ws(x) = kind(x) in KSet"Whitespace NewlineWs"
+is_ws(x) = JuliaSyntax.is_whitespace(x)
 
 function is_multiline(fst::FST)
     fst.endline > fst.startline &&
@@ -505,16 +505,6 @@ function is_opcall(x::JuliaSyntax.GreenNode)
         return is_opcall(childs[idx])
     end
     return false
-end
-
-function is_prefix_op_call(x::JuliaSyntax.GreenNode)
-    is_opcall(x) || return false
-    haschildren(x) || return false
-
-    idx = findfirst(!JuliaSyntax.is_whitespace, children(x))
-    isnothing(idx) && return false
-
-    return JuliaSyntax.is_operator(x[idx])
 end
 
 function is_gen(x::JuliaSyntax.GreenNode)
--- a/src/fst.jl
+++ b/src/fst.jl
@@ -525,23 +525,6 @@ function is_gen(x::FST)
     x.typ in (Generator, Filter, Flatten)
 end
 
-function extract_operator_indices(childs::Vector{JuliaSyntax.GreenNode{T}}) where {T}
-    args = findall(n -> !JuliaSyntax.is_whitespace(n), childs)
-    op_indices = Int[]
-    i = 2
-    while i <= length(args)
-        push!(op_indices, args[i])
-        if i < length(args) &&
-           kind(childs[args[i]]) === K"." &&
-           !haschildren(childs[args[i]])
-            push!(op_indices, args[i+1])
-            i += 1
-        end
-        i += 2
-    end
-    return op_indices
-end
-
 function _callinfo(x::JuliaSyntax.GreenNode)
     if !haschildren(x)
         return 0, 0
--- a/src/styles/default/pretty.jl
+++ b/src/styles/default/pretty.jl
@@ -137,8 +137,8 @@ function do_block_index(childs::Vector{JuliaSyntax.GreenNode{T}}) where {T}
 end
 
 function has_do_block_call(cst::JuliaSyntax.GreenNode)
-    kind(cst) in KSet"call dotcall" && haschildren(cst) || return false
-    !isnothing(do_block_index(children(cst)))
+    kind(cst) in KSet"call dotcall" && haschildren(cst) || return nothing
+    do_block_index(children(cst))
 end
 
 function call_args(childs::Vector{JuliaSyntax.GreenNode{T}}) where {T}
@@ -174,6 +174,7 @@ function pretty(
 )::FST
     k = kind(node)
     style = getstyle(ds)
+    do_block_idx = has_do_block_call(node)
     push!(lineage, (k, is_iterable(node), is_assignment(node)))
 
     ret = if k == K"Identifier" && !haschildren(node)
@@ -272,8 +273,8 @@ function pretty(
     elseif k === K"?" && haschildren(node)
         p_conditionalopcall(style, node, s, ctx, lineage)
     # Example: `map(xs) do x; x + 1; end` is a call node with a do child.
-    elseif has_do_block_call(node)
-        p_do_call(style, node, s, ctx, lineage)
+    elseif !isnothing(do_block_idx)
+        p_do_call(style, node, s, ctx, lineage, do_block_idx)
     # Example: `+(x)` parses as a prefix-op call.
     elseif JuliaSyntax.is_prefix_op_call(node)
         p_unaryopcall(style, node, s, ctx, lineage)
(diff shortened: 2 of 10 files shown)


svchb:
@domluna all tests pass locally

src/styles/default/pretty.jl
```diff
@@ -2129,21 +2409,32 @@ function p_call(
     s::State,
     ctx::PrettyContext,
     lineage::Vector{Tuple{JuliaSyntax.Kind,Bool,Bool}},
+    ;
```
pfitzseb: Formatting is weird here.

pfitzseb (commented):
Generally LGTM

svchb:
@penelopeysm can you review this?

penelopeysm:
I could, but honestly, it would take me a bit of time! I haven't really gotten into the codebase as much as I'd like, and have been quite busy IRL too. If you don't mind waiting, I could try to get to it over the weekend?

svchb:
Sure no worries as long as we get this project moved forward again 🚂

src/fst.jl
```diff
@@ -646,6 +661,13 @@ function is_function_or_macro_def(cst::JuliaSyntax.GreenNode)
     return false
 end
 
+function is_short_function_def(cst::JuliaSyntax.GreenNode)
+    kind(cst) === K"function" && haschildren(cst) || return false
+    idx = findfirst(n -> !JuliaSyntax.is_whitespace(n), children(cst))
+    isnothing(idx) && return false
+    return kind(cst[idx]) !== K"function"
+end
```
pfitzseb, line 641:
```suggestion
is_short_function_def(cst::JuliaSyntax.GreenNode) = JuliaSyntax.has_flags(cst, JuliaSyntax.SHORT_FORM_FUNCTION_FLAG)
```

src/styles/default/pretty.jl
```diff
@@ -118,6 +268,12 @@ function pretty(
         p_whereopcall(style, node, s, ctx, lineage)
     elseif k === K"?" && haschildren(node)
         p_conditionalopcall(style, node, s, ctx, lineage)
+    elseif has_do_block_call(node)
+        p_do_call(style, node, s, ctx, lineage)
```
pfitzseb: With this, does the
```
    elseif k === K"do"
        p_do(style, node, s, ctx, lineage)
```
ever get called? Seems like only having one of them should be enough (and specifically, why can't you re-use the existing case here)?

src/styles/default/pretty.jl
```diff
@@ -21,6 +21,151 @@ function newctx(s::PrettyContext; kwargs...)
...
+    elseif kind(cst) === K"Identifier" && !haschildren(cst)
+        k = source_kind(s, cst, offset)
+        if !isnothing(k) && JuliaSyntax.is_operator(k)
+            return k
+        end
+    end
+    return nothing
+end
+
+function is_source_operator(s::State, cst::JuliaSyntax.GreenNode, offset::Integer)
```
pfitzseb, line 52: In theory this should be covered by the `JuliaSyntax.is_operator_*` calls no? Would be nice if we didn't have to look into the file for this.

src/styles/default/pretty.jl
```diff
@@ -21,6 +21,151 @@ function newctx(s::PrettyContext; kwargs...)
...
+    if JuliaSyntax.is_operator(cst) && !haschildren(cst)
+        return kind(cst)
+    elseif kind(cst) === K"Identifier" && !haschildren(cst)
+        k = source_kind(s, cst, offset)
+        if !isnothing(k) && JuliaSyntax.is_operator(k)
+            return k
+        end
+    end
+    return nothing
+end
```
pfitzseb, line 50: We need this because of https://github.com/JuliaLang/JuliaSyntax.jl/issues/591, right? Wonder whether this will get fixed upstream at some point...

src/styles/default/pretty.jl
```diff
@@ -21,6 +21,151 @@ function newctx(s::PrettyContext; kwargs...)
...
+        offset += Int(span(c))
+    end
+    return opkind
+end
+
+function update_operator_indices(childs::Vector{JuliaSyntax.GreenNode{T}}) where {T}
+    args = findall(n -> !JuliaSyntax.is_whitespace(n), childs)
+    length(args) < 4 && return Int[]
+    return args[2:(end-1)]
+end
```
pfitzseb: Seems like the operator should be easy enough to find by checking in which position it occurs (`JS.is_[prefix|infix|suffix]_op_call`) and then directly accessing that argument.

pfitzseb (commented):
Ok, finally got a chance to take a proper look at this. It would generally be nice to get some more unit tests for the new utility functions if possible.

src/styles/default/pretty.jl
```diff
@@ -21,6 +21,151 @@ function newctx(s::PrettyContext; kwargs...)
...
+           !haschildren(childs[args[1]])
+            args[2]
+        else
+            args[1]
+        end
+    offset = s.offset + sum(span, childs[1:(op_arg-1)]; init = 0)
+    return is_source_operator(s, childs[op_arg], offset) ? op_arg : nothing
+end
+
+function is_source_prefix_op_call(cst::JuliaSyntax.GreenNode, s::State)
```
penelopeysm: I think this entire function can just be replaced with `JuliaSyntax.is_prefix_op_call(cst)`. I ran the test suite with this change and it all passes

src/styles/default/pretty.jl
penelopeysm: I guess this is an intentional choice in JuliaSyntax

https://github.com/JuliaLang/JuliaSyntax.jl/issues/474
https://github.com/JuliaLang/JuliaSyntax.jl/pull/523

So I think we have to work around it somehow. As mentioned in that PR, `JuliaSyntax.tokenize()` has a flag to retain the information, but we aren't using that, and doing so would probably be a large refactor. So I'm happy to use this approach of re-parsing.

(I think the JuliaFormatter codebase needs more comments in general to explain things like this, but that's a separate issue.)

src/styles/default/pretty.jl
```diff
@@ -118,6 +268,12 @@ function pretty(
         p_whereopcall(style, node, s, ctx, lineage)
     elseif k === K"?" && haschildren(node)
         p_conditionalopcall(style, node, s, ctx, lineage)
+    elseif has_do_block_call(node)
+        p_do_call(style, node, s, ctx, lineage)
+    elseif is_source_prefix_op_call(node, s)
+        p_unaryopcall(style, node, s, ctx, lineage)
```
penelopeysm:
```suggestion
    elseif JuliaSyntax.is_prefix_op_call(node)
        p_unaryopcall(style, node, s, ctx, lineage)
```

src/styles/default/pretty.jl
```diff
@@ -118,6 +268,12 @@ function pretty(
...
+    elseif !isnothing(source_prefix_operator_index(node, s))
+        p_call(style, node, s, ctx, lineage)
```
penelopeysm: I'm pretty sure that this dispatch is not necessary -- they all fall through to the other branches correctly. For example, I think the intent of this branch is to catch something like `">=(x)"`? (Please correct me if I'm wrong.) However, this doesn't actually get picked up by `is_unary` (because, since JuliaSyntax v1, the kind of the first token is `Identifier`), so it falls through all the way to `is_func_call`, which is exactly what we want. Again, I checked with this change and the test suite passes.

```suggestion
```

src/styles/default/pretty.jl
penelopeysm: I think we should rename this function to `p_iteration` to avoid confusion (since there's no longer a cartesian iterator kind in JuliaSyntax v1, and this function handles all types of iterators regardless of whether there are multiple of them).

src/styles/default/pretty.jl
```diff
@@ -1462,8 +1626,14 @@ function p_cartesian_iterator(
...
 
     childs = children(cst)
+    has_comma = iteration_has_comma(cst)
     for (i, c) in enumerate(childs)
         n = pretty(style, c, s, ctx, lineage)
+        if has_comma && kind(c) in KSet"in ∈" && is_iterable(iteration_rhs(c))
+            if n.typ === Binary && n[end].typ === Vect
+                n[end].nest_behavior = AlwaysNest
+            end
+        end
```
penelopeysm: I don't think I fully understand the purpose of this yet. Why do we need to force nesting, and if so why are vectors special-cased and not (say) tuples or function calls? It doesn't seem needed for any tests afaict (even though it _does_ change the output of something like `for i in xs, j in [1, 2, 3]; g; end`, none of the tests seem to rely on its behaviour).

src/styles/default/pretty.jl
```diff
@@ -1534,6 +1702,48 @@ function p_do(
...
+    cst::JuliaSyntax.GreenNode,
+    s::State,
+    ctx::PrettyContext,
+    lineage::Vector{Tuple{JuliaSyntax.Kind,Bool,Bool}},
+)
+    t = FST(Do, nspaces(s))
+    childs = children(cst)
+    idx = do_block_index(childs)
+    if isnothing(idx)
+        return p_call(ds, cst, s, ctx, lineage)
```
penelopeysm:
```suggestion
        error("p_do_call: expected do block")
```

This code is unreachable because `p_do_call` should only ever be called if there is a do block -- I think it's better to loudly error (so that any bugs are surfaced) than to fall back which might silently yield results we're not keen on.

Though I would say it is even better if we make `has_do_block_call` return the index that we want, and then pass it into `p_do_call`.

src/styles/default/pretty.jl
```diff
@@ -1688,9 +1898,13 @@ function p_kw(
             add_node!(t, pretty(style, c, s, ctx, lineage), s; join_lines = true)
             add_node!(t, Whitespace(1), s)
         else
+            source_prefix = !isnothing(source_prefix_operator_index(c, s))
```
penelopeysm, line 1938:
```suggestion
            # Need to explicitly check source_prefix rather than relying on
            # JuliaSyntax.is_prefix_op_call, as the latter doesn't identify e.g.
            # `>=(x)` as operators.
            source_prefix = !isnothing(source_prefix_operator_index(c, s))
```

Just a comment to make life easier for future readers.

src/styles/default/pretty.jl
```diff
@@ -1688,9 +1898,13 @@ function p_kw(
...
             n = pretty(style, c, s, ctx, lineage)
-            if !s.opts.whitespace_in_kwargs &&
-               ((n.typ === IDENTIFIER && endswith(n.val, "!")) || (is_prefix_op_call(c)))
+            if !s.opts.whitespace_in_kwargs && (
+                (n.typ === IDENTIFIER && endswith(n.val, "!")) ||
+                is_prefix_op_call(c) ||
```
penelopeysm:
```suggestion
```

This function actually always return false with JuliaSyntax v1 so is no longer needed (once this call point is removed, the function can also be eliminated).

The `source_prefix` check will pick up all the things we need to pick up.

src/fst.jl
penelopeysm: This function is now dead code (and gives the wrong behaviour anyway with JuliaSyntax v1 because prefix ops are parsed as `K"Identifier"`) so can be removed.

src/styles/default/pretty.jl
```diff
@@ -1778,16 +2003,23 @@ function p_binaryopcall(
     end
 
     from_colon = ctx.from_colon
+    from_typedef = ctx.from_typedef || any(x -> x[1] === K"where", lineage)
```
penelopeysm: Tests seem to pass without this `any(...)` check. What's the rationale for adding it?

src/styles/default/pretty.jl
```diff
@@ -2128,22 +2408,32 @@ function p_call(
     cst::JuliaSyntax.GreenNode,
     s::State,
     ctx::PrettyContext,
-    lineage::Vector{Tuple{JuliaSyntax.Kind,Bool,Bool}},
+    lineage::Vector{Tuple{JuliaSyntax.Kind,Bool,Bool}};
+    child_limit::Union{Int,Nothing} = nothing,
```
penelopeysm: I think this keyword argument is pretending to be more general than it really needs to be. Looking at where it gets used, it really just means 'one before the do block, if a do block exists'. I'd suggest we could just pass in the do block index, which would be clearer, and would also avoid recalculating it inside this function.

src/fst.jl
```diff
@@ -813,6 +835,23 @@ function eq_to_in_normalization!(fst::FST, always_for_in::Bool, for_in_replaceme
             op.val = "="
             op.len = length(op.val)
         end
+        if !isnothing(fst.metadata)
+            metadata = fst.metadata::Metadata
+            opkind = try
+                JuliaSyntax.Kind(op.val)
+            catch
+                metadata.op_kind
+            end
```
penelopeysm: `Kind(op.val)` can't fail because `op.val` can only be one of `in`, `=` or `∈` so I think we can elide the try/catch.

src/fst.jl
```diff
@@ -813,6 +835,23 @@ function eq_to_in_normalization!(fst::FST, always_for_in::Bool, for_in_replaceme
...
+            fst.metadata = Metadata(
+                opkind,
+                metadata.op_dotted,
+                metadata.is_standalone_shortcircuit,
+                metadata.is_short_form_function,
+                op.val == "=",
```
penelopeysm:
```suggestion
                opkind === K"=",
```

penelopeysm (changes requested):
Thanks @svchb and sorry it took a while! I'm familiar enough with parsing but I had to figure out all the exact data structures that JuliaFormatter and JuliaSyntax use.

I'm conscious that bumping compat for JuliaSyntax has been a longstanding issue on JuliaFormatter. So I think that we should just try to merge this in as long as it works. Although I think that there are quite a few things that ought to be cleaned up (especially with the helper functions that do similar things to JuliaSyntax), we can defer these.

However I still do have some questions / comments which hopefully won't hold things up too much.

src/styles/default/pretty.jl
penelopeysm: That branch is still hit for macros like `@m(x) do y; false; end`. To be honest I didn't even know this was a valid syntactic construct until now

svchb: I have added comments to the new items in pretty()

src/styles/default/pretty.jl
svchb: No this changed in JuliaSyntax v1:
```
using JuliaSyntax

text = "a + b"
root = JuliaSyntax.parseall(JuliaSyntax.GreenNode, text)
node = only(filter(!JuliaSyntax.is_whitespace, JuliaSyntax.children(root)))
op = JuliaSyntax.children(node)[3]

@show Base.pkgversion(JuliaSyntax)
@show kind(op)
@show JuliaSyntax.is_infix_op_call(node)
@show JuliaSyntax.is_operator(op) <- true with <v1, false with v1+
@show text[3:3]
```

src/styles/default/pretty.jl
svchb: yes this changed in v1. See also above.

src/styles/default/pretty.jl
pfitzseb: Yeah, I know that `is_operator` isn't what we want here, but a `is_operator2(cst) = is_infix_op_call(cst) || is_prefix_op_call(cst) || is_suffix_op_call(cst)` should recover that functionality.

svchb:
```
is_operator2(cst) =
    JuliaSyntax.is_infix_op_call(cst) ||
    JuliaSyntax.is_prefix_op_call(cst) ||
    JuliaSyntax.is_postfix_op_call(cst)
```

```
text = "a + b"
node = only(filter(!JuliaSyntax.is_whitespace,
    JuliaSyntax.children(JuliaSyntax.parseall(JuliaSyntax.GreenNode, text))))

is_operator2(node) # true
```

```
op = JuliaSyntax.children(node)[3]

kind(op)           # K"Identifier"
is_operator2(op)   # false
text[3:3]          # "+"
```

pfitzseb: Yeah, it wouldn't be a drop-in replacement and need to be used on the parent node:
```
julia> cst = JuliaSyntax.parseall(JuliaSyntax.GreenNode, "a+b"); cst[1]
     1:3      │[call]
     1:1      │  Identifier             ✔
     2:2      │  Identifier             ✔
     3:3      │  Identifier             ✔


julia> is_operator2(cst[1])
true
```

src/styles/default/pretty.jl
svchb: removed

src/styles/default/pretty.jl
svchb: removed

src/styles/default/pretty.jl
```diff
@@ -72,74 +72,79 @@ function source_prefix_operator_index(cst::JuliaSyntax.GreenNode, s::State)
...
+        return args[2:(end-1)]
+    elseif kind(cst) === K"comparison"
+        length(args) < 3 && return Int[]
+        return args[2:2:(end-1)]
+    elseif is_short_function_def(cst)
+        length(args) < 3 && return Int[]
+        return Int[args[2]]
+    elseif JuliaSyntax.is_operator(cst)
+        length(args) < 3 && return Int[]
+        return args[2:(end-1)]
```
penelopeysm: I think I'd prefer to be bit more aggressive here and error loudly. This passes the test suite (including formatting :-))

```julia
    nonws_args = length(args)
    nonws_args == 0 &&
        error("no non-whitespace children found for operator node; should not happen")

    if kind(cst) === K"op="
        # 4 args for `x += y`, 5 args for `x .+= y` (the dot is preserved as a flag in
        # JuliaSyntax but the kind is the same)
        (nonws_args == 4 || nonws_args == 5) ||
            error("unexpected number of args for op= node", nonws_args, cst)
        return args[2:(end-1)]
    elseif kind(cst) === K"comparison"
        # x < y < z < ... < w has 2n - 1 args where n >= 2
        (iseven(nonws_args) || nonws_args < 3) &&
            error("unexpected number of args for comparison node", nonws_args, cst)
        return args[2:2:(end-1)]
    elseif is_short_function_def(cst)
        nonws_args != 3 && error(
            "unexpected number of args for short function definition node",
            nonws_args,
            cst,
        )
        return Int[args[2]]
    elseif JuliaSyntax.is_operator(cst)
        length(args) != 3 &&
            error("unexpected number of args for operator node", length(args), cst)
        return args[2:(end-1)]
```

src/styles/default/pretty.jl
```diff
@@ -33,6 +33,10 @@ function source_kind(s::State, cst::JuliaSyntax.GreenNode, offset::Integer)
     end
 end
 
+# JuliaSyntax v1 can encode source operators as Identifier leaves in call
+# forms, e.g. `a + b`, `x .+ y`, `+(x)`, and `>=(x)`. The parent call flags
+# describe the shape, but the leaf itself is not `is_operator`, and the exact
+# operator kind is only recoverable from the source span.
 function source_operator_kind(s::State, cst::JuliaSyntax.GreenNode, offset::Integer)
```
penelopeysm: We now have two functions, `source_op_kind` and `source_operator_kind`, which is a bit confusing. Looking at the implementation, the difference seems to be that `source_operator_kind` takes the offset, so maybe we can call this something like `source_op_kind_from_offset`?

penelopeysm (commented):
Thanks, just a couple of notes!

src/styles/default/pretty.jl
penelopeysm: (I'd have made a suggestion, but for some reason the GitHub UI doesn't let me, sorry)

penelopeysm (approved):
I'm happy with it!

penelopeysm:
I'll bump the minor version (just in case we need to backport changes) and make a release in a moment.

PR #984, opened 2026-05-13
Fix CI warnings, make cleanups and improvements to workflows

digital-carver:
Fix the [various warnings](https://github.com/JuliaEditorSupport/JuliaFormatter.jl/actions/runs/25787399735) when running CI, and make some additional improvements.

* Allow manual CI runs with `workflow_dispatch`
* Give each job a 2 hour timeout duration (instead of the default 6 hours)
* `macos-15` is ARM, not Intel, so set the Intel one to the correct `macos-15-intel`
    * This also allows us to upgrade setup-julia from 2 to 3 which was [failing due to this error](https://github.com/JuliaEditorSupport/JuliaFormatter.jl/pull/980) - or rather, aborting redundant runs, since there's already a separate `macos-latest - aarch64` which was the same runner. (Subsumes https://github.com/JuliaEditorSupport/JuliaFormatter.jl/pull/980 )
* Remove an unnecessary `exclude` of `x86` which wasn't included in the first place
* Upgrade `actions/checkout` to v6 (subsumes https://github.com/JuliaEditorSupport/JuliaFormatter.jl/pull/955 )
* Make use of `julia-actions/cache` in CI too, not just PreCommit
* In PreCommit.yaml:
    * Upgrade setup-uv action to v8 and use exact version as [recommended](https://github.com/astral-sh/setup-uv/releases/tag/v8.0.0)
    * Remove `pre-commit install` automatic git hook setup, as we're explicitly running `pre-commit run` and not depending on any automatic hooks
    * Subsume https://github.com/JuliaEditorSupport/JuliaFormatter.jl/pull/981 for cache action upgrade

### DIFF
--- a/.github/workflows/CompatHelper.yml
+++ b/.github/workflows/CompatHelper.yml
@@ -1,8 +1,10 @@
 name: CompatHelper
+
 on:
   schedule:
     - cron: 0 0 * * *
   workflow_dispatch:
+
 jobs:
   CompatHelper:
     runs-on: ubuntu-latest
--- a/.github/workflows/PreCommit.yml
+++ b/.github/workflows/PreCommit.yml
@@ -16,17 +16,19 @@ on:
 jobs:
   pre-commit:
     runs-on: ubuntu-latest
+    timeout-minutes: 120
+
     steps:
       - uses: actions/checkout@v6
-      - uses: julia-actions/setup-julia@v2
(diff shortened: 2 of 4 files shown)


penelopeysm (approved):

penelopeysm:
This is great!

PR #985, opened 2026-05-15
Some tidying up

penelopeysm:
- Reverts the v3.0 bump for a `jlfmt` based pre-commit workflow (it's probably easier to release this as a distinct pre-commit hook, that doesn't require a major version bump, which I think would be quite disruptive)

- Add a CI test for the pre-commit hook, to catch any regressions in behaviour.

- Update links to point to new GitHub organisation. They currently redirect but this is not technically guaranteed

- Remove HELP.md since there is help now (hopefully :))


.github/workflows/PreCommit.yml
penelopeysm, line 1: @digital-carver I took this one from your PR, #984, if that's OK, I had to change it anyway and hopefully it'll make it easier to handle merge conflicts.

penelopeysm (commented):
I was thinking that we need to revert the major version bump first before making new fixes, since I'm not very keen on releasing a 3.x of JuliaFormatter. Once this is done, we can merge and release #983 as well :)

.github/workflows/PreCommitHooks.yml
```diff
@@ -0,0 +1,46 @@
+# This workflow tests that the pre-commit hook works as expected.
```
penelopeysm, line 1: Since I reverted the jlfmt pre-commit hook, we only have the 'old' hook to test, i.e. the one that calls `julia -e 'using JuliaFormatter' ...`. Once we add the jlfmt hook back, this will be useful for also testing its behaviour.

.pre-commit-hooks.yaml
```diff
@@ -1,6 +1,6 @@
 - id: julia-formatter
   name: "Julia Formatter"
-  entry: "jlfmt --inplace"
+  entry: "julia -e 'import JuliaFormatter: format; format(ARGS)'"
```
penelopeysm, line 3: cf. https://github.com/JuliaEditorSupport/JuliaFormatter.jl/issues/968.

.github/workflows/PreCommitHooks.yml
```diff
@@ -0,0 +1,46 @@
...
+      - uses: actions/setup-python@v6
+        with:
+          python-version: '3.x'
+
+      - name: Test pre-commit hook
+        working-directory: .github/workflows/pre-commit
+        run: |
+          # Run pre-commit hook, check that it fails.
+          # pre-commit try-repo will fail if the hook makes changes
+          pipx run pre-commit try-repo ../../../ julia-formatter --files main.jl --verbose && exit 1
```
pfitzseb, line 40: Any reason this uses a different way to set up pre-commit than `PreCommit.yml`?

.github/workflows/pre-commit/main.jl
pfitzseb, line 1: Any way to check something in the test dir instead? It seems a bit weird to put a test fixture here.

.github/workflows/PreCommitHooks.yml
penelopeysm: I would argue PreCommit.yml does not need uv because pipx is available on github runners, so this is less bloated. Can unify both if preferred

.github/workflows/pre-commit/main.jl
penelopeysm: Hmm, I don't find it weird: I've done this before in other projects. It's a fixture just for github actions, it's not related to the actual test suite.

.github/workflows/PreCommitHooks.yml
pfitzseb: Yeah that's fair. I don't have any opinion on which to use, but using the same installation method in both actions is preferable imho.

.github/workflows/pre-commit/main.jl
pfitzseb: Alright. I think this is just a pattern I've never seen before and that's why it seems weird to me.

.github/workflows/PreCommitHooks.yml
penelopeysm: Done :)

pfitzseb (approved):

PR #989, opened 2026-05-24
Fix multiline strings & display width interaction

### LINKED ISSUE #974: JuliaFormatter v2.3.0 does not "converge" for some line continued expressions with anonymous functions
If I format the following Julia code using `JuliaFormatter` v2.3.0, one space is removed from the second line every time I apply the formatter:

```julia
error_str = Δa -> "Change in activity must be within the valid range: \
                    Δa ∈ [1, 2], but Δa = $Δa"
```

Consequently, JuliaFormatter (e.g. set up as a CI job) will request changes every time until there are no spaces left and the code looks like:

```julia
error_str = Δa -> "Change in activity must be within the valid range: \
Δa ∈ [1, 2], but Δa = $Δa"
```

penelopeysm:
Closes #974

If a multiline string begins on a line where the byte width is not equal to the display width, this caused the indent for subsequent lines to be calculated incorrectly. This fixes it by using the source to calculate the display width for the contents of the first line up to the opening quote.

The above is really just a one-line fix. But I also took the liberty to add some comments to the parts which I changed. I think there's a balance to be struck between complete absence of comments and over-commenting, but I believe that (1) right now we're too close to the former, and (2) it's better to err on the side of the latter for any compiler-like project that relies on representation of programmes, because there tend to be a ton of edge cases and weird quirks.

[Fun fact: this PR still fails with emojis.](https://github.com/JuliaEditorSupport/JuliaFormatter.jl/issues/991) However, this is a bigger problem that requires a larger refactor, so I don't want to lump it in here.

### DIFF
--- a/test/default_style.jl
+++ b/test/default_style.jl
@@ -1433,6 +1433,9 @@
 
         @testset "multiline string starting on line with display width != bytes" begin
             # https://github.com/JuliaEditorSupport/JuliaFormatter.jl/issues/974
+            #
+            # Anonymous function form. Test both the case where the second line begins
+            # after the starting quote of the first line, and the opposite.
             s1a = raw"""
             error_str = Δa -> "Change in activity must be within the valid range: \
                                 Δa ∈ [1, 2], but Δa = $Δa"
@@ -1441,6 +1444,7 @@
             error_str = Δa -> "Change in activity must be within the valid range: \
                   Δa ∈ [1, 2], but Δa = $Δa"
             """
+            # Explicit function definition
             s2a = raw"""
             error_str(Δa) = "Change in activity must be within the valid range: \
                                 Δa ∈ [1, 2], but Δa = $Δa"
@@ -1449,7 +1453,16 @@
             error_str(Δa) = "Change in activity must be within the valid range: \
               Δa ∈ [1, 2], but Δa = $Δa"
             """
-            for s in (s1a, s1b, s2a, s2b)
+            # Using triple quoted strings
+            s3a = raw"""
+            error_str(Δa) = \"""Change in activity must be within the valid range: \
+                                Δa ∈ [1, 2], but Δa = $Δa\"""
+            """
+            s3b = raw"""
+            error_str(Δa) = \"""Change in activity must be within the valid range: \
+              Δa ∈ [1, 2], but Δa = $Δa\"""
+            """
+            for s in (s1a, s1b, s2a, s2b, s3a, s3b)
                 # Need to normalise line endings because \ at the end of a raw_str behaves
                 # weirdly on Windows causing CI to fail.
                 # https://github.com/JuliaLang/julia/issues/38908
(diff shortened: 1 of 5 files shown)


penelopeysm:
Fun fact: this PR still fails with emojis.

```julia
julia> using JuliaFormatter: format_text; s = "e = 🙃 -> \"x\\\n   y\"";

julia> s |> println
e = 🙃 -> "x\
   y"

julia> format_text(s) |> println
e = 🙃 -> "x\
  y"
```

However, this is a bigger problem that requires a larger refactor, so I don't want to lump it in here. Essentially `State.line_offset` is tracking `length(x)` rather than `textwidth(x)`.

penelopeysm:
@pfitzseb, maybe I should take this opportunity to ask how much you want to review JuliaFormatter code. Either way is fine, and I totally understand if it's too much of a time commitment -- I did volunteer to maintain the codebase, so I'm fairly comfortable with a world where I self-review / merge, and bear the responsibility for occasionally mucking things up :-) (I pretty much did that on Turing.jl for a few months)

test/default_style.jl
```diff
@@ -1431,6 +1431,33 @@
...
+                  Δa ∈ [1, 2], but Δa = $Δa"
+            """
+            s2a = raw"""
+            error_str(Δa) = "Change in activity must be within the valid range: \
+                                Δa ∈ [1, 2], but Δa = $Δa"
+            """
+            s2b = raw"""
+            error_str(Δa) = "Change in activity must be within the valid range: \
+              Δa ∈ [1, 2], but Δa = $Δa"
+            """
```
pfitzseb, line 1455:
```suggestion
            s2b = raw"""
            error_str(Δa) = \"\"\"Change in activity must be within the valid range: \
              Δa ∈ [1, 2], but Δa = $Δa\"\"\"
            """
```
AFAICT JuliaFormatter currently handles `"` and `"""` multi-line strings the same, but maybe that'll change at some point (since they have very different semantics when it comes to leading whitespace), so a regression test for both cases can't hurt.

pfitzseb (approved):
The fix LGTM.

I have my [gripes](https://github.com/JuliaEditorSupport/JuliaFormatter.jl/issues/501) with how multi-line string indentation is handled, but that's not particularly relevant here :D

pfitzseb:
> so I'm fairly comfortable with a world where I self-review / merge, and bear the responsibility for occasionally mucking things up

Fine by me :) I'm happy to chime in with opinions/PR reviews, so feel free to tag me if you think it'd be useful. The main thing I want to get over the line is the JuliaSyntax v1 PR, because that's needed to finally get the language server running a newer version of JuliaFormatter.

penelopeysm:
Ah! The docstring indentation does strike me as counterintuitive. I see Dom's points, but I think my tastes are similar to yours haha.

test/default_style.jl
penelopeysm: Sounds good, let me add one in

PR #992, opened 2026-05-25
Add jlfmt pre-commit hook

### LINKED ISSUE #633: Formatting is very slow
I've just started trying out this package, and it's extremely slow. I just wanted to be sure I wasn't doing something wrong.

For a simple file, `print("hello, world")`, the following command takes over 20 seconds:

```sh
julia -e 'using JuliaFormatter; format_file("test.jl")'
```

I'm using the version in the arch linux repos (`julia version 1.7.3`).

### LINKED ISSUE #990: Reinstate jlfmt-based pre-commit hook...
... but make it a separate hook, so as to avoid breakage.

penelopeysm:
Closes #633
Closes #990

This is essentially the same as https://github.com/JuliaEditorSupport/JuliaFormatter.jl/pull/959, but with some changes:

- It's a separate pre-commit hook, rather than replacing the old one, which would have been a breaking change.

  It's a minor annoyance that the hooks are now tied to the repo and that means we can't have breaking changes in hooks without also releasing a breaking change in the repo. Runic sidesteps this by [having a separate repo for the hooks](https://github.com/fredrikekre/runic-pre-commit). However, we can live with it.

- Instead of calling `jlfmt` directly, it calls a shell script which prints a nice error if jlfmt isn't found, and additionally allows a user to pass a custom path to a jlfmt executable if so desired. (It struck me that since jlfmt's CLI arguments were designed to be similar to Runic's, one could go rogue and instead give it the path to a Runic executable lol)

- I've basically rewritten the docs and added it to the sidebar (for some reason it wasn't being displayed there). I also took the opportunity to add docs on using PackageCompiler to speed up time-to-first-format.

### DIFF
--- a/.github/workflows/PreCommitHooks.yml
+++ b/.github/workflows/PreCommitHooks.yml
@@ -11,7 +11,7 @@ on:
       - 'master'
 
 jobs:
-  pre-commit:
+  pc-julia-formatter:
     name: julia-formatter hook
     runs-on: ubuntu-latest
 
@@ -44,3 +44,44 @@ jobs:
 
           # Check that running the hook again succeeds.
           pipx run pre-commit try-repo ../../../ julia-formatter --files main.jl --verbose
+
+
+  pc-jlfmt:
+    name: jlfmt hook
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          persist-credentials: false
+
+      - uses: julia-actions/setup-julia@v3
+        with:
+          version: '1'
+
+      - name: Install jlfmt app
+        # Need to explicitly add registry because of
+        # https://github.com/JuliaLang/Pkg.jl/issues/4360 which is not
+        # backported to 1.12 yet
+        run:
+          julia -e 'using Pkg; Pkg.Registry.add("General"); Pkg.Apps.add(; path=pwd())'
+
+      - uses: actions/setup-python@v6
+        with:
+          python-version: '3.x'
+
+      - name: Test pre-commit hook
+        working-directory: .github/workflows/pre-commit
+        run: |
+          # Make sure jlfmt app can be found
+          export PATH="$HOME/.julia/bin:$PATH"
+
+          # Run pre-commit hook, check that it fails.
+          # pre-commit try-repo will fail if the hook makes changes
+          pipx run pre-commit try-repo ../../../ jlfmt --files main.jl --verbose && exit 1
+
+          # Check that the file was changed.
+          git diff --exit-code main.jl && exit 1
+
+          # Check that running the hook again succeeds.
+          pipx run pre-commit try-repo ../../../ jlfmt --files main.jl --verbose
--- a/.github/workflows/PreCommitHooks.yml
+++ b/.github/workflows/PreCommitHooks.yml
@@ -73,11 +73,42 @@ jobs:
       - name: Test pre-commit hook
         working-directory: .github/workflows/pre-commit
         run: |
-          # Make sure jlfmt app can be found
+          # At this point, jlfmt won't be on PATH since the GitHub runner
+          # doesn't automatically add Julia apps to PATH. This is intentional,
+          # since it lets us check that there is a nice error message if jlfmt
+          # is not on PATH.
+          output=$(pipx run pre-commit try-repo ../../../ jlfmt --files main.jl --verbose 2>&1 || true)
+          echo "$output" | grep -q "make sure that jlfmt is on your PATH" || { echo "Expected graceful error message when jlfmt was not found"; exit 1; }
+
+          # Now make sure jlfmt app can be found
+          OLDPATH=$PATH
           export PATH="$HOME/.julia/bin:$PATH"
 
           # Run pre-commit hook, check that it fails.
           # pre-commit try-repo will fail if the hook makes changes
+          pipx run pre-commit try-repo ../../../ jlfmt --files main.jl --verbose && exit 1
+
+          # Check that the file was changed.
+          git diff --exit-code main.jl && exit 1
+
+          # Check that running the hook again succeeds.
+          pipx run pre-commit try-repo ../../../ jlfmt --files main.jl --verbose
+
+          # Restore PATH for future steps
+          export PATH=$OLDPATH
+
+      - name: Test pre-commit hook with --jlfmt-path argument
+        working-directory: .github/workflows/pre-commit
+        run: |
+          # Reset the test file
+          git checkout main.jl
+
+          # Add --jlfmt-path arg to the existing jlfmt hook. pre-commit
+          # try-repo doesn't allow passing hook arguments via the command line,
+          # so we inject it directly into .pre-commit-hooks.yaml. A tad ugly.
+          sed -i '/^- id: jlfmt$/a\  args: ["--jlfmt-path=~/.julia/bin/jlfmt"]' ../../../.pre-commit-hooks.yaml
+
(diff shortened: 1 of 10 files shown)


penelopeysm:
@gdalle Since you did the last PR, could I trouble you to test this out and see if there are any issues?

```yaml
repos:
- repo: https://github.com/JuliaEditorSupport/JuliaFormatter.jl
  rev: 3b196afa4a8ed92ba816edec6d122145e311ae35 # This PR.
  hooks:
  - id: "jlfmt"
```

penelopeysm:
~~... Never mind, it fell over at the first step for me. Let me debug a bit more~~

Okay, that should work!

gdalle:
> It's a minor annoyance that the hooks are now tied to the repo and that means we can't have breaking changes in hooks without also releasing a breaking change in the repo.

I'm not even sure what a "breaking change in the _repo_" would mean. As far as the Julia package is concerned, the tags on the repo don't really matter, it's the content of the `Project.toml` that matters at the time of registration. And for the rest, we don't _have_ to follow SemVer (although it's arguably nice).
Basically I'm skeptical about version of the pre-commit hook having to follow version of the package itself. There could be two separate tag prefixes for example, like I do in DI.

.github/workflows/PreCommitHooks.yml
```diff
@@ -44,3 +44,75 @@ jobs:
...
+          persist-credentials: false
+
+      - uses: julia-actions/setup-julia@v3
+        with:
+          version: '1'
+
+      - name: Install jlfmt app
+        # Need to explicitly add registry because of
+        # https://github.com/JuliaLang/Pkg.jl/issues/4360 which is not
+        # backported to 1.12 yet
```
gdalle, line 75: Should we ask for a backport on the issue?

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "JuliaFormatter"
 uuid = "98e50ef6-434e-11e9-1051-2b60c6c9e899"
-version = "2.3.2"
+version = "2.4.0"
```
gdalle, line 3: See my comment about package versioning vs hook versioning. Here I think the bump might be justified because of a docs change, but in general it depends whether you consider the hook to be "part of" the Julia package, which is debatable. For example, it won't be installed when people do `Pkg.add("JuliaFormatter")`.

.github/workflows/PreCommitHooks.yml
```diff
@@ -44,3 +44,75 @@ jobs:
...
+          pipx run pre-commit try-repo ../../../ jlfmt --files main.jl --verbose
+
+          # Restore PATH for future steps
+          export PATH=$OLDPATH
+
+      - name: Test pre-commit hook with --jlfmt-path argument
+        working-directory: .github/workflows/pre-commit
+        run: |
+          # Reset the test file
+          git checkout main.jl
```
gdalle, line 103: Maybe add a comment specifying that `main.jl` should remain ill-formatted according to JuliaFormatter itself, for this check to work?

bin/jlfmt-hook.sh
```diff
@@ -0,0 +1,32 @@
...
+    case "$arg" in
+        --jlfmt-path=*) jlfmt="${arg#--jlfmt-path=}" ;;
+        *) args+=("$arg") ;;
+    esac
+done
+
+jlfmt="${jlfmt/#\~/$HOME}"
+
+if ! command -v "$jlfmt" &>/dev/null; then
+    echo "ERROR: '$jlfmt' not found."
```
gdalle: The friendly error message is a nice touch!

docs/src/cli.md
```diff
@@ -1,18 +1,24 @@
-# Command Line Interface
+# [Command Line Interface](@id cli)
```
gdalle, line 1: Stellar docs work as always :)

docs/src/integrations.md
```diff
@@ -2,22 +2,174 @@
...
-# ... other repos you may have
 ```
 
-You can find a list of releases [here](https://github.com/JuliaEditorSupport/JuliaFormatter.jl/releases).
-**Be sure to use the entire version string!** (You can double-check this by opening the
-release and looking at the part of the URL that follows `.../releases/tag/VERSION`.)
+(If you have other pre-commit hooks, just add the `repo: ...` block to your pre-existing list of repos.)
+
+**Note that `rev` controls the version of the _hook_ that is checked out; it does not control which version of JuliaFormatter is actually used to format your code.**
+The version used to do the actual formatting is determined by the version of JuliaFormatter that is installed in your global Julia environment.
```
gdalle, line 43: Not really in your global Julia _environment_, it's not in `@v1.12` since it is an app.

gdalle:
In case you want to point people to the required steps for implementing pre-commit on their repos: https://github.com/JuliaDiff/SparseMatrixColorings.jl/pull/320

.github/workflows/PreCommitHooks.yml
penelopeysm: Yes, I had actually meant to investigate this in more depth because it seems like it *has* been backported: https://github.com/JuliaLang/Pkg.jl/pull/4493 I'm not sure why the CI runner doesn't pick it up.

Project.toml
penelopeysm: Right, thanks for the suggestions! There are indeed a few good options to decouple semver of the package from semver of the hook.

But also, I don't really envision that there would be much need to update the hooks much in the future, so my inclination right now is that it's not hugely important to sort it out, and we can continue to live in a regime where hook behaviour is in practice guaranteed by package semver but not officially documented. I think if there is a need to make a breaking change in the future then we can revisit this design?

docs/src/integrations.md
penelopeysm: Good catch! I'll rephrase.

Project.toml
penelopeysm: (I think if I were to change this design, I would probably use the tag prefixes, that seems like the most elegant way to solve it. I'd also be interested in finding out how that interacts with things like Dependabot, but that'd be a minor detail.)

docs/src/integrations.md
penelopeysm: Oh, hold on, this part is actually right because it's the old hook that uses `julia -e 'using JuliaFormatter...'`. That does indeed use the global env. We should really allow passing a project argument though.

penelopeysm: > We should really allow passing a project argument though.

Actually this is already possible via pre-commit `args`, so I'll just document it and call it a day

docs/src/integrations.md
```diff
@@ -2,22 +2,193 @@
...
+
+To use the hook, you can add the following to your `.pre-commit-config.yaml`:
 
 ```yaml
 repos:
-# ... other repos you may have
-- repo: "https://github.com/JuliaEditorSupport/JuliaFormatter.jl"
-  rev: "v1.0.18"  # or whatever the desired release is
+- repo: https://github.com/JuliaEditorSupport/JuliaFormatter.jl
+  rev: baf84dcde3e7d39a3339fecb51a5d853f8aa35af
```
gdalle, line 24: It seems a bit dangerous to specify the commit hash before the PR's merge commit?

gdalle (changes requested):

docs/src/integrations.md
penelopeysm: This is the current release (https://github.com/JuliaEditorSupport/JuliaFormatter.jl/commit/baf84dcde3e7d39a3339fecb51a5d853f8aa35af), so it is fine, I have not reverse engineered SHA :-)

PR #994, opened 2026-05-26
Fix docs warnings, add docs preview comment in CI

### LINKED ISSUE #993: Fix docs warnings and add a docs preview workflow

penelopeysm:
Closes #993


.github/workflows/docs.yml
```diff
@@ -0,0 +1,66 @@
...
+      - name: Deploy docs to gh-pages branch
+        working-directory: docs
+        shell: julia --color=yes --project=. {0}
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
+        # Must pass `root` when `deploydocs()` is run from outside make.jl file
+        # Also, `root` must be an absolute path (hence the call to `pwd()`)
+        run: |
+          using Documenter
+          deploydocs(; root=pwd(), repo="github.com/JuliaEditorSupport/JuliaFormatter.jl.git", push_preview=true)
```
penelopeysm: Hmmmm, I stole this from FlexiChains which in turn stole it from Turing, but I realise this is probably overcomplicated. In Turing we had to inject the header HTML after makedocs but before deploydocs. This isn't the case here, so we shouldn't need this.

PR #933, opened 2025-07-31
Further improve SciMLStyle line breaking for mathematical expressions

ChrisRackauckas:
## Summary

This PR builds on #931 to address additional line breaking issues in SciMLStyle, specifically for mathematical expressions as observed in SciML/NonlinearSolve.jl#660.

## Changes

1. **Enhanced `find_optimal_nest_placeholders` in `nest_utils.jl`**:
   - Added handling for Binary/Chain/Comparison nodes (mathematical expressions) 
   - Added handling for Conditional nodes (ternary operators)
   - Optimized by calculating `total_length` once at the beginning

2. **Added extra protection in `sciml/nest.jl`**:
   - Added a second layer of conservative nesting rules at the nesting decision point
   - This catches cases that might slip through the placeholder optimization

## Why these changes are needed

While #931 fixed many issues, it didn't specifically address mathematical expressions like:
```julia
du[i, j, 1] = alpha * (u[im1, j, 1] + u[ip1, j, 1] + u[i, jp1, 1] + u[i, jm1, 1] - 4u[i, j, 1]) + B + u[i, j, 1]^2 * u[i, j, 2] - (A + 1) * u[i, j, 1] + brusselator_f(x, y)
```

These expressions were still being broken unnecessarily. This PR ensures that:
- Binary operations with ≤3 operators stay on one line if within margin + 20
- Binary operations with ≤6 operators stay on one line if within margin
- Ternary operators are preserved when reasonable

## Testing

- All 1551 existing tests pass
- The new tests from my initial attempt were already added in #931
- Tested specifically against the NonlinearSolve.jl#660 examples

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #931, opened 2025-07-29
Fix excessive line breaking in SciMLStyle for readable expressions

ChrisRackauckas:
## Summary

Fixes the formatting regression introduced in JuliaFormatter v2 where commonly-used expressions were being broken across multiple lines, reducing code readability in the SciML ecosystem.

Fixes https://github.com/domluna/JuliaFormatter.jl/issues/930

## Problem

JuliaFormatter v2's rewrite with JuliaSyntax introduced overly aggressive line breaking through the `find_optimal_nest_placeholders` function. This caused several problematic patterns as shown in https://github.com/SciML/DiffEqGPU.jl/pull/356/files:

**Before (problematic):**
```julia
du[II[i,
    j,
    1]] = α * (...)

@unpack γ, a31, a32, a41, a42, a43, btilde1, btilde2, btilde3, btilde4, c3, α31,
α32 = integ.tab

beta1, beta2, qmax, qmin, gamma, qoldinit, _ = build_adaptive_controller_cache(
beta1, beta2, qmax, qmin, gamma, qoldinit,
_ = build_adaptive_controller_cache(
integ.alg,
T)
```

**After (fixed):**
```julia
du[II[i, j, 1]] = α * (...)

@unpack γ, a31, a32, a41, a42, a43, btilde1, btilde2, btilde3, btilde4, c3, α31, α32 = integ.tab

beta1, beta2, qmax, qmin, gamma, qoldinit, _ = build_adaptive_controller_cache(integ.alg, T)
```

## Solution

Enhanced `find_optimal_nest_placeholders()` in `src/nest_utils.jl` to be more conservative for specific expression types that benefit from staying compact:

### Fixed Expression Types:
- **Array indexing (`RefN`)**: `II[i, j, 1]` no longer broken across lines
- **Macro calls (`MacroCall`)**: `@unpack` statements keep reasonable line breaks  
- **Function calls (`Call`)**: Avoid awkward breaks in function calls
- **Type parameters (`Curly`)**: Keep `Type{A, B, C}` compact
- **Vector literals (`Vect`)**: Short vectors stay on one line when possible

### Implementation Details:
- Each expression type gets appropriate thresholds for placeholder count and margin tolerance
- Graduated approach: some types (MacroCalls) get more margin tolerance than others
- Still respects margin limits but gives more leeway for readable expressions
- Centralized logic in one location for maintainability

## Testing

Added comprehensive test suite in `test/sciml_style.jl` covering all fixed patterns to prevent regressions.

## Test plan

[...]

PR #997, opened 2026-05-28
Fix SciML assignment ref nesting

svchb:
Before:
```julia
function update_cache!()
    cache_table[key,
                slot] = source_table[key, slot] + step * delta_table[key, slot]
end
```

After:
```julia
function update_cache!()
    cache_table[key, slot] = source_table[key, slot] +
                             step * delta_table[key, slot]
end
```

### DIFF
--- a/src/styles/sciml/nest.jl
+++ b/src/styles/sciml/nest.jl
@@ -25,12 +25,15 @@ for f in [
     end
 end
 
-function _nearest_binary_is_assignment(
+function _is_lhs_of_assignment(
+    s::State,
     lineage::Vector{Tuple{FNode,Union{Nothing,Metadata}}},
 )
+    # If it's not a lhs of a binary, it can't be a lhs of an assignment
+    s.is_lhs_of_binary || return false
+    # Check lineage to see what the binary operator was (if any)
     i = findlast(x -> x[1] === Binary, lineage)
     i === nothing && return false
-
     metadata = lineage[i][2]
     return !isnothing(metadata) && metadata.is_assignment
 end
@@ -94,17 +97,8 @@ function _n_tuple!(
     lineage::Vector{Tuple{FNode,Union{Nothing,Metadata}}},
 )
     style = getstyle(ss)
-    
-    # TODO(penelopeysm): In principle, this should only check whether we are on the LHS of
-    # an assignment. However, this function checks whether we are an argument to an
-    # assignment. For example, if we have:
-    #
-    #     a[x] = b[y]
-    #
-    # then this would trigger for both operands, even though we probably only want to
-    # trigger it for `a[x]`. In practice, this doesn't seem to be a problem, so doesn't need
-    # to be fixed immediately.
-    line_margin = if _nearest_binary_is_assignment(lineage)
+
+    line_margin = if _is_lhs_of_assignment(s, lineage)
         # If on the LHS of an assignment, don't take the ` = RHS` into account when
         # considering whether to nest, as that would lead to overly eager nesting of the
         # LHS.
@@ -163,12 +157,6 @@ function _n_tuple!(
 
     # Check if we should apply conservative nesting rules
     should_nest = line_margin > s.opts.margin || must_nest(fst) || src_diff_line
-
-    println("---------")
-    @info _nearest_binary_is_assignment(lineage)
-    @info first(fst.nodes).val
-    @info line_margin s.opts.margin must_nest(fst) src_diff_line should_nest
-    println("---------")
 
     # For certain types, be more conservative about nesting
     if should_nest && !must_nest(fst) && !src_diff_line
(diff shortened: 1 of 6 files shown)


penelopeysm:
@svchb First of all, thanks so much for splitting up the PRs!

For this one, could I check with you specifically what is the regression relative to? I was looking at the git blame, and this part of the code was last modified in #931 because JuliaFormatter used to _actively_ insert line breaks inside the LHS ref.

But if the LHS ref was already broken up prior to formatting (as is the case in the test this PR adds), I haven't found any version of JuliaFormatter that will move it back on to the same line. This is running with v2.1.5 (which corresponds to right after #931):

```julia
julia> using JuliaFormatter

julia> s = """function update_cache!()
           cache_table[key,
                       slot] = source_table[key, slot] + step * delta_table[key, slot]
       end"""
"function update_cache!()\n    cache_table[key,\n                slot] = source_table[key, slot] + step * delta_table[key, slot]\nend"

julia> s |> println
function update_cache!()
    cache_table[key,
                slot] = source_table[key, slot] + step * delta_table[key, slot]
end

julia> format_text(s, SciMLStyle()) |> println
function update_cache!()
    cache_table[key,
    slot] = source_table[key, slot] + step * delta_table[key, slot]
end
```

Now, IMO, you could definitely argue that the output above is really ugly, and this PR introduces better behaviour by re-formatting the entire expression from the start. I'm totally on board with that, just wanted to make sure I wasn't misunderstanding something.

svchb:
Previous JuliaFormatter versions introduced alot of these ugly line breaks. So the intention here is to revert these ugly line breaks. So this is not really a regression, but in line with the intentions of the SciML style.

svchb:
@penelopeysm All the PRs are just basically there to fix the SciML style applied via JuliaFormatter on the trixi ecosystem.

penelopeysm:
> Previous JuliaFormatter versions introduced alot of these ugly line breaks. So the intention here is to revert these ugly line breaks. So this is not really a regression, but in line with the intentions of the SciML style.

Brilliant, thanks! That is what I was thinking, I just wanted to double check. Great, let me review this and get this merged.

penelopeysm:
(For the record and for future self, `DefaultStyle` will re-format this nicely:

```julia
julia> format_text(s; margin=80) |> println
function update_cache!()
    cache_table[key, slot] =
        source_table[key, slot] + step * delta_table[key, slot]
end
```

so I think it's consistent for SciMLStyle to also reformat it back into a prettier option. Also, the current way it gets formatted (with the indent on the third line) is really ugly and probably wrong by any sensible style guide.)

src/styles/sciml/nest.jl
penelopeysm: This code (introduced in #933) is a rather bizarre shortcircuit imo. #931 already introduced rules inside `_n_tuple!` and `find_optimal_nest_placeholders` to avoid breaking ref expressions up. So it's very odd that we should do an early return here and not make use of that code in `_n_tuple!`.

I've looked into this a fair amount, and I think there is a much better way of achieving the same thing, which at the same time will hopefully reduce some of the technical debt from the LLM PRs

penelopeysm:
OK. Having thought about this quite a bit, I've concluded that I **can't** actually accept the formatting change in this PR (sorry 😞). The reason for this is because SciMLStyle sets the option `join_lines_based_on_source = true` by default:

- https://juliaeditorsupport.github.io/JuliaFormatter.jl/stable/sciml_style/
- https://juliaeditorsupport.github.io/JuliaFormatter.jl/stable/#join_lines_based_on_source

which is really quite an explicit command that says to _not_ remove newlines that are already present in code. Making JuliaFormatter reformat the text to remove the newline is directly contravening this. Besides, it means that it's impossible to distinguish between a case where the newline was inserted spuriously by an old version of JuliaFormatter, and a case where the user actually wanted a newline. I'm sorry that the old version messed things up, but that's not in my power to fix, and I don't want to do two wrongs to make a right.

You can, however, disable this option if you want to remove that newline. In particular, the last of these will reformat to the intended output:

```julia
julia> s = """function update_cache!()
                  cache_table[key,
                              slot] = source_table[key, slot] + step * delta_table[key, slot]
              end"""
"function update_cache!()\n    cache_table[key,\n                slot] = source_table[key, slot] + step * delta_table[key, slot]\nend"

julia> s |> println
function update_cache!()
    cache_table[key,
                slot] = source_table[key, slot] + step * delta_table[key, slot]
end

julia> format_text(s, SciMLStyle()) |> println
function update_cache!()
    cache_table[key,
        slot] = source_table[key, slot] + step * delta_table[key, slot]
end

julia> format_text(s, SciMLStyle(); join_lines_based_on_source=false) |> println
function update_cache!()
    cache_table[key, slot] = source_table[key, slot] + step * delta_table[key, slot]
end

julia> s2 = format_text(s, SciMLStyle(); join_lines_based_on_source=false, margin=80); s2 |> println
function update_cache!()
    cache_table[key, slot] = source_table[key, slot] +
                             step * delta_table[key, slot]
end
```

and, once the linebreak has been removed, formatting it again will not re-introduce the linebreaks:

```julia
julia> format_text(s2, SciMLStyle(); margin=80) |> println
function update_cache!()
    cache_table[key, slot] = source_table[key, slot] +
                             step * delta_table[key, slot]
end
```

PR #998, opened 2026-05-28
reduce SciML indentation on lhs of ops, make soft margin overrun configurable

svchb:
## Summary

- Add a configurable `margin_overrun` option, defaulting to `20`, to make the existing soft-margin tolerance explicit.
- Replace the hard-coded soft-margin tolerances (`+10`, `+20`, `+30`, `+60`) with factors derived from `margin_overrun`.
- Tighten the SciML `for (...) in ...` tuple-binding special case so it considers the full first-line segment, not only the tuple binding.

## Behavior

The default preserves the current soft-margin behavior. Users can set `margin_overrun = 0` to make these soft-margin checks strict.


src/styles/sciml/nest.jl
```diff
@@ -44,6 +44,39 @@ end
...
+function _is_for_tuple_binding(
+    fst::FST,
+    s::State,
+    lineage::Vector{Tuple{FNode,Union{Nothing,Metadata}}},
+)
+    length(lineage) >= 2 &&
+        lineage[end-1][1] === CartesianIterator &&
+        op_kind(fst) in KSet"in ∈" &&
+        fst[1].typ === TupleN &&
+        s.line_offset + length(fst[1]) <= s.opts.margin
```
penelopeysm: Is this perhaps a bit too lenient? After the `for (...)`, there's still on top of that an `in (...)` which has to fit on the same line. Basically here you're saying that, as long as the `for (...)` fits on this line, put it on this line.

This works nicely when we're well under the margin, because the formatter can then choose to nest the `in (...)` section instead, like in the example you demonstrated.

However, if we're close to the margin this can result in things going over margin quite drastically:

```julia
str = """
@threaded destination.scheduler for (dest_id, src_id, aaaa, bbbb, ccccc, dddd) in zip(eachindex(destination),
          eachindex(source))
    @inbounds destination[dest_id] = source[src_id]
end
"""
using JuliaFormatter
format_text(str, SciMLStyle(); yas_style_nesting=true, margin=92) |> println
```

In this example, `ddddd)` goes up to column 78, which is well under the margin of 92.

Disabling nesting causes the line to go up to column 109:

```julia
julia> format_text(str, SciMLStyle(); yas_style_nesting=true, margin=92) |> println
@threaded destination.scheduler for (dest_id, src_id, aaaa, bbbb, ccccc, dddd) in zip(eachindex(destination),
                                                                                      eachindex(source))
    @inbounds destination[dest_id] = source[src_id]
end
```

whereas on the current release the nesting is indeed a bit ugly, but at least it respects the margins:

```julia
julia> format_text(str, SciMLStyle(); yas_style_nesting=true, margin=92) |> println
@threaded destination.scheduler for (dest_id, src_id, aaaa, bbbb, ccccc,
                                     dddd) in zip(eachindex(destination),
                                                  eachindex(source))
    @inbounds destination[dest_id] = source[src_id]
end
```

svchb: I updated the code to make the current hardcoded soft margins configurable and respect the current default value of 20 for soft margins.

penelopeysm:
Thanks!

I know you didn't add these, but are any of these rules actually in [the SciML style guide](https://docs.sciml.ai/SciMLStyle/stable/)?

svchb:
> Thanks!
> 
> I know you didn't add these, but are any of these rules actually in [the SciML style guide](https://docs.sciml.ai/SciMLStyle/stable/)?

No the guide says something like judgment should be used. Anyway I think making it configurable already makes it better since it allows even for strict margins with sciml style. For example, we basically want soft margins of 90 to 120.

src/styles/sciml/nest.jl
```diff
@@ -25,6 +25,41 @@ for f in [
...
+)
+    if _is_for_tuple_binding(fst, s, lineage)
+        lhs = fst[1]
+        nest_behavior = lhs.nest_behavior
+        lhs.nest_behavior = NeverNest
+        try
+            return n_binaryopcall!(DefaultStyle(getstyle(ss)), fst, s, lineage; indent)
+        finally
+            lhs.nest_behavior = nest_behavior
+        end
```
penelopeysm: I don't think there are any cases where we actually want to catch an error from `n_binaryopcall!` so can we skip the try/finally as it's bad for performance.

src/styles/sciml/nest.jl
```diff
@@ -25,6 +25,41 @@ for f in [
...
+    fst::FST,
+    s::State,
+    lineage::Vector{Tuple{FNode,Union{Nothing,Metadata}}};
+    indent::Int = -1,
+)
+    if _is_for_tuple_binding(fst, s, lineage)
+        lhs = fst[1]
+        nest_behavior = lhs.nest_behavior
+        lhs.nest_behavior = NeverNest
+        try
```
penelopeysm: I don't know if this is possible (I can look into this later as well), but I would really prefer if this logic could somehow be shifted into `find_optimal_nest_placeholders` because conceptually all the special case nesting logic should be self-contained there. Overloading `n_binaryopcall!` feels like using a large hammer for a small nail.

svchb: done and i added more tests for all the config options

test/sciml_style.jl
```diff
@@ -640,6 +640,119 @@
...
+            @threaded destination.scheduler for (dest_id, src_id, aaaa, bbbb, ccccc, dddd) in zip(eachindex(destination),
+                                                                                                  eachindex(source))
+                @inbounds destination[dest_id] = source[src_id]
+            end
+            """
+            @test format_text(str, SciMLStyle(); yas_style_nesting = true, margin = 92) == fstr
+
+            fstr = """
+            @threaded destination.scheduler for (dest_id, src_id, aaaa, bbbb, ccccc,
+                                                dddd) in zip(eachindex(destination),
```
penelopeysm: I think dddd should be indented one space more?

src/nest_utils.jl
```diff
@@ -257,43 +262,149 @@ end
...
+    isempty(newline_inds) && return false
+
+    r = linerange(s, fst.startline)
+    line = s.doc.srcfile.code[first(r):last(r)]
+    call_offset = fst[1].line_offset
+    indent = nothing
+
+    # Nest lineage does not carry sibling nodes, so the source token `for (`
+    # is the narrow tuple-binding guard for this RHS call.
+    for m in eachmatch(r"\bfor\s+\(", line)
```
penelopeysm: `for` and `(` don't have to be on the same line, though. And honestly even if this heuristic worked all the time, I don't really like these checks since they seem fragile. The comment here is exactly right, it's because lineage doesn't give you enough info to do so.

Now, imo fst/lineage structs need to be reworked a bit so that some of these predicates can be more robustly checked for. I recognise that this will take a bit of time, but in the meantime, I'm afraid I'm very hesitant to introduce more code like this that does almost the right thing but not quite. It's just going to cause more headaches down the line because it's more code to refactor and more backwards compatibility cases to consider.

Finally note that the FST transformation that this function implements directly contravenes `join_lines_based_on_source` since it removes newlines that are present in the source. At the minimum it needs to be disabled when that option is true, but I would go one step further and say that it should just be removed. After all the `join_lines_based_on_source` function already does exactly the same thing - on v2.5.6 the newline between `eachindex(destination)` and `eachindex(source)` is already removed. For example:

(the for-tuple is still nested, but that's orthogonal to what this function tries to do.)

```julia
julia> str = """
       @threaded destination.scheduler for (dest_id, src_id, aaaa, bbbb, ccccc, dddd) in zip(eachindex(destination),
                 eachindex(source))
           @inbounds destination[dest_id] = source[src_id]
       end
       """
"@threaded destination.scheduler for (dest_id, src_id, aaaa, bbbb, ccccc, dddd) in zip(eachindex(destination),\n          eachindex(source))\n    @inbounds destination[dest_id] = source[src_id]\nend\n"

julia> str |> println
@threaded destination.scheduler for (dest_id, src_id, aaaa, bbbb, ccccc, dddd) in zip(eachindex(destination),
          eachindex(source))
    @inbounds destination[dest_id] = source[src_id]
end


julia> ft(str, SciMLStyle(); join_lines_based_on_source=false) |> println
@threaded destination.scheduler for (dest_id, src_id, aaaa, bbbb, ccccc,
    dddd) in zip(eachindex(destination), eachindex(source))
    @inbounds destination[dest_id] = source[src_id]
end
```

src/styles/default/nest.jl
```diff
@@ -988,8 +988,8 @@ function n_binaryopcall!(
         return nested
     end
 
-    # length of operator, surrounding whitespace, and right operand
-    op_and_rhs_len = sum(length.(fst[2:end]))
+    # length of operator, surrounding whitespace, and the first line of the right operand
+    op_and_rhs_len, _ = length_to(fst, (NEWLINE,); start = 2)
```
penelopeysm, line 992: #1012 does similar code here and is also more general, so I think should also help with some of the cases here.

penelopeysm (changes requested):
Thanks, I'm happy with most of it sans the function for rejoining lines, which I don't really think is needed.

penelopeysm:
> > I know you didn't add these, but are any of these rules actually in [the SciML style guide](https://docs.sciml.ai/SciMLStyle/stable/)?
>
> No the guide says something like judgment should be used. Anyway I think making it configurable already makes it better since it allows even for strict margins with sciml style. For example, we basically want soft margins of 90 to 120.

General point regarding these extra nesting rules: These seem to have been added in #931 or #933. I'm not saying that I want to remove them from JuliaFormatter, but I would **much** prefer to be in a situation where JuliaFormatter implements a spec that is written up in the style guide, rather than having JuliaFormatter's code be the *de facto* implementation of "SciML style". (cc @ChrisRackauckas) Would it be possible to at least add something to the style guide that lets me justify having these extra lines of code in here?

penelopeysm:
Though I just discovered sciml is now just runic https://github.com/SciML/SciMLStyle#runic so I'm not really sure about the extent to which I actually need to maintain this part of the codebase? Some clarification would be really helpful.

I don't begrudge anyone for switching to Runic (some of my own projects use it) but I'm acutely conscious that fixing JuliaFormatter is an uphill task and even if I pour all my free time into it (which I'd rather not do), it's still going to take a long time.

test/sciml_style.jl
svchb: fixed

svchb:
@penelopeysm trixi.jl also switched to runic. There was a large PR by chris that fixed a lot of the issues, but it never got merged. 
Anyway if all the issues are fixed trixi.jl might switch back. Trixi also uses sciml style so this is not limited to the sciml projects.

JoshuaLampert:
> trixi.jl also switched to runic.

To clarify: Trixi.jl uses JuliaFormatter.jl with the SciML style (still v1 because of the issues with v2 @svchb tries to solve in the recent series of PRs (thanks a lot for that!)), but we tried Runic in this PR https://github.com/trixi-framework/Trixi.jl/pull/2070, but were not convinced by Runic.jl.

svchb:
@JoshuaLampert Thanks for the clarification. We tried runic for trixi-particles as well and I understood Erik as if you would switch.

src/nest_utils.jl
svchb: I simplified the code one of the format string tests is a todo item now.

penelopeysm:
@svchb @JoshuaLampert Thanks for the explanation! I think my general point still stands: I'm not really sure what style guide we're adhering to here and that makes it really hard for me to justify having this code. If "SciML style" === Runic, and Trixi is not using Runic, then Trixi is not using "SciML style". Again, I'm not proposing to just completely drop the code from JuliaFormatter (I'm not that radical 😉), but I do think it has to stand on firmer ground.

I think there are several good ways forward, for example:

- Fork the (pre-Runic) SciML style guide to trixi and call the style 'trixi' instead;

- I could figure out a way to make JuliaFormatter more extensible (i'm not sure that would work with the CLI app, but it would definitely be possible using JuliaFormatter as a library), and then you could own an in-house 'trixi style' which you could tweak to your liking

All of these wouldn't be in v2 of JuliaFormatter though -- more likely v3.

In the meantime in JuliaFormatter I should probably make it very clear that sciml style is not actually sciml style but rather some legacy version with a smattering of custom rules -- then at least I don't have to claim that it lines up with the sciml style guide.

penelopeysm (commented):
On the PR itself: Thanks for doing this, I really appreciate it! I'll make some tweaks and also merge it into #1012 just to make sure that there aren't any obvious bad interactions.

I'll check the output of this PR against a couple of trixi repos and if it doesn't blow up, I'll be happy to approve.

penelopeysm:
!formatbot trixi-framework/Trixi.jl use_config=true

JoshuaLampert:
> "SciML style" === Runic

SciML style is Runic, but `SciMLStyle` is not Runic 😉. SciML used to use the `SciMLStyle` defined in JuliaFormatter.jl and other projects like the Trix.jl ecosystem (but I guess there are others, too) adopted it. But at some point SciML switched their formatting style from the JuliaFormatter.jl `SciMLStyle` to Runic. Other projects sticked with the `SciMLStyle`. In the end `SciMLStyle` is just a name for some set of style choices that some people currently prefer. It probably makes sense to rename the style in a breaking release if SciML does not use this style (anymore).

penelopeysm:
!formatbot trixi-framework/TrixiParticles.jl use_config=true

penelopeysm:
> It probably makes sense to rename the style in a breaking release if SciML does not use this style (anymore).

Yes, exactly 😄

penelopeysm (approved):

PR #999, opened 2026-05-28
Align SciML Dict tuple pairs

svchb:
Before:
```julia
algorithm_cases = Dict(
    "default search" => (),
    "with static scheduler" => (backend=StaticBackend(),),
    "with spline kernel" => (smoothing_length=1.1 *
                                              item_spacing,
                                smoothing_kernel=SplineKernel{2}())
)
```

After:
```julia
algorithm_cases = Dict(
    "default search"        => (),
    "with static scheduler" => (backend=StaticBackend(),),
    "with spline kernel"    => (smoothing_length=1.1 * item_spacing,
                                smoothing_kernel=SplineKernel{2}())
)
```

### DIFF
--- a/src/styles/sciml/nest.jl
+++ b/src/styles/sciml/nest.jl
@@ -44,6 +44,162 @@ end
 #     nest!(style, fst[end], s, lineage)
 # end
 
+function _align_tuple_comments!(fst::FST)
+    for n in fst.nodes::Vector
+        n.typ === NOTCODE && (n.indent = fst.indent)
+    end
+end
+
+function _is_dict_call(fst::FST)
+    fst.typ === Call &&
+        length(fst.nodes::Vector) > 0 &&
+        fst[1].typ === IDENTIFIER &&
+        fst[1].val == "Dict"
+end
+
+function _is_tuple_pair(fst::FST)
+    fst.typ === Binary && op_kind(fst) === K"=>" && fst[end].typ === TupleN
+end
+
+function _align_dict_tuple_pair_arrows!(fst::FST)
+    pair_inds = findall(_is_tuple_pair, fst.nodes::Vector)
+    length(pair_inds) > 1 || return
+
+    align_len = maximum(node_align_length(fst[i][1]) for i in pair_inds)
+    for i in pair_inds
+        pair = fst[i]
+        ws = align_len - node_align_length(pair[1]) + 1
+        align_binaryopcall!(pair, ws)
+    end
+end
+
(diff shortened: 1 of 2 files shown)


penelopeysm:
This is another slightly nuanced one...

(1) It shouldn't join the `1.1 * item_spacing`, unless `join_lines_based_on_source = false`. I think this is similar to previous examples we've discussed.

(2) I don't think there's any style where pair arrows are _forced_ to align. There is the `align_pair_arrow` option, which you can set to true **but** this only aligns the pair arrows when there's pre-existing alignment. That is, if `align_pair_arrow=true`, 

```julia
a => b
cccc => dddd
```

**won't** be converted to 

```julia
a    => b
cccc => dddd
```

. However, if it's already in that latter format, then `align_pair_arrow` is supposed to make sure it doesn't revert to the former.

This does actually behave correctly (kind of) -- I've tweaked this input so that `1.1 * item_spacing` is on the same line, and added a few more spaces to the `default search` line to enable the custom alignment heuristic

```julia
julia> s = """
       algorithm_cases = Dict(
           "default search"        => (),
           "with static scheduler" => (backend=StaticBackend(),),
           "with spline kernel" => (smoothing_length=1.1 * item_spacing,
                                       smoothing_kernel=SplineKernel{2}())
       )
       """
"algorithm_cases = Dict(\n    \"default search\"        => (),\n    \"with static scheduler\" => (backend=StaticBackend(),),\n    \"with spline kernel\" => (smoothing_length=1.1 * item_spacing,\n                                smoothing_kernel=SplineKernel{2}())\n)\n"

julia> s |> println
algorithm_cases = Dict(
    "default search"        => (),
    "with static scheduler" => (backend=StaticBackend(),),
    "with spline kernel" => (smoothing_length=1.1 * item_spacing,
                                smoothing_kernel=SplineKernel{2}())
)

julia> format_text(s, SciMLStyle(); align_pair_arrow=true) |> println
algorithm_cases = Dict(
    "default search"        => (),
    "with static scheduler" => (backend = StaticBackend(),),
    "with spline kernel"    => (smoothing_length = 1.1 * item_spacing,
    smoothing_kernel = SplineKernel{2}())
)
```

So the alignment is behaving correctly, although the indentation of `smoothing_kernel` is probably not right -- it should be indented one more level.

All in all, I think there are some bugs here, but I'm not sure they're the same bugs that the PR tries to fix.

penelopeysm:
There could also be a flag that says "yes really force the alignment even if it's not already aligned to begin with" -- that'd be a separate config option though that would apply across all styles.

PR #1008, opened 2026-05-30
Add FormatBot workflow

### LINKED ISSUE #1007: github bot or similar to test against repositories
It would be very nice if I could set up a bot to:

- clone one or more named repository
- run the formatter on main with a given set of options
- run the formatter on a PR with the same set of options
- comment the diff on the PR

It would probably be worth setting this up for selected repositories (julialang/julia?) on CI as well

penelopeysm:
Closes #1007

Clauded this, but it took a LOT of manual cleanup even for a simple Python script :/.

Anyway, this workflow should trigger whenever someone comments on a PR with the following invocation:

```
!formatbot owner/repo@rev style=DefaultStyle() indent=2 margin=80
```

where `@rev` is optional, and all the kwargs are optional (and should just be parsed exactly as Julia expects it). Style is specified as a kwarg as well just for ease of parsing, even though in `format()` it's a positional arg.

Since this allows for arbitrary code execution, it's actually kind of unsafe. For example, one could do

```
!formatbot owner/repo@rev style=do_something_dangerous()
```

I need to think about how to mitigate this possibility.

In general anybody who can run a github actions workflow can perform this kind of attack without needing to inject code into a trigger invocation (just modify the workflow to run said code). That means that (for example) if somebody has merge rights to the repo, they can do this sort of thing anyway, so we don't need to protect against anything more than that. The main point would be to avoid allowing untrusted users to run code.

For now, the workflow only runs if it's invoked by me.

### DIFF
--- a/.github/workflows/FormatBot.yml
+++ b/.github/workflows/FormatBot.yml
@@ -7,6 +7,7 @@ on:
 jobs:
   format:
     name: FormatBot
+    # TODO: Let other people trigger the bot as well. It's restricted for now so that I can test.
     if: >-
       github.event.issue.pull_request
       && github.actor == 'penelopeysm'
--- a/.github/workflows/FormatBot.yml
+++ b/.github/workflows/FormatBot.yml
@@ -1,0 +1,239 @@
+name: FormatBot
+
+on:
+  issue_comment:
+    types: [created]
+
+jobs:
+  format:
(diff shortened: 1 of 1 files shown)


penelopeysm:
Because of the way GHA works, this has to be merged to the default branch before I can test it.

PR #1012, opened 2026-05-30
Avoid nesting LHS if not over margin and if RHS can be nested

penelopeysm:
This fixes a similar issue as #998, but more generally. It essentially does what the comments say. If we have a binary expression `LHS op RHS`, we try to avoid nesting the LHS if there is any point in the `op RHS` segment where we can nest it.

<details><summary>More discussion (outdated as the regression was fixed)</summary>

However, it also causes some regressions. For example this causes

```julia
julia> str = "foo(a, b, c)::Rtype where {A,B} = 10"
"foo(a, b, c)::Rtype where {A,B} = 10"

julia> JuliaFormatter.format_text(str; indent=4, margin=32) |> println
foo(a, b, c)::Rtype where {
    A,
    B,
} = 10
```

when previously it would be formatted instead to

```julia
foo(
    a,
    b,
    c,
)::Rtype where {A,B} = 10
```

I still think that the solution in this PR is the correct way to deal with the nesting preference, but it needs to be more tightly scoped to avoid causing regressions in other cases that we aren't trying to improve.

</summary>

### DIFF
--- a/HISTORY_v2.md
+++ b/HISTORY_v2.md
@@ -1,3 +1,14 @@
+# v2.5.7
+
+Fixed a number of cases where the left-hand operand of binary operators would be aggressively nested.
+(In general, it is better to nest the right-hand operand as that keeps as much of the operation as possible on the same line.)
+These fixes were applied to DefaultStyle and thus should propagate to other styles as well.
+However, SciMLStyle goes slightly further than this and also allows expressions to extend beyond the stated margin in the interests of not nesting the LHS. (#998, #1012)
+
+Added a new formatting option, `sciml_margin_overrun`, which controls the extent to which SciMLStyle allows expressions to extend beyond the margin in order to avoid nesting the LHS of binary operators.
+The default is 20, but it can be set to 0 to prevent this behaviour if undesired.
+This option has no effect for styles apart from SciML. (#998)
+
 # v2.5.6
 
 Fixed a bug where JuliaFormatter would emit invalid code when parsing an `if`/`elseif` with a block condition (e.g. `(a; b)` or `begin ... end`). (#1025, #1026)
--- a/HISTORY_v2.md
+++ b/HISTORY_v2.md
@@ -1,4 +1,4 @@
-# v2.5.7
+# v2.6.0
 
 Fixed a number of cases where the left-hand operand of binary operators would be aggressively nested.
 (In general, it is better to nest the right-hand operand as that keeps as much of the operation as possible on the same line.)
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,1 @@
-Manifest.toml
+Manifest*.toml
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -8,6 +8,11 @@
 
 - Code changes in the local checkout should be immediately visible without having to reload the session (thanks to Revise.jl).
 
+# Running tests
+
+- Do not run the full test suite unless explicitly instructed to.
+  Only run specific, targeted tests that are relevant to the changes you are making.
+
 # Tracing JuliaFormatter's output
 
 - Use the `JuliaFormatter.Internal.format_to_stage` function to inspect the output of each stage of the formatting process.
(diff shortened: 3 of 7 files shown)


penelopeysm:
!formatbot TuringLang/DynamicPPL.jl

penelopeysm:
!formatbot SciML/Catalyst.jl style=sciml ignore=["test"]

penelopeysm:
!formatbot penelopeysm/FlexiChains.jl

penelopeysm:
!formatbot trixi-framework/Trixi.jl use_config=true

penelopeysm:
!formatbot trixi-framework/Trixi.jl use_config=true

penelopeysm:
!formatbot trixi-framework/TrixiParticles.jl use_config=true

HISTORY_v2.md
```diff
@@ -1,3 +1,14 @@
+# v2.6.0
+
+Fixed a number of cases where the left-hand operand of binary operators would be aggressively nested.
+(In general, it is better to nest the right-hand operand as that keeps as much of the operation as possible on the same line.)
+These fixes were applied to DefaultStyle and thus should propagate to other styles as well.
+However, SciMLStyle goes slightly further than this and also allows expressions to extend beyond the stated margin in the interests of not nesting the LHS. (#998, #1012)
+
+Added a new formatting option, `sciml_margin_overrun`, which controls the extent to which SciMLStyle allows expressions to extend beyond the margin in order to avoid nesting the LHS of binary operators.
+The default is 20, but it can be set to 0 to prevent this behaviour if undesired.
+This option has no effect for styles apart from SciML. (#998)
```
penelopeysm, line 10: new feature so should be minor not patch

penelopeysm:
Honestly, I think that looks great.

penelopeysm:
!formatbot penelopeysm/FlexiChains.jl

penelopeysm:
!formatbot TuringLang/DynamicPPL.jl

penelopeysm:
!formatbot TuringLang/DynamicPPL.jl use_config=true

penelopeysm:
Darn that makes me happy

JoshuaLampert:
Just to double-check I understand the bot correctly (this is really cool btw!): It formats the repo with JuliaFormatter.jl master and with JuliaFormatter.jl from the PR and compares the results? I.e., the way the code is currently formatted on the main branch of the repo does not play a role?

penelopeysm:
Yes, that's exactly right! There are definitely some downsides to this, most notably, I (and probably you :)) would like to know whether bumping the formatter version will introduce huge diffs vs main. I'd like to extend it to show the main-vs-(formatted with PR) diffs, and also main-vs-(formatted with master) diffs, but I haven't yet figured out how to elegantly display a three-way diff haha

JoshuaLampert:
Ok, thanks for the confirmation. That totally makes sense. For the current status I think it makes more sense to have the current version where main of the repo does not show up in the diff because it is easier to see the improvement from the PR compared to the current version of JuliaFormatter.jl master. For example for Trixi.jl the diff is currently still really big between v1 of JuliaFormatter.jl and v2, but once the regressions are fixed it might make more sense to see how current main of Trixi.jl changes with the new versions of JuliaFormatter.jl v2 (and hopefully at some point only see improvements in that diff :)).
Anyway, thanks a lot for the effort you put into JuliaFormatter.jl!

penelopeysm:
> because it is easier to see the improvement from the PR compared to the current version of JuliaFormatter.jl master

Yup-- it was intended as a way for me to track what this PR specifically does :) But what you said actually made me think a bit more, and I think it can quite often be the case that the repos I test against _are_ already formatted with JuliaFormatter. In which case, if we want to see the difference of (target repo main) vs (target repo formatted with this PR), all I need to do is to change a config setting for which revisions of JuliaFormatter I compare. For trixi that would mean just changing the old revision to v1.

> Anyway, thanks a lot for the effort you put into JuliaFormatter.jl!

Thank you! 😊

PR #1021, opened 2026-06-01
Fix jlfmt not picking up style from config file + ignore paths not working on Windows

### LINKED ISSUE #898: Ignore field does not work on windows with unix path separator
If we use `ignore = ["foo/bar"]`, the `foo/bar` directory will not be ignored on windows.

### LINKED ISSUE #951: When using CLI `jlfmt`, `style` is never taken from config file, unless switch `--prioritize-config-file` is given
It is possible to define the style to be used in the config file, using `style = "yas"`, for example. When using the CLI, however, this option is disregarded, unless `--prioritize-config-file` is given.

The root cause of this is that inside the `main` function, the CLI options are merged onto the options read from the config file. This is fine for most options, the style option, however, is always set, even if it is not given explicitly in a CLI option:

```jl
    format_options[:style] = if style_name == "default"
        DefaultStyle()
    elseif style_name == "yas"
        YASStyle()
    elseif style_name == "blue"
        BlueStyle()
    elseif style_name == "sciml"
        SciMLStyle()
    elseif style_name == "minimal"
        MinimalStyle()
    else
        return panic("unknown style: \"$style_name\"")
    end
```

penelopeysm:
Closes #951
Closes #898

This could be merged and released as-is, but I think the fact that the CLI app has around 1000 LoC and a plethora of options but is completely untested is not ideal. I'll add more tests for the CLI before merging.

### DIFF
--- a/.github/workflows/PreCommit.yml
+++ b/.github/workflows/PreCommit.yml
@@ -15,6 +15,10 @@ on:
   pull_request:
   workflow_dispatch:
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ startsWith(github.ref, 'refs/pull/') }}
+
 jobs:
   pre-commit:
     runs-on: ubuntu-latest
--- a/.github/workflows/PreCommitHooks.yml
+++ b/.github/workflows/PreCommitHooks.yml
@@ -10,6 +10,10 @@ on:
     branches:
       - 'master'
 
+concurrency:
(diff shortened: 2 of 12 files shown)


penelopeysm:
I've at least added tests for the CLI-specific options. The formatting options like `--indent` or `--margin` are untested, but I think those are a bit less pressing to test.

penelopeysm:
Gee, the ignore flag is failing for files in subdirectories on Windows, because of path separator differences. Classic.

penelopeysm:
That is #898.

PR #1023, opened 2026-06-01
Fix various issues with `pipe_to_function_call`

### LINKED ISSUE #439: VSCode Extension Replaces Pipes with Function Calls
Even with the setting turned off, VSCode replaces pipes with function calls. 

It also does this when not appropriate -- I'm using Tullio, and `@tullio` uses the pipe to indicate functions that should be applied after reduction. This means `@tullio x := y[i] |> f` is not the same as `@tullio x := f(y[i])` -- the former takes a sum, then applies `f`, while the latter applies `f` and then takes a sum. Similarly, macros that take keyword arguments (like `@tullio`) have spaces inserted, even when the settings specify that keywords should not have spaces. To deal with situations like this, maybe there should be an option to disable formatting of lines that have a macro on them that's not defined in base?

### LINKED ISSUE #647: Broadcast collection of functions
I ran into a bug, where the formatter will change the meaning of code:
```julia
julia> 1 .|> (sin, cos)
(0.8414709848078965, 0.5403023058681398)

julia> format_text("1 .|> (sin, cos)"; pipe_to_function_call=true)
"(sin, cos)(1)"

julia> (sin, cos)(1)
ERROR: MethodError: objects of type Tuple{typeof(sin), typeof(cos)} are not callable
 ...

julia> (sin, cos).(1)
ERROR: MethodError: objects of type Tuple{typeof(sin), typeof(cos)} are not callable
 ...
```

This is, at least in part, an issue with Julia syntax, because there's nothing stopping me from defining
```julia
julia> f = (sin, cos)
(sin, cos)
```
and then 
```julia
julia> 1 .|> f
(0.8414709848078965, 0.5403023058681398)
```
isn't identical to 
```julia
julia> f.(1)
ERROR: MethodError: objects of type Tuple{typeof(sin), typeof(cos)} are not callable
```

But perhaps in the special case of the consumer in a broadcasted pipe being an array or tuple, JuliaFormatter could leave the pipe be?

(This is not urgent for me, just wanted to report it while I saw it)

### LINKED ISSUE #927: Incorrect reformatting of broadcasted pipe, i.e., `.|>` expression
**Package**: JuliaFormatter.jl  
**Version**: v2.1.2
**Julia Version**: v1.11.6

### Description
`JuliaFormatter.jl` incorrectly reformats a broadcasted function call using the `|>` operator, producing invalid Julia code by omitting the dot (`.`) in the broadcasted call syntax.

### Input Code
```julia
f() = Float32

a = 1:10

a .|> f()
```

### Expected Output
```julia
f() = Float32

a = 1:10

f().(a)
```

### Actual Output
```julia
f() = Float32

a = 1:10

f()(a)
```

The output `f()(a)` is invalid, as it lacks the dot (`.`) required for broadcasting, altering the code’s semantics.

### Configuration
```toml
style = "blue"
short_to_long_function_def = false
```

### Steps to Reproduce
1. Save the input code in a file (e.g., `test.jl`).
2. Use the above `JuliaFormatter.toml` configuration.
3. Run `format("test.jl")`.
4. Check the reformatted output.

penelopeysm:
Closes #439
Closes #647
Closes #927

1. The transformation is fundamentally unsafe, and is enabled by default for Blue and YAS styles. This patch therefore causes JuliaFormatter to emit a warning when this transformation is invoked (with maxlog=1 to avoid being spammy). Unfortunately, in VSCode it will be swallowed into the LSP logs, but this is pretty much all I can do without outright removing the option.

2. This PR disables the pipe-to-function-call transformation, _even if the option is enabled_, in three cases:
    - inside macro calls (#439, out of an abundance of caution); 
    - the weird broadcast-over-function case (#647, there's no valid transformed output for this construct); and
    - inside `Expr`s (transforming it causes the `Expr` to become something completely different, which is incorrect).

3. This PR also fixes a number of other inconsistencies. Example:

```julia
using JuliaFormatter

s = """
x |> f
x |> !
x .|> !
x .|> f()
(x) |> f
x |> a.b
x |> a + b
"""

format_text(s; pipe_to_function_call=true) |> print
```

Here's the difference on current release (v2.5.4) vs this PR. As can be seen, this PR fixes several cases where the formatted text is either different to the original, or invalid Julia code. (Note that case 4 is #927)

```
     input          v2.5.4       this PR       notes

1    x |> f         f(x)         f(x)          unchanged
2    x |> !         !(x)         !(x)          unchanged
3    x .|> !        !.(x)        .!(x)         v2.5.4 is invalid code
4    x .|> f()      f()(x)       f().(x)       v2.5.4 transforms code wrongly
5    (x) |> f       f((x))       f(x)          v2.5.4 has extraneous parentheses
6    x |> a.b       a.b(x)       a.b(x)        unchanged
7    x |> a + b     a + b(x)     (a + b)(x)    v2.5.4 transforms code wrongly
```

For case 7, note that `+` has higher precedence than `|>`

```julia
julia> dump(:(x |> a + b))
Expr
  head: Symbol call
  args: Array{Any}((3,))
    1: Symbol |>
    2: Symbol x
    3: Expr
      head: Symbol call
      args: Array{Any}((3,))
        1: Symbol +
        2: Symbol a
        3: Symbol b
```

4. Fixes some cases where formatting with `pipe_to_function_call=true` was not idempotent, probably because of the transformation not making the same nesting decisions as it usually would. e.g.

```julia
s2 = """
function f()
    ps_mods = map(
        layer_mods -> (
            layer_mods === nothing ? () :
                map(l -> initialparameters(rng, l), layer_mods) |> Tuple
        ),
        mods
    ) |> Tuple
end
"""

using JuliaFormatter, Test
f(s) = format_text(s; pipe_to_function_call=true)
@test f(s2) == f(f(s2))
```

fails on v2.5.4, passes on this PR. (FWIW this is the output:)

```julia
julia> f(s2) |> print
function f()
    ps_mods = Tuple(
        map(
            layer_mods -> (
                layer_mods === nothing ? () :
                Tuple(map(l -> initialparameters(rng, l), layer_mods))
            ),
            mods,
        ),
    )
end
```

### DIFF
--- a/HISTORY_v2.md
+++ b/HISTORY_v2.md
@@ -7,6 +7,9 @@ Fixed a number of issues with `pipe_to_function_call=true`:
 - Transforming pipes inside macros was fundamentally dangerous.
   This patch conservatively refuses to transform pipes inside macros. (#439, #1023)
 
+- Transforming pipes inside an `Expr` changes the `Expr`.
+  This patch prevents this. (#1023)
+
 - The expression `1 .|> (sin, cos)` cannot be transformed into a function call as there is no equivalent syntax for this. This patch leaves such expressions unchanged. (#647, #1023)
 
 - For the transformations that do happen, this patch elides unnecessary parentheses. For example `(x) |> f` is now transformed into `f(x)` rather than `f((x))`. (#1023)
--- a/HISTORY_v2.md
+++ b/HISTORY_v2.md
@@ -1,22 +1,43 @@
+# v2.5.5
+
+Fixed a number of issues with `pipe_to_function_call=true`:
+
+- Various constructions such as `x .|> f()`, `x .|> !`, and `x |> a + b` were being transformed into code that had a different meaning from the original. (#927, #1023)
+
+- Transforming pipes inside macros was fundamentally dangerous.
+  This patch conservatively refuses to transform pipes inside macros. (#439, #1023)
+
+- The expression `1 .|> (sin, cos)` cannot be transformed into a function call as there is no equivalent syntax for this. This patch leaves such expressions unchanged. (#647, #1023)
+
+- For the transformations that do happen, this patch elides unnecessary parentheses. For example `(x) |> f` is now transformed into `f(x)` rather than `f((x))`. (#1023)
+
+- Fixed cases where newly transformed function calls would be generated with the wrong nesting, leading to a loss of idempotency. (#1023)
+
+Finally, this patch also causes a warning to be emitted whenever a pipe is transformed into a function call. (#1023)
+
+This is because `|>` is an ordinary Julia function, and can be overloaded by users such that `x |> f` may not always be equivalent to `f(x)`.
+A formatter should _never_ change the meaning of code.
+Thus, it is possible, or even likely, that this option will be completely removed in the future (it could be turned into a _linter_ rule, for example, where the user can be alerted to the presence of `|>` in their code; but the _formatter_ should not change it).
+
 # v2.5.4
 
-Fixed a bug where `jlfmt` would not use the style set in a `.JuliaFormatter.toml` configuration file (unless `--prioritize-config-file` was specified) (#951, #1021).
+Fixed a bug where `jlfmt` would not use the style set in a `.JuliaFormatter.toml` configuration file (unless `--prioritize-config-file` was specified). (#951, #1021)
 
-Fixed a bug where JuliaFormatter (both the library and app) would not correctly ignore files in subdirectories on Windows due to path separator differences (#898, #1021).
+Fixed a bug where JuliaFormatter (both the library and app) would not correctly ignore files in subdirectories on Windows due to path separator differences. (#898, #1021)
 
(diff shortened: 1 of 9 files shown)


penelopeysm:
!formatbot SciML/ReservoirComputing.jl style=default pipe_to_function_call=true

penelopeysm:
okay that is bad, how did that even affect unrelated formatting

edit: the nesting thing is a bug in v2.5.4 as well where formatting the first time didn't get the correct nesting decisions, so is actually a fix in this PR, added a regression test

the `ortho` line is a parenthesisation issue as the original is

```julia
ortho = qr(randn(Float32, D, D)).Q[:, :] |> Matrix{Float32}
```

penelopeysm:
!formatbot SciML/ReservoirComputing.jl style=default pipe_to_function_call=true

penelopeysm:
!formatbot JuliaDiff/ChainRules.jl style=default pipe_to_function_call=true

PR #1026, opened 2026-06-02
fix block if conditions

### LINKED ISSUE #1025: if or elseif with block condition fails
```julia
julia> s = """
       if (a; b)
           foo
       end
       """
"if (a; b)\n    foo\nend\n"

julia> fs(:gn, s)
     1:21     │[if]
     1:2      │  if
     3:3      │  Whitespace
     4:9      │  [block]
     4:4      │    (
     5:5      │    Identifier           ✔
     6:6      │    ;
     7:7      │    Whitespace
     8:8      │    Identifier           ✔
     9:9      │    )
    10:18     │  [block]
    10:14     │    NewlineWs
    15:17     │    Identifier           ✔
    18:18     │    NewlineWs
    19:21     │  end


julia> ft(s)
┌ Warning: Formatted text is not parsable ... no change made.
└ @ JuliaFormatter ~/jl/others/JuliaFormatter.jl/src/JuliaFormatter.jl:280
ERROR: ParseError:
# Error @ line 1:3
if
# └ ── missing condition in `if`
```

penelopeysm:
Closes #1025

Prior to this PR, the following

```julia
if (a; b)
    foo
end
```

would get reformatted to

```julia
if
(a; b)
    foo
end
```

which would fail to parse. This patch prevents this. It also fixes the equivalent problem with `while` (even though the output of that one parses correctly, it's ugly and almost certainly unintended).

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,5 +1,6 @@
-# NOTE: Do NOT use this config in your own repositories. Instead, see 
-#   docs/integrations.md
+# NOTE: This pre-commit configuration is meant for the JuliaFormatter repo itself.
+# Do not use this config in your own repositories! Instead, see
+# https://juliaeditorsupport.github.io/JuliaFormatter.jl/stable/precommit/
 repos:
 - repo: local
   hooks:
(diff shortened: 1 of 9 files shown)


penelopeysm:
```julia
while BLOCK
    BLOCK
end
```

also has the same 'problem' -- although Julia parses it fine which is weird -- but it is ugly so we should fix it

PR #1029, opened 2026-06-03
Fix wrapped hcat ncat formatting

svchb:
## Summary

Fix formatting for wrapped horizontal concatenation rows followed by repeated semicolon n-dimensional concatenation separators.

JuliaSyntax can parse a formatted result such as `[a b;;\n c d]` as an `Hcat` node with semicolon leaves rather than as the original `Ncat` shape. The formatter previously joined that syntactic newline and inserted whitespace between semicolon tokens, producing invalid Julia like `a b ; ; c d`.

This PR preserves the required newline after wrapped ncat separators, avoids manufacturing whitespace before semicolon separators, and adds explicit Hcat nesting hooks so YAS/SciML keep their second-pass alignment.

### DIFF
--- a/test/sciml_style.jl
+++ b/test/sciml_style.jl
@@ -156,6 +156,7 @@
     end
 
     @testset "wrapped hcat with ncat separator" begin
+        # SciML's fixpoint pass reparses this as Hcat; it must keep `;;` wrapped.
         str = raw"""
         x = [f(a=1)
              f(a=1);;
--- a/test/sciml_style.jl
+++ b/test/sciml_style.jl
@@ -155,6 +155,22 @@
         @test format_text(str2, SciMLStyle()) == str2
     end
 
+    @testset "wrapped hcat with ncat separator" begin
+        str = raw"""
+        x = [f(a=1)
+             f(a=1);;
+             f(a=2)
+             f(a=2)]
+        """
+        formatted_str = raw"""
+        x = [f(a = 1) f(a = 1);;
+             f(a = 2) f(a = 2)]
+        """
+
+        @test format_text(str, SciMLStyle(); join_lines_based_on_source = false) ==
+              formatted_str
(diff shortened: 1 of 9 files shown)


svchb:
@penelopeysm I tried your suggestion from #997 but it fails immediately with incorrectly formatting hcats in our code. This fix will produce valid Julia code again with SciMLStyle.

penelopeysm (commented):
Gah, this is bad, thanks for finding it. I'll look at the code more properly when I can, please feel free to ignore this for now.

penelopeysm:
I took the liberty to update some of the code so that default-style prints nicely when over the margin. For example, with a small margin,

```julia
x = [1 2;;
    3 4]
```

will be converted to 

```julia
x = [
    1 2;;
    3 4
]
```

which is still correct, but is more in line with the way default-style nests things with brackets/parens.

test/sciml_style.jl
```diff
@@ -155,6 +155,23 @@
...
+        """
+        formatted_str = raw"""
+        x = [f(a = 1) f(a = 1);;
+             f(a = 2) f(a = 2)]
+        """
+
+        @test format_text(str, SciMLStyle(); join_lines_based_on_source = false) ==
+              formatted_str
+    end
+
```
penelopeysm, line 174: oops --- actually this test is not right -- removing the newlines changes the meaning.....

```julia
julia> x = [1
            2;;
            3
            4]
2×2 Matrix{Int64}:
 1  3
 2  4

julia> y = [1 2;;
            3 4]
1×4 Matrix{Int64}:
 1  2  3  4
```

penelopeysm: i'll sort it tomorrow. this escaped my notice - but ncat shouldn't ever be reparsed into hcat- that would be a bug.

PR #1041, opened 2026-06-06
Add `always_use_return=true` for BlueStyle

### LINKED ISSUE #906: `return` forgotten when changing an inline function `f(x) = ...` into a `function f; ...; end` block
See the example in https://github.com/gdalle/SparseMatrixColorings.jl/pull/235: the new JuliaFormatter v2.1 transformed

```julia
f(x) = verylongexpression
```
into
```julia
function f(x)
    verylongexpression
end
```
when BlueStyle says it should have been
```julia
function f(x)
    return verylongexpression
end
```

penelopeysm:
Closes #906

https://github.com/JuliaDiff/BlueStyle says "When using long-form functions [always use the return keyword](https://groups.google.com/forum/#!topic/julia-users/4RVR8qQDrUg)"

### DIFF
--- a/docs/src/contributing.md
+++ b/docs/src/contributing.md
@@ -32,6 +32,8 @@ The exact label for an issue is left to my discretion, but generally these are t
 Additionally where possible I'll also add **good first issue** tags for changes that are quite self-contained.
 These may include any priority level but should generally be P3 or lower.
 
+You can see a list of all triaged issues [here](https://github.com/orgs/JuliaEditorSupport/projects/1/views/1).
+
 ## AI usage
 
 LLMs and AI in general can be really useful for tracing through some of the logic in the codebase.
--- a/HISTORY_v2.md
+++ b/HISTORY_v2.md
@@ -1,3 +1,7 @@
+# v2.6.2
+
+Re-enabled `always_use_return=true` for BlueStyle, in line with the Blue style guide. (#906, #1041)
+
 # v2.6.1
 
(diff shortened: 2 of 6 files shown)


penelopeysm:
~~still needs a patch bump, tests, etc etc etc~~ done

docs/src/contributing.md
```diff
@@ -32,6 +32,8 @@ The exact label for an issue is left to my discretion, but generally these are t
 Additionally where possible I'll also add **good first issue** tags for changes that are quite self-contained.
 These may include any priority level but should generally be P3 or lower.
 
+You can see a list of all triaged issues [here](https://github.com/orgs/JuliaEditorSupport/projects/1/views/1).
```
schillic, line 35: @penelopeysm This link does not seem to be valid...

penelopeysm: Right! Thats because it's not public. I'll make it so.

penelopeysm: Turns out I can't make it public 😓 I'll replace the link with https://github.com/JuliaEditorSupport/JuliaFormatter.jl/issues?q=is%3Aissue%20state%3Aopen%20label%3AP5%2CP4%2CP3%2CP2%2CP1, which is pretty much the same thing.

penelopeysm: (Thanks for letting me know!)

schillic: Thanks! If you have admin access, I think you can go to https://github.com/orgs/JuliaEditorSupport/projects/1/settings and at the bottom next to _Visibility_ you can make it public.

PR #1043, opened 2026-06-06
stop using *fmt test helpers

### LINKED ISSUE #1042: `<foo>fmt` test helpers silently does the wrong thing (probably?)
https://github.com/JuliaEditorSupport/JuliaFormatter.jl/blob/71cbe42d324092dcea555299af99cbbc0a1ee75c/test/runtests.jl#L11-L43

if you trace the calls here

```julia
bluefmt(s)
-> fmt(s; style=BlueStyle())
-> fmt1(s; merge(options(DefaultStyle()), (; style=BlueStyle()))...)
== fmt(s; options(DefaultStyle())..., style=BlueStyle())
-> JuliaFormatter.format_text(s; options(DefaultStyle())..., style=BlueStyle()
-> JuliaFormatter.format_text(s, BlueStyle(), options(DefaultStyle())...)
```

but this actually doesn't test BlueStyle -- it tests BlueStyle but with defaultstyle options injected on top of it. This essentially means that many of the tests are testing the wrong thing.

penelopeysm:
Closes #1042

I got Claude to do the boring textual replacements, but that spawned a new bunch of test failures, which I looked at individually. Some of them are just because of the default style options being injected, so I changed the expected output accordingly. Some of the test failures surfaced other genuine bugs, which I'm not happy with -- however, this isn't the PR to fix them, so I've opened separate issues.

### DIFF
--- a/docs/src/contributing.md
+++ b/docs/src/contributing.md
@@ -32,7 +32,7 @@ The exact label for an issue is left to my discretion, but generally these are t
 Additionally where possible I'll also add **good first issue** tags for changes that are quite self-contained.
 These may include any priority level but should generally be P3 or lower.
 
-You can see a list of all triaged issues [here](https://github.com/orgs/JuliaEditorSupport/projects/1/views/1).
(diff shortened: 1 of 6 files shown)


penelopeysm:
I changed test/default_style.jl and that surfaced a ton more idempotence bugs, so I'll defer that to another PR.

PR #1055, opened 2026-06-07
Force JuliaSyntax to parse everything up to 1.12

### LINKED ISSUE #890: `public` keyword is misformatted
At least with YAS, you get this rewrite:

```julia
public ExcludedRecordingV1,ExcludedRecordingV1SchemaVersion
```

```julia
publicExcludedRecordingV1,ExcludedRecordingV1SchemaVersion
```

penelopeysm:
This ensures that, for example, `public Foo` is parsed correctly even when running JuliaFormatter with Julia 1.10, even though that string is not valid Julia on 1.10.

Closes #890 (properly)


penelopeysm:
This is actually a code change, so deserves a patch bump.

Issue #862, opened 2024-08-14
Strips inline comments


MilesCranmer:
```julia
julia> using JuliaFormatter

julia> format_text("T <: Tuple #=e.g., Tuple{Union{Float32,Float64}}=# && !(T isa Union) #= e.g., Union{Tuple{A},Tuple{B}}=#")
"T <: Tuple && !(T isa Union) #= e.g., Union{Tuple{A},Tuple{B}}=#"
```

Seems to strip the first inline comment completely

domluna:
I think I can get this to work in the rewrite to JuliaSyntax

domluna:
fixed with JuliaSyntax, the comments will now be there although the whitespace might be different for the time being (0 vs 1 space or vice-versa), but that is something that can be improved later on.

penelopeysm:
The whitespace is indeed ugly, but the comment is preserved on current release (v2.6.4)

Issue #571, opened 2022-05-01
Error message not helpful


vchuravy:
The file being formatted has 3771 lines so line 4916 is not that helpful, the code is Enzyme.jl `compiler.jl` in https://github.com/EnzymeAD/Enzyme.jl/pull/294

```
 Error occurred on line 4916, offset 33.
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:35
  [2] format_text(cst::CSTParser.EXPR, style::DefaultStyle, s::JuliaFormatter.State)
    @ JuliaFormatter ~/.julia/packages/JuliaFormatter/ohMeP/src/JuliaFormatter.jl:590
  [3] format_text(text::String, style::DefaultStyle, opts::JuliaFormatter.Options)
    @ JuliaFormatter ~/.julia/packages/JuliaFormatter/ohMeP/src/JuliaFormatter.jl:538
  [4] #format_text#206
    @ ~/.julia/packages/JuliaFormatter/ohMeP/src/JuliaFormatter.jl:530 [inlined]
  [5] #format_text#205
```

domluna:
That refers to the formatted output so that's why the line numbers may not make much sense but I'll see about mapping it back to the input text

domluna:
gonna put up for a PR to help with debugging this but in the meantime I found the error

the formatted section

```
       function arraycopy_rev(
            B::LLVM.API.LLVMBuilderRef,
            OrigCI::LLVM.API.LLVMValueRef,
            gutils::API.EnzymeGradientUtilsRef,
            tape::LLVM.API.LLVMValueRef,
        )::Cvoid
            orig = LLVM.Instruction(OrigCI)
            origops = LLVM.operands(orig)
            arraycopy_common(false, LLVM.Builder(B), orig, origops[1], gutils)#=fwd=#
            return nothing
       end
```

https://github.com/EnzymeAD/Enzyme.jl/blob/main/src/compiler.jl#L1670-L1675

for some reason CSTParser finds an error during parsing for this block

```
julia> CSTParser.parse("""
       arraycopy_common(false, LLVM.Builder(B), orig, origops[1], gutils)#=fwd=#
       return nothing
       """)
  1:89  errortoken
  1:74   call
  1:16    arraycopy_common
 17:21    FALSE: false
 22:36    call
 22:33       1:1   OP: .
 22:25      LLVM
 26:32      quotenode
 26:32       Builder
 34:34     B
 37:40    orig
 41:50    ref
[...]

PR #1061, opened 2026-06-08
Fix macrocall do-blocks, comprehension body assignments, and dropped #= =# comments

### LINKED ISSUE #562: Comments inside function calls are moved to the end
Not sure if this is a conscious choice, or a bug:
```
julia> using JuliaFormatter

julia> code = """
       f(x, #= y,=# z)
       """;

julia> print(code)
f(x, #= y,=# z)

julia> print(format_text(code))
f(x, z) #= y,=#
```

### LINKED ISSUE #905: Changed behavior of `always_for_in` option
The [documentation](https://domluna.github.io/JuliaFormatter.jl/stable/#always_for_in) of `always_for_in` says:

> > default: `false`
>
> If true, `=` is always replaced with `in` if part of a `for` loop condition.
> For example, `for i = 1:10` will be transformed to `for i in 1:10`. Set
> this to `nothing` to leave the choice to the user.

1. How can I pass `nothing`? Is it meant that this option is simply not defined in the config file?

2. In v2, setting `always_for_in=false` actually changes `in` to `=`. Since the documentation does not mention that, it looks like a regression.
Otherwise, the documentation should be updated. But notice that this implies enforcing a code style both with settings `true` and `false`, and the option name being misleading.

3. In v2, using `always_for_in=true` (or no setting) with the YAS/Blue styles modifies the input

```julia
all(v ∈ P for v in mylist)
```
to
```julia
all(v in P for v in mylist)
```

The `∈` here should not be modified because it is not part of the `for` loop.

### LINKED ISSUE #944: Assignment replaced with in operator when using list comprehension
**JuliaFormatter version**: 2.1.6
**Julia version**: 1.11.5

Given
```
x = repeat([""], 5)
y = ["a", "b", "c", "d", "e"],
```
the following line
```
[x[i] = y[i] for i = 1:length(y)]
```
gets converted to
```
[x[i] in y[i] for i = 1:length(y)]
```
when calling `format` on a file that contains this. This is not the end of the world, as that line throws a
```
ERROR: LoadError: use occursin(needle, haystack) for string containment
```

However, the same replacement `=` -> `in` happens when given
```
x = zeros(5)
y = rand(5)
```
which completely and silently changes the the meaning of the constructed vector as the code runs without error.

I realize these are not common patterns, but this seems like a bug nonetheless.

### LINKED ISSUE #946: weird behaviour with multiline comments
Formatting this with  default configuration of 2.1.6
```julia
function f()
    #==Pairs Tuple or ValidationResult==#
    #=
    So far
    =#
end

```
we get

```julia
#=
So far
=#
function f()
    #==Pairs Tuple or ValidationResult==#

    #=
    So far
    =#
end

```
Then, at each consecutive formatting the 3 first times are replicated again and again

### LINKED ISSUE #1018: Macro do-block with keyword-style macro args is partially unformatted in v2.5.0+
On JuliaFormatter v2.5.0 and newer, a macro call using `do` syntax with keyword-style macro arguments appears to be partially skipped by the formatter. The same shape as a normal function call is formatted, but the macro-call form preserves extra spaces in the macro call and can leave multiline `do` arguments unindented.

Minimal reproducer:

```julia
using JuliaFormatter

src = """
expr_spec = @template_spec(parameters=(p1=10, p2=10, p3=1),  expressions=(f, g))  do x1,
x2,
class

    x1 * 2
end
"""

print(format_text(src; style=BlueStyle()))
```

Actual output on v2.5.2, and current `main` as of 2026-05-31:

```julia
expr_spec = @template_spec(parameters=(p1=10, p2=10, p3=1),  expressions=(f, g))  do x1,
x2,
class
    x1 * 2
end
```

The analogous function-call form is formatted much more reasonably:

```julia
expr_spec = template_spec(;
    parameters=(p1=10, p2=10, p3=1), expressions=(f, g)
) do x1, x2, class
    x1 * 2
end
```

I checked released JuliaFormatter versions with Julia 1.12.5:

- v1.0.0 through v2.4.0 format the snippet, although they split the assignment awkwardly:

```julia
expr_spec =
    @template_spec(parameters=(p1=10, p2=10, p3=1), expressions=(f, g)) do x1, x2, class
        x1 * 2
    end
```

- v2.5.0, v2.5.1, v2.5.2, and current `main` produce the partially unformatted output shown above.

A shorter case also preserves extra spaces in the macro call on v2.5.2:

```julia
julia> print(format_text("y = @foo(a=1,  b=2)  do x,\ny\n    x + y\nend\n"; style=BlueStyle()))
y = @foo(a=1,  b=2)  do x, y
    x + y
end
```

Expected behavior: the macro-call `do` form should either be formatted consistently with other macro calls / do-blocks, or at least not preserve extra spaces and unindented multiline `do` arguments. If the assignment split from <= v2.4.0 is the intended fallback, that still seems less surprising than the v2.5.x output.

jw3126:
Closes #905
Closes #944
Closes #946
Closes #1018

Several independent formatting bugs, each with a regression test (full test suite green).

### 1. Macrocall followed by a `do`-block gains spurious whitespace

A macrocall with both parenthesized args and a `do`-block was treated as a macroblock, because the trailing `do` node is not a closer. This produced a double space before `do`, extra whitespace between args, and left multiline `do` args unindented:

```julia
# before
y = @foo(a=1,  b=2)  do x,
y
    x + y
end

# after
y = @foo(a=1, b=2) do x, y
    x + y
end
```

Fix: route macrocalls with a `do`-block through `p_do_call`, and give `p_macrocall` a `do_block_idx` keyword (mirroring `p_call`) so the head is formatted as a normal closer-terminated macrocall and the `do`-block is appended afterwards.

### 2. `always_for_in` rewrites a comprehension body / membership test (AST change)

With `always_for_in` (the default for Blue/YAS), the comprehension *body* was normalized too, changing the meaning of the code:

```julia
# before  (always_for_in = true)
[x[i] in y[i] for i in 1:length(y)]   # was: [x[i] = y[i] for i = 1:length(y)]
all(v in P for v in mylist)           # was: all(v ∈ P for v in mylist)

# after
[x[i] = y[i] for i in 1:length(y)]
all(v ∈ P for v in mylist)
```

Fix: in `p_generator` (both the default and the YAS style, which has its own copy), only normalize iteration specifications between the `for` and `if` keywords, never the body that precedes `for`.

### 3. `#= =#` comments inside blocks are dropped / duplicated

`#= =#` comments are reported as whitespace by JuliaSyntax, so block handlers skipped (deleted) them; a leading comment could also corrupt line tracking and duplicate unrelated comments:

```julia
# before  (the leading single-line comment is dropped)
function f()

    #=
    So far
    =#
end

# after
function f()
    #==Pairs Tuple or ValidationResult==#
    #=
    So far
    =#
end
```

Fix: preserve them via a small `add_hasheq_comment!` helper. A leading comment must also establish the block's line tracking; otherwise the custom-leaf path copies `t.endline` (still `0`) and the following node computes a bogus `NOTCODE` range that duplicates unrelated comments.

### 4. Inline `#= =#` comment glued to the next argument

For the same reason (comments reported as whitespace, source whitespace dropped), an inline comment in a call was glued to the following argument:

```julia
# before
f(x, #= y,=#z)

# after
f(x, #= y,=# z)
```

Fix: emit a space after an inline `#= =#` comment when the following code stays on the same line. This matches the existing "glue before, keep the space after" convention already asserted by the #571 and #862 tests.

### Resolved issues

- Closes #1018 (macro `do`-block with keyword-style args partially unformatted) — fix 1
- Closes #944 (assignment replaced with `in` in a list comprehension) — fix 2
- Closes #946 (block `#= =#` comment dropped / not idempotent) — fix 3
- Closes #562 (inline comment in a function call glued / moved) — fix 4
- Addresses point 3 of #905 (`∈` in a generator body wrongly normalized under `always_for_in`/Blue/YAS) — fix 2. Points 1–2 of that issue concern the option's documented semantics and are out of scope here.

Regression tests are added in `test/issues.jl` as issue-numbered testsets (#562, #905, #944, #946, #1018).


penelopeysm:
Thanks for picking off some of the highest priority issues! 😄 These look like quite reasonable fixes at first glance. I'll review more carefully later today.

src/styles/default/pretty.jl
```diff
@@ -3900,11 +3951,15 @@ function p_generator(
             add_node!(t, n, s; join_lines = true)
         end
 
-        if from_for && !past_if
+        if from_for && seen_for && !past_if
```
penelopeysm: I think we could accomplish this more directly (and arguably more correctly -- because we're testing the exact predicate we want) by restricting the transformation to the CST node that needs to be targeted. I replaced this line with `if from_for && kind(a) === K"iteration"` (both `seen_for` and `past_if` can then be completely removed) and it behaves nicely (and all tests pass).

I'll take the liberty of just pushing these changes directly since I already have them written up :)

```julia
s = """
[x = f() for i = xs]
(x = f() for i = xs)
[j = 1 for j = (i = 1 for i = [1, 2, 3])]
(j = 1 for j = (i = 1 for i = [1, 2, 3]))
[j = 1 for j = (i = 1 for i = [1, 2, 3]) if (j = 1) ∈ 2]
(j = 1 for j = (i = 1 for i = [1, 2, 3]) if (j = 1) ∈ 2)
"""

format_text(s; always_for_in=true) |> println
```

gives

```julia
[x = f() for i in xs]
(x = f() for i in xs)
[j = 1 for j in (i = 1 for i in [1, 2, 3])]
(j = 1 for j in (i = 1 for i in [1, 2, 3]))
[j = 1 for j in (i = 1 for i in [1, 2, 3]) if (j = 1) ∈ 2]
(j = 1 for j in (i = 1 for i in [1, 2, 3]) if (j = 1) ∈ 2)
```

penelopeysm (approved):
The other changes look perfect! I'll try to see if I can find any idempotence edge cases with the inline comments, if not, then very happy to merge & release. Thanks :)

PR #1071, opened 2026-06-09
Fix non-idempotent blank line after a #= =# comment in a parameter list

### LINKED ISSUE #1070: Non-idempotent: blank line inserted after a #= =# comment in a nested parameter list
**JuliaFormatter version**: 2.6.8
**Julia version**: 1.12

A `#= =#` comment placed on its own line inside a nested keyword-argument (`parameters`) list gains an extra blank line on **every** format pass, i.e. formatting is not idempotent. With the default options the blank lines accumulate without bound; with `remove_extra_newlines = true` it is capped at a single spurious blank line (but the input still changes on reformat).

### MWE

```julia
using JuliaFormatter

s = """
function f(
    obj;
    #=c=#
    keyword_argument_one,
    keyword_argument_two,
)
    return obj
end
"""

print(format_text(s))
```

Output — a blank line is inserted after `#=c=#`:

```julia
function f(
    obj;
    #=c=#

    keyword_argument_one,
    keyword_argument_two,
)
    return obj
end
```

Formatting that output again inserts yet another blank line, and so on. Counting blank lines after each successive pass with the default options gives `[1, 1, 2, 2, 3, 3]` — it never stabilises.

### Cause

When a `HASHEQCOMMENT` leaf is appended in `add_node!`, its `startline`/`endline` are overwritten with `t.endline` (the previous node's line) instead of the comment's real source line. The `NOTCODE` range computed for the following argument then treats the comment's line as uncovered and emits an extra newline.

A fix is in the PR linked below.

jw3126:
Fixes #1070.

A `#= =#` comment on its own line inside a nested keyword-argument (`parameters`) list gains an extra blank line on **every** format pass — unbounded with the default options, capped at a single spurious blank line by `remove_extra_newlines`.

### Before

```julia
julia> print(format_text("""
       function f(
           obj;
           #=c=#
           keyword_argument_one,
           keyword_argument_two,
       )
           return obj
       end
       """))
function f(
    obj;
    #=c=#
                          # <- blank line inserted, and another on each reformat
    keyword_argument_one,
    keyword_argument_two,
)
    return obj
end
```

### Cause

When a `HASHEQCOMMENT` leaf is appended in `add_node!`, its `startline`/`endline` were overwritten with `t.endline` (the previous node's line) instead of the comment's real source line. The `NOTCODE` range computed for the following argument then treated the comment's line as uncovered and emitted an extra newline.

### Fix

Preserve the comment's true source position when appending it, and propagate it to the parent's `endline`.

Regression test added in `test/issues.jl`. Full test suite green.

### DIFF
--- a/HISTORY_v2.md
+++ b/HISTORY_v2.md
@@ -1,3 +1,9 @@
+# v2.6.9
+
+Fixed a bug where formatting inline comments on their own line caused extra blank lines to be added. (#1070, #1071)
+
+Slightly improved the separation of inline comments from code. (#1071)
+
 # v2.6.8
 
 Fixed a bug where end-of-line comments were being silently dropped in YASStyle. (#690, #1046, #1068)
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "JuliaFormatter"
 uuid = "98e50ef6-434e-11e9-1051-2b60c6c9e899"
-version = "2.6.8"
+version = "2.6.9"
 authors = ["Dominique Luna <dluna132@gmail.com> and contributors"]
 
 [deps]
--- a/src/fst.jl
+++ b/src/fst.jl
@@ -1069,8 +1069,19 @@ function add_node!(
             end
         end
         t.len += length(n)
-        n.startline = t.endline
-        n.endline = t.endline
+        if n.typ === HASHEQCOMMENT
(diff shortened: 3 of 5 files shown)


penelopeysm (commented):
Thanks for the spot! I'll refactor this a little bit so that the HASHEQCOMMENT logic is on its own.

Issue #1080, opened 2026-06-09
Yet another ncat/hcat bug


penelopeysm:
```julia
julia> s = "[1 2 ;;\n 3 4 ; 2 3 4 5]"
"[1 2 ;;\n 3 4 ; 2 3 4 5]"

julia> Meta.parse(s)
:([1 2 3 4; 2 3 4 5])

julia> ft(s)
┌ Warning: Formatted text is not parsable ... no change made.
└ @ JuliaFormatter ~/jl/jf/src/JuliaFormatter.jl:296
ERROR: ParseError:
# Error @ line 2:8
[
    1 2;; 3 4;
#      └─┘ ── cannot mix space and ;; separators in an array expression, except to wrap a line
Stacktrace:
```

this seems like a recent regression ... no idea why/where/etc

penelopeysm:
actually not a recent regression since it's present on v2.5.0 ... phew! not my fault 🙄

penelopeysm:
I gave Fable 5 a spin on this. (I know this isn't my blog, but whatever.) Just getting it to analyse the issue, without any actual code generation, took 16% of my 5-hour limit on a Pro plan. The output isn't really 100% correct fwiw (but to be fair Julia's array literal syntax and the way that it gets parsed into CSTs is pretty complicated).

On the plus side, it's about 90% of the way there. I'm also amazed that it told me _which file_ it was referring to, because Opus 4.6 has a horrible habit of telling me line numbers but not the file.

My main point of contention with it is that it's not true that newline-after-two-semicolon is always semantically important: for example in `s2 = "[1; 2 ;;\n 3; 4 ;;; 5; 6 ;;\n 7; 8]"` the newlines can be safely dropped. It's only in the case where within a row there is a mixture of whitespace + `;;\n`, i.e., the case that `p_hcat` picks up, but applied to `p_row` as well.

Also, Fable 5 is correct in that the comment I wrote here is not accurate

https://github.com/JuliaEditorSupport/JuliaFormatter.jl/blob/440db8ea3066017374ef6b15ceac3de4946736c6/src/styles/default/pretty.jl#L3816-L3820

(line (4) is wrong), but it didn't discern that the comment is true for any other number of semicolons != 2, and that it was a typo on my part. Fair enough though, I can't expect it to read my mind.

[...]

PR #1081, opened 2026-06-09
Fable 5 fix for #1080

penelopeysm:
This PR was two-shotted with Fable 5 (could have been one-shotted, but I wanted to see its analysis before getting it to generate code). As described in #1080, analysing the bug and implementing this took 28% of my 5-hour usage limit on a Pro plan. This was also a fairly self-contained bug that's small in scope (it's just one function plus a helper function), so I can definitely imagine a larger-scale project like implementing a new feature would guzzle even more tokens.

I personally wouldn't do everything exactly the way it did, but it did do pretty well tbh.

This isn't meant for merging, more for me to write some thoughts on how well it did.

### DIFF
--- a/HISTORY_v2.md
+++ b/HISTORY_v2.md
@@ -1,3 +1,7 @@
+# v2.6.12
+
+Fixed a bug where a `;;` line continuation inside a space-separated row of a `vcat` or `ncat` expression (e.g. `[1 2 ;;\n 3 4 ; 2 3 4 5]`) had its newline removed, producing invalid code that mixed `;;` with space separators.
+
 # v2.6.11
 
 Fixed a bug where linebreaks were not preserved in the middle of a chain of binary operators when a comment was present (e.g. `a + # comment\n b + c`), causing any part of the chain after the comment to be lost. (#1076, #1077)
--- a/src/styles/default/pretty.jl
+++ b/src/styles/default/pretty.jl
@@ -3573,17 +3573,26 @@ end
 """
     is_newline_after_2semicolons(cst::JuliaSyntax.GreenNode, i::Int)
 
-Detect if child node `i` of `cst` is a newline that follows two semicolons. For example,
-this will detect the newline in constructs such as
+Detect if child node `i` of `cst` is a newline that follows exactly two semicolons. For
+example, this will detect the newline in constructs such as
 
     [a b;;
      c d]
+
+In an array expression that contains space separators, `;;` followed by a newline is a
+line continuation (the example above is the same as `[a b c d]`), and the newline must
+not be removed because `;;` cannot otherwise be mixed with space separators.
+
+Note that this requires *exactly* two semicolons: runs of three or more semicolons, e.g.
+`[a b ;;;\\n c d]`, are dimension separators which *may* legally be mixed with space
+separators on a single line, so their trailing newline can be safely removed.
 """
 function is_newline_after_2semicolons(cst::JuliaSyntax.GreenNode, i::Int)
     return i >= 3 &&
            kind(cst[i]) === K"NewlineWs" &&
            kind(cst[i-1]) === K";" &&
-           kind(cst[i-2]) === K";"
+           kind(cst[i-2]) === K";" &&
+           (i == 3 || kind(cst[i-3]) !== K";")
 end
 
 function p_hcat(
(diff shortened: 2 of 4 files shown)


src/styles/default/pretty.jl
```diff
@@ -3816,9 +3825,22 @@ end
...
+# collapsed: a run of exactly two semicolons followed by a newline. In a space-separated
+# row this is a line continuation, e.g.
+#
+#     [1 2 ;;
+#      3 4 ; 5 6 7 8]
+#
+# is the same as [1 2 3 4; 5 6 7 8]. Here JuliaSyntax places the `;;` and the newline
+# inside the first `row` node. The newline cannot be removed, because `;;` cannot be
+# mixed with space separators on a single line ([1 2;; 3 4] is a parse error). See also
+# the corresponding comment in `p_hcat`, which has to deal with the same construct.
```
penelopeysm, line 3842: This comment reads better than most LLM-generated ones -- unsure if the model is better or if it's just mimicking the surrounding style? Unfortunately, I think that instead of tacking on a new paragraph, it would have been better to do a larger rewrite of the comment as a whole so that it flows better. This is a recurring problem with LLMs IME - they tend to do local fixes rather than considering the broader context.

src/styles/default/pretty.jl
```diff
@@ -3573,17 +3573,26 @@ end
...
     [a b;;
      c d]
+
+In an array expression that contains space separators, `;;` followed by a newline is a
+line continuation (the example above is the same as `[a b c d]`), and the newline must
+not be removed because `;;` cannot otherwise be mixed with space separators.
+
+Note that this requires *exactly* two semicolons: runs of three or more semicolons, e.g.
+`[a b ;;;\\n c d]`, are dimension separators which *may* legally be mixed with space
+separators on a single line, so their trailing newline can be safely removed.
```
penelopeysm, line 3588: This is very nice

src/styles/default/pretty.jl
```diff
@@ -3875,9 +3897,15 @@ function p_row(
...
         elseif JuliaSyntax.is_whitespace(a)
-            # is_semantically_important_newline handles case (a)
-            if ctx.from_nrow &&
-               is_semantically_important_newline(cst, i, is_last_arg_of_parent)
+            if kind(cst) === K"row" && is_newline_after_2semicolons(cst, i)
+                # `;;` followed by a newline in a space-separated row is a line
+                # continuation, and the newline must be kept; see the comment above this
+                # function.
+                add_node!(t, Newline(; nest_behavior = AlwaysNest), s)
+                forced_newline = true
```
penelopeysm, line 3905: I think the thing I'm most unconvinced about by Fable's analysis is the fact that it gates the fix to `kind(cst) === K"row"`. Now, it's certainly enough to get the tests to pass, and in fact I do think it's correct (AFAIK, this scenario where there's a mixture of spaces and `;;\n` can only happen in `K"hcat"` and `K"row"`). But even so I think this code needs some justification as to why that's the case, for example, what the difference between `row` and `nrow` is.

src/styles/default/pretty.jl
```diff
@@ -3875,9 +3897,15 @@ function p_row(
...
+            elseif ctx.from_nrow &&
+                   is_semantically_important_newline(cst, i, is_last_arg_of_parent)
+                # is_semantically_important_newline handles case (a)
```
penelopeysm, line 3908: I appreciate that it retained this comment when it shifted it, because Opus has an incredibly annoying habit of deleting pre-existing comments

test/multidimensional_array.jl
```diff
@@ -127,6 +129,28 @@ end
...
+            "$(T)[\n    1 2;;\n    3 4;\n    2 3 4 5\n]",
+        )
+        # Continuations in rows inside an ncat
+        test_format(
+            "$(T)[1 2 ;;\n 3 4 ;;; 5 6 ;;\n 7 8]",
+            "$(T)[\n    1 2;;\n    3 4;;;\n    5 6;;\n    7 8\n]",
+        )
+        # A run of three or more semicolons is a dimension separator, not a line
+        # continuation; it may legally be mixed with space separators on a single line.
+        test_format("$(T)[1 2 ;;; 3 4]", "$(T)[1 2;;; 3 4]")
```
penelopeysm, line 150: The input to this test should probably have a newline after the `;;;` too (or a separate test with that added).

HISTORY_v2.md
```diff
@@ -1,3 +1,7 @@
+# v2.6.12
+
+Fixed a bug where a `;;` line continuation inside a space-separated row of a `vcat` or `ncat` expression (e.g. `[1 2 ;;\n 3 4 ; 2 3 4 5]`) had its newline removed, producing invalid code that mixed `;;` with space separators.
```
penelopeysm, line 3: Nice!

PR #1087, opened 2026-06-11
Add regression test against external repos

penelopeysm:
Clauded with human review.

This workflow checks that JuliaFormatter can format a repo and that formatting is idempotent. Right now the only case is JuliaLang/julia@v1.12.6, for which formatting succeeds, but is not idempotent 😢.


penelopeysm:
merged this into #1090 offline

PR #1097, opened 2026-06-13
Fix format:on end of file bug

### LINKED ISSUE #949: `format: off` deletes marked source code if a trailing whitespace is present
the code should not be deleted in either case, but on 2.2.0 I get the following:

```
julia> format_text("""
       #! format: off
       1+1
       #! format: on

       """)
""

julia> format_text("""
       #! format: off
       1+1
       #! format: on
       """)
"#! format: off\n1+1\n#! format: on\n"
```

penelopeysm:
Even with LLM assistance this was a really tricky one to hunt down...

Closes #949

### DIFF
--- a/HISTORY_v2.md
+++ b/HISTORY_v2.md
@@ -4,6 +4,8 @@ Fixed a bug where the combination of `always_use_return` and `short_circuit_to_i
 
 Fixed a bug where `short_circuit_to_if` would fire inside the condition of a `while` loop, leading to invalid code. (#940, #1096)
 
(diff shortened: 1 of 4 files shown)


penelopeysm:
Bugs like these I think strengthen the case for having a different IR that is neither CST nor FST. Oh well, maybe one year from now.

The point is that the CST (or at least JuliaSyntax's one) is not really the right data structure to use for making formatting decisions. We can do so, indeed the fact that JuliaFormatter v2 works is evidence that it can be done, but the logic needed to do so correctly is so complicated and finicky that it leads to many bugs.

It would be simpler if we had an IR that contained precisely the exact amount of information needed to make formatting choices, since that would let us disentangle the issues of "JuliaSyntax's CST is weird" from "formatting is difficult".

PR #1098, opened 2026-06-13
app: Clarify jlfmt check guidance for multiple inputs

aviatesk:
Before, the CLI suggested: `jlfmt --check src` -> "Some files are not formatted correctly. Run without `--check` to format them". Following that advice with a directory led to: `jlfmt src` -> "ERROR: option `--inplace` or `--check` required with multiple input files."

With this commit, `--check` failures now say: "Some files are not formatted correctly. Run again with `--inplace` instead of `--check` to format them". Running a directory without an action now says: "ERROR: multiple input files require either `--inplace` to write changes or `--check` to verify formatting".

### DIFF
--- a/HISTORY_v2.md
+++ b/HISTORY_v2.md
@@ -1,3 +1,7 @@
+# v2.7.0
+
+Improved usage messages for the `jlfmt` command-line tool. (#1098)
+
 # v2.6.15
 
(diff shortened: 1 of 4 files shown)


penelopeysm (approved):
Thanks @aviatesk, nice UX touch!

I'll just bump the version number here in anticipation of the other PRs you opened.

PR #1115, opened 2026-06-15
fix comments in array literals

### LINKED ISSUE #1113: comments in array literals are utterly messed up
```julia
julia> s = """
       X = [
           # foo
           1 2
           3 4
           # bar
       ]"""
"X = [\n    # foo\n    1 2\n    3 4\n    # bar\n]"

julia> ft(s) |> println
X = [

    1 2
    3 4

    # bar
]

julia> ft(ft(s)) |> println
X = [

    1 2
    3 4



    # bar
]
```

### LINKED ISSUE #1116: hcat idempotence
```
 Test failure.

  First pass:
  X = [

      # only
      1 2
  ]


  Second pass:
  X = [


      # only
      1 2
  ]


  Repro:
  s = "X = [\n    # only\n    1 2\n]\n"
  format_text(s)
```

penelopeysm:
Fixes #1113 
Fixes #1116

claude did this, and i will probably merge this after reviewing it because it's a really, really ugly issue and it should be fixed, but in reality this is a gigantic hack.

the right fix would be to do one or both of the following

1) restructure the CST so that the comments are placed at the top level of the vcat rather than inside the row.
2) completely rework the handling of line comments. right now they're not present in the FST and must be added back later via a notcode mechanism which means lots of moving parts that all interact with one another.

### DIFF
--- a/src/styles/default/pretty.jl
+++ b/src/styles/default/pretty.jl
@@ -3698,6 +3698,7 @@ function p_hcat(
     # to decide whether to place boundary newlines. In other words, we can safely insert
     # Placeholder nodes at these positions.
 
+    prev_comment_was_dropped = false
     for (i, a) in enumerate(childs)
         n = pretty(style, a, s, ctx, lineage)
         if kind(a) === K"["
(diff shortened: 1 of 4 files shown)


src/styles/default/pretty.jl
```diff
@@ -3862,6 +3864,25 @@ function is_semantically_important_newline(
     n = length(children(row_cst))
     # Start of the row - not important
     i == 1 && return false
+    # A newline is not important if it's adjacent to a comment and all the children
+    # between the comment and the boundary of the row (start or end) are whitespace.
```
penelopeysm, line 3892: like what does this even mean. it's incredibly non-obvious to a reader that this is the correct thing to do.

penelopeysm:
!formatbot JuliaLang/julia@v1.12.6 v2_stable_multiline_strings=true

penelopeysm:
!formatbot SciML/OrdinaryDiffEq.jl v2_stable_multiline_strings=true

PR #1117, opened 2026-06-15
Fix parenthesised caller bug

### LINKED ISSUE #1114: callable struct idempotence
```julia
julia> s = """
       function (foo::Foo)(a, b)
           foo
       end
       """
"function (foo::Foo)(a, b)\n    foo\nend\n"

julia> ft(s; margin=10) |> println
function (
    foo::Foo
)(
    a,
    b,
)
    foo
end


julia> ft(ft(s; margin=10); margin=10) |> println
function (
    foo::Foo
)
    (a, b)
    foo
end
```

penelopeysm:
Closes #1114 by not inserting newlines around the caller. This guarantees that even on Julia 1.12, your code won't break. ref:

https://github.com/JuliaLang/julia/issues/62124

https://github.com/JuliaLang/julia/issues/60558

### DIFF
--- a/HISTORY_v2.md
+++ b/HISTORY_v2.md
@@ -1,3 +1,8 @@
+# v2.8.2
+
+Fixed a bug where JuliaFormatter would insert newlines around a parenthesised caller in a function definition, causing the function to be parsed differently on Julia 1.12.
+This is probably a Julia bug and not really JuliaFormatter's fault, but this patch works around it. (#1114, #1117)
+
 # v2.8.1
 
 Fixed a bug causing line comments inside array literals to be dropped or otherwise cause non-idempotent formatting. (#1113, #1115, #1116)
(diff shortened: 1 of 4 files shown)


penelopeysm:
!formatbot JuliaLang/julia@v1.12.6 v2_stable_multiline_strings=true

penelopeysm:
!formatbot SciML/OrdinaryDiffEq.jl v2_stable_multiline_strings=true

penelopeysm:
It really shouldn't change anything, because if it did, then the output would have been non-idempotent, and we would have caught it awhile ago. Doesn't hurt to check I guess.

penelopeysm:
ok wtf

penelopeysm:
!formatbot JuliaLang/julia@v1.12.6 v2_stable_multiline_strings=true

penelopeysm:
!formatbot SciML/OrdinaryDiffEq.jl v2_stable_multiline_strings=true

PR #1120, opened 2026-06-15
Try regression tests with BlueStyle

penelopeysm:
Now that DefaultStyle has for the most part been fixed, it's time to start looking at other styles.

I know from running this locally that `BlueStyle` is broken on at least `JuliaLang/julia`. In particular, I'm pretty sure that `always_use_return` is broken. However, I'd like to see how broken it is across all our test repos.

~~NOT mergeable.~~


penelopeysm:
actually, i'll merge it so that i can stop rebasing it all the time.

PR #1128, opened 2026-06-16
Reimplement always_use_return at CST level + more fixes with short_circuit_to_if

### LINKED ISSUE #1125: Properly fix always_use_return for the sake of proper idempotence
its clear that many idempotence bugs with BlueStyle are to do with always_use_return

see e.g. https://github.com/JuliaEditorSupport/JuliaFormatter.jl/actions/runs/27587451750/job/81560852814?pr=1120

just fix it properly, as described in #1121

### LINKED ISSUE #1129: short_circuit_to_if doesn't recognise comments
```julia
julia> s = """
       function f(a, b)
           a && b
           #= A comment! Gosh I hate these bugs so much! =#
       end"""
"function f(a, b)\n    a && b\n    #= A comment! Gosh I hate these bugs so much! =#\nend"

julia> format_text(s; short_circuit_to_if=true) |> println
function f(a, b)
    if a
        b
    end #= A comment! Gosh I hate these bugs so much! =#
end
```

penelopeysm:
Same old motivation, basically, do not mutate the FST after it has been constructed, because bad for idempotence.

Closes #1125.

This inadvertently causes different behaviour with the combination of `always_use_return` + `short_circuit_to_if`. Previously, if both of these options were turned on, then

```julia
function f()
    a && b
```

would be converted to 

```julia
function f()
    return if a
        b
    else
        false
    end
end
```

With this PR, it is converted to 

```julia
function f()
    return a && b
end
```

This is, IMO, much more sensible anyway, and an improvement (see https://github.com/JuliaEditorSupport/JuliaFormatter.jl/issues/887#issuecomment-4723784027), so I've decided to not 'fix' that.

Closes #1129 as well while we're at it.


penelopeysm:
hmmmm ill take it though because that indicates this PR fixes a nice chunk of idempotence bugs

penelopeysm:
!formatbot JuliaLang/julia@v1.12.6 v2_stable_multiline_strings=true always_use_return=true

penelopeysm:
!formatbot SciML/OrdinaryDiffEq.jl v2_stable_multiline_strings=true always_use_return=true

penelopeysm:
!formatbot trixi-framework/Trixi.jl v2_stable_multiline_strings=true always_use_return=true

penelopeysm:
!formatbot oscar-system/Oscar.jl v2_stable_multiline_strings=true always_use_return=true

penelopeysm:
!formatbot JuliaLang/julia@v1.12.6 v2_stable_multiline_strings=true always_use_return=true

penelopeysm:
ahhhhhh damn the massive diff is because it fails to format julia on base. i thought formatbot was supposed to catch that.......

penelopeysm:
maybe its time to rework the format_text dispatch chain....

penelopeysm:
!formatbot trixi-framework/Trixi.jl v2_stable_multiline_strings=true always_use_return=true

penelopeysm:
!formatbot SciML/OrdinaryDiffEq.jl v2_stable_multiline_strings=true always_use_return=true

penelopeysm:
!formatbot MakieOrg/Makie.jl v2_stable_multiline_strings=true always_use_return=true

schillic:
As described in the release notes, this changeset leads to `return throw`s, which I find quite cumbersome; the proposal to add `return nothing` is even confusing, as it means adding unreachable code. Is there a low-hanging alternative to drop the `return` in such cases (probably with a new option)?

penelopeysm:
> Is there a low-hanging alternative to drop the return in such cases (probably with a new option)?

I'm afraid in v2 that won't happen. I wrote an explanation of this [in the docs](https://juliaeditorsupport.github.io/JuliaFormatter.jl/stable/formatting_options/#options-always-use-return). Basically, there are a couple of things here: firstly the technical objection (knowing that the identifier is `throw` requires information from a later stage of formatting, but you can't insert `return` at that point because formatting decisions have already been made based on the absence of `return`); and secondly, my personal objection to the idea that `return` shouldn't be added before `throw()`. There's a nonzero chance that I could be convinced by overwhelming community demand to drop the latter objection, but even so, the former is a blocker in v2.

penelopeysm:
(There are some other potential solutions to this, for example, making sure that formatting is independent of whether or not there is a `return` in front of it would sidestep the technical bits, but that's also not something I really want to change in v2 because broadly speaking I'm trying my best to adhere to the stylistic choices that were already present in v2, even if it's not my personally preferred style!)

schillic:
I had not seen that long explanation, sorry. Alright, thanks. Good to know that it will not be supported at least.

penelopeysm:
No apology needed! I didn't expect you to have seen it before this :) 

And it's just not for now, but maybe in a future version, or (slightly more ambitious, but also more correctly) I would actually like transformations like these to be done by a different tool rather than a formatter. I actually don't think a formatter should be changing the AST, and in this respect I think JuliaFormatter is committing many sins. But I could envision a different tool that does something like JET, to see if the return value of a function has type `Union{}` and if not, suggest adding a return.

PR #1137, opened 2026-06-18
Simplify options / config / format_text dispatch

penelopeysm:
I promise the line additions are mostly new tests...


penelopeysm:
So I made sure to add tests before doing this rework, and the tests all pass which gives me some confidence that there aren't any regressions in behaviour. However, Claude still found some complaints:

<details>

# Code Review Findings — `api` branch

Reviewed the full diff of `master...HEAD` (37 files, ~2400 lines changed).
Five finder angles (line-by-line, removed-behavior, cross-file, reuse/simplify, conventions)
followed by single-vote verification on top candidates.

## Confirmed Bugs

### 1. `find_config_file` crashes on nonexistent `--config-dir`

- **File**: `src/config.jl:85`
- **Severity**: Bug (crash)
- `find_config_file` calls `realpath(path)` which throws `SystemError` if the
  path doesn't exist. The CLI app calls `find_config_file(args.config_dir)`
  directly at `src/app.jl:397` with no existence check or try/catch.
- **Repro**: `jlfmt --config-dir=/nonexistent file.jl` → `SystemError`
- **Fix**: Check `isdir(path)` before `realpath`, or wrap in try/catch.

### 2. `--format_markdown` cannot be negated to override config file

- **File**: `src/argparse.jl:837`
- **Severity**: Design gap (not a regression — old code had the opposite bug of ignoring config files)
- `--format_markdown` is a flag (can only be present or absent). The encoding
  `get(raw, :format_markdown, false) ? true : nothing` maps both "not passed"
  and "false" to `nothing`. So if a `.JuliaFormatter.toml` sets
  `format_markdown = true`, the CLI cannot override it to `false`.
- **Fix**: Add a `--no-format-markdown` flag, or change it to an option
  (`--format-markdown=true|false`).

### 3. `FORMAT_OPTION_KEYS` is a manual duplicate of `fieldnames(Options)`

- **File**: `src/argparse.jl:266`
- **Severity**: Maintenance hazard
- `FORMAT_OPTION_KEYS` is `fieldnames(Options)` plus `:style`, manually listed.
  The loop at line 824 already skips `:style`. Adding a new field to `Options`
  without updating this set silently drops it from CLI parsing.
- **Fix**: Replace with `Set{Symbol}(fieldnames(Options))` (and handle `:style`
  via the PARSER specs, not via this set).

### 4. `options(::DefaultStyle)` duplicates `Options()` defaults

- **File**: `src/styles/default/pretty.jl:1`
- **Severity**: Maintenance hazard
- `options(::DefaultStyle)` manually lists every field with the same defaults as
  the `Options` struct definition. Adding a new field to `Options` without
  updating `options(::DefaultStyle)` causes DefaultStyle to silently use a
  different default.
- **Fix**: `options(::DefaultStyle) = Options()`.

## Confirmed but Lower Priority

### 5. `_format_file` uses `Union{Nothing,Bool}` fields in boolean context

- **File**: `src/format.jl:169`
- **Severity**: Latent bug (safe today, fragile for future callers)
- `config.verbose && println(...)`, `config.format_markdown || return true`, and
  `config.overwrite && !already_formatted` all throw `TypeError` if the field is
  `nothing`. Currently safe because `format()` always merges with
  `Configuration()` defaults first, but a future call site that skips the merge
  would crash.
- **Fix**: Use `something(config.verbose, false)` etc., or assert non-nothing at
  the top of `_format_file`.

### 6. Recursive directory formatting re-parses TOML per file

- **File**: `src/JuliaFormatter.jl:255`
- **Severity**: Performance (not correctness)
- `format(dir)` recursively calls `format(subpath, style; options...)` for each
  file, and each call independently runs `find_config_file` + TOML parsing.
  For a directory with 500 files and one config, the same TOML is parsed 500
  times.
- **Note**: This is pre-existing behavior (the old code did the same), but worth
  noting for future optimization.

## Refuted

### `valid_for_in_op` guard removal in `eq_to_in_normalization!`

- **File**: `src/fst.jl:932`
- Callers only invoke the function in for-iteration contexts, and the recursion
  pattern only descends into Block/Brackets/Filter — non-iteration operators
  are never reached. The guard was defense-in-depth; removing it is safe.

## Cleanup Suggestions (from Angle D)

### `merge_config` repetitive code

- **File**: `src/config.jl:117`
- Five repetitive `config2.X === nothing ? config1.X : config2.X` lines could be
  factored into a loop or helper, matching how `merge_options` works.

### Config resolution duplication between `format()` and `process_file()`

- **File**: `src/JuliaFormatter.jl:249` and `src/app.jl:392`
- Both do `find_config_file` → `configuration_from_file` → `merge_config` with
  defaults. Could be extracted into a shared `resolve_config(path, caller_config)`
  function in `config.jl`.

### `_format_text` docstring inaccuracy

- **File**: `src/format.jl:45`
- Docstring says `maxiters` defaults to 1, but the actual default is
  `_maxiters(style)` which returns 4 for `SciMLStyle`.

</details>

penelopeysm:
Okay, I've done a lot of local testing already and am happy that nothing is messed up. Once CI passes (excluding the known Blue failures) I'll merge and release.

PR #1141, opened 2026-06-18
Fix `separate_kwargs_with_semicolon` overcautiousness

### LINKED ISSUE #625: Misplaced semicolon in arbitrarily ordered keywords.
```julia
foo(x, bar=0, y)
foo(x, y, bar=0)
```

Becomes:

```julia
foo(x; bar=0, y)
foo(x, y; bar=0)
```

😭

### LINKED ISSUE #1133: short to long function def + semicolon to kwarg idempotence
https://github.com/MakieOrg/Makie.jl/blob/31de53c4643b41605b3a4e23ad68449e13de14de/Makie/src/basic_recipes/contourf.jl#L53

```julia
julia> s = "_get_isoband_levels(levels::Int, mi, ma) = collect(range(Float32(mi), nextfloat(Float32(ma)), length = levels + 1))"
"_get_isoband_levels(levels::Int, mi, ma) = collect(range(Float32(mi), nextfloat(Float32(ma)), length = levels + 1))"

julia> test_format(s, nothing, BlueStyle())
Formatting was not idempotent.

First pass:
function _get_isoband_levels(levels::Int, mi, ma)
    return collect(range(Float32(mi), nextfloat(Float32(ma)), length=levels + 1))
end

Second pass:
function _get_isoband_levels(levels::Int, mi, ma)
    return collect(range(Float32(mi), nextfloat(Float32(ma)); length=levels + 1))
end

Repro:
s = "_get_isoband_levels(levels::Int, mi, ma) = collect(range(Float32(mi), nextfloat(Float32(ma)), length = levels + 1))"
format_text(s, BlueStyle())
Test Failed at /Users/pysm/jl/jf/src/internal/utils.jl:221
  Expression: out2 == out
   Evaluated: "function _get_isoband_levels(levels::Int, mi, ma)\n    return collect(range(Float32(mi), nextfloat(Float32(ma)); length=levels + 1))\nend" == "function _get_isoband_levels(levels::Int, mi, ma)\n    return collect(range(Float32(mi), nextfloat(Float32(ma)), length=levels + 1))\nend"

ERROR: There was an error during testing
```

### LINKED ISSUE #1140: `separate_kwargs_with_semicolon` doesn't fire even when it probably should
```julia
julia> s3 = """function foo(a, b=goo(x, y=z))
           body
       end"""

"function foo(a, b=goo(x, y=z))\n    body\nend"

julia> format_text(s3, BlueStyle()) |> print
function foo(a, b=goo(x, y=z))
    return body
end
```

One would expect that the inner call `goo(x; y=z)` is transformed, but not the outer one. This is actually an overly eager fix to #494 -- that fix caused the transformation to be disabled for `foo(a, b)` which is correct, but it also propagated to its children, which is not necessary.

penelopeysm:
Closes #625
Closes #1140
Closes #1133 

~~Needs tests.~~ Tests written.


penelopeysm:
Reminder to self that I also need to add tests for anonymous functions, which go through the same code paths.

PR #1176, opened 2026-06-22
Disable pipe_to_function_call in CI

penelopeysm:


penelopeysm:
Okay, that did indeed fix a lot of idempotence issues.

Issue #1203, opened 2026-07-20
Don't triple-quote short docstrings.


iago-lito:
For unexposed, internal, private functions, I am sometimes very happy with:
```julia
"Frob the snats."
function frsn end

"Drit the brogs."
function drbr end

"Deregal the shotapins."
function dersh end
```
But julia formatter turns the above into:
```julia
"""
Frob the snats.
"""
function frsn end

"""
Drit the brogs.
"""
function drbr end

"""
Deregal the shotapins.
"""
function dersh end
```
This introduces too much useless vertical noise as it basically doubles the size of each item.
Would it be an option to request `enforce_triplequote_docstring = false`?

penelopeysm:
You can use `format_docstrings = false`, at the cost of not formatting code inside docstrings (is that very important?)

iago-lito:
Err, it is, because the docstrings for exposed functions within the same project are much more fleshed and do include example snippets as doctests ^ ^"

I mean: I *would* keep `format_docstrings = true`.

penelopeysm:
Okay. I'm not super duper enthusiastic about adding more options in general but this one would probably be fairly self contained in `src/format_docstring.jl`, so could see it happening

penelopeysm:
Closed by #1204

PR #1204, opened 2026-07-20
Opting out of `enforce_triplequote_docstring`.

iago-lito:
Here is a first shot at addressing #1203 it @penelopeysm.
I am happy with it yet except that I don't understand why [this (new) test](https://github.com/JuliaEditorSupport/JuliaFormatter.jl/pull/1204#discussion_r3613772816) is failing..?
It seems that it is failing due to something outside the strict `format_docstring()` function, because this one does return the correct result IIUC.
Maybe that leading whitespace is preventing the docstring to be recognized as such?

### DIFF
--- a/src/format_docstring.jl
+++ b/src/format_docstring.jl
@@ -49,14 +49,10 @@ function block_modifier(rule::FormatRule)
 end
 
 function format_docstring(style::AbstractStyle, state::State, text::AbstractString)
-    state_indent = state.indent
+    is_triple_quoted =
+        state.opts.enforce_triplequote_docstring ||
+        (startswith(text, "\"\"\"") && endswith(text, "\"\"\""))
     start_boundary = findfirst(!=('"'), text)
-    is_triple_quoted = state.opts.enforce_triplequote_docstring || let
-        # Assuming well-formedness, the string is triple-quoted iif
-        # the chars after/before (o)pening/(c)losing quotes are also quotes.
-        o, c = map(f -> f(==('"'), text), (findfirst, findlast))
-        text[o+1] == text[c-1] == '"'
-    end
     # if the docstring is non-empty
     if !isnothing(start_boundary)
         _end_boundary = findlast(!=('"'), text)
@@ -127,22 +123,22 @@ function format_docstring(style::AbstractStyle, state::State, text::AbstractStri
     end
     # Render into text lines, taking care of original indentation,
     quot = is_triple_quoted ? "\"\"\"" : '"'
-    indentation = " "^state_indent
+    indentation = " "^state.indent
     indent(line) = indentation * line
     clean(line) = all(isspace, line) ? "" : line # don't write empty lines #667
-    lines = split(formatted, '\n') # Always contains at least an empty last line.
+    lines = split(formatted, '\n')
+    last(lines) == "" || # Legacy guarantee.
+        error("unreachable: `formatted` should end in empty string. Please report this at \
+               https://github.com/JuliaEditorSupport/JuliaFormatter.jl/issues!")
     if is_triple_quoted
         prep = line -> clean(indent(line)) # All lines are prepared the same way.
         lines = Iterators.map(prep, lines)
         quot * '\n' * join(lines, '\n') * indent(quot)
     else
         # The first line needs no indentation.
-        lines = Iterators.Stateful(lines[1:end-1]) # (drop last empty line)
-        first = popfirst!(lines) |> clean
-        isempty(lines) && return quot * first * quot
-        # Upcoming ones do.
+        first = clean(lines[1])
         prep = line -> '\n' * indent(line)
-        lines = Iterators.map(prep, lines)
+        lines = Iterators.map(prep, lines[2:(end-1)]) # (drop last empty line)
         quot * first * join(lines) * quot
     end
 end
--- a/src/format_docstring.jl
+++ b/src/format_docstring.jl
@@ -51,7 +51,7 @@ end
 function format_docstring(style::AbstractStyle, state::State, text::AbstractString)
     state_indent = state.indent
     start_boundary = findfirst(!=('"'), text)
-    is_triple_quoted = let
+    is_triple_quoted = state.opts.enforce_triplequote_docstring || let
         # Assuming well-formedness, the string is triple-quoted iif
         # the chars after/before (o)pening/(c)losing quotes are also quotes.
         o, c = map(f -> f(==('"'), text), (findfirst, findlast))
--- a/src/format_docstring.jl
+++ b/src/format_docstring.jl
@@ -51,6 +51,12 @@ end
 function format_docstring(style::AbstractStyle, state::State, text::AbstractString)
     state_indent = state.indent
     start_boundary = findfirst(!=('"'), text)
+    is_triple_quoted = let
+        # Assuming well-formedness, the string is triple-quoted iif
+        # the chars after/before (o)pening/(c)losing quotes are also quotes.
+        o, c = map(f -> f(==('"'), text), (findfirst, findlast))
+        text[o+1] == text[c-1] == '"'
+    end
     # if the docstring is non-empty
     if !isnothing(start_boundary)
         _end_boundary = findlast(!=('"'), text)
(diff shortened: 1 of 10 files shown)


test/options.jl
```diff
@@ -711,6 +711,65 @@ end
...
+                "$("  ")
+                doc
+                "
+                f() = 0
+            end
+            """,
+            """
+            begin
+                "doc"   # <- This docstring is correctly formatted.
+                f() = 0 # Can somebody help explain why newlines are being inserted here?
```
iago-lito: Here, can you help me explain?

test/options.jl
```diff
@@ -711,6 +711,65 @@ end
         end
     end
 
+    @testset "issue 1203" begin
```
penelopeysm: Could I trouble you to move this testset to `test/options/enforce_triplequote_docstring.jl` please? I think the options file is too big. I haven't split the file all up yet (it's a boring job, I suppose I could get an LLM to do it... 😅) but I'd like for new tests to start using the new system.

src/format_docstring.jl
```diff
@@ -51,6 +51,12 @@ end
 function format_docstring(style::AbstractStyle, state::State, text::AbstractString)
     state_indent = state.indent
     start_boundary = findfirst(!=('"'), text)
+    is_triple_quoted = state.opts.enforce_triplequote_docstring || let
+        # Assuming well-formedness, the string is triple-quoted iif
+        # the chars after/before (o)pening/(c)losing quotes are also quotes.
+        o, c = map(f -> f(==('"'), text), (findfirst, findlast))
+        text[o+1] == text[c-1] == '"'
+    end
```
penelopeysm: You could just check `startswith(text, "\"\"\"") && endswith(text, "\"\"\"")`, which would be clearer in terms of intent. It also avoids a bug where an empty single-quoted string tests positive:

```julia
julia> text = "\"\""
"\"\""

julia> o, c = map(f -> f(==('"'), text), (findfirst, findlast))
(1, 2)

julia> text[o+1] == text[c-1] == '"'
true
```

test/options.jl
```diff
@@ -711,6 +711,65 @@ end
...
+            f() = 0
+        end
+        """
+
+        # Normalize to triple by default.
+        test_format(triple, triple; opts...)
+        test_format(single, triple; opts...)
+
+        # Unless we opt out.
+        opts = (; enforce_triplequote_docstring = false, opts...)
```
penelopeysm:
```suggestion
        opts = (; format_docstring = true, enforce_triplequote_docstring = false)
```

Reduces the chance of messing things up if code is shifted around

test/options.jl
penelopeysm: Yeah, this is a slightly ugly consequence of the way JuliaFormatter tracks non-code lines, such as comments or whitespace. Essentially it "remembers" which line expressions are on (in this case `f() = 0` is on line 5), and when the three lines of the docstring are contracted to one line, it tries to put the `f() = 0` back on the usual line. The docstring `"doc"` now ends on line 2 so you get lines 3-4 being empty.

I'm not a massive fan of this but I think it's a bit of a historical artifact and I'm hesitant to change it because it would break everything ... Anyway you can work around this by changing line 936 of `src/styles/default/pretty.jl` as follows:

```diff
         elseif i == length(childs)
-            add_node!(t, pretty(style, c, s, ctx, lineage), s; max_padding = 0)
+            end_node = pretty(style, c, s, ctx, lineage)
+            end_node.startline = t.endline
+            add_node!(t, end_node, s; max_padding = 0)
         else
             add_node!(t, pretty(style, c, s, ctx, lineage), s)
         end
```

Here `end_node` is the expression corresponding to `f() = 0`. We're basically patching it to claim that it begins at the same line that the thing before it ends at. `add_node!` by default adds a newline between expressions, so that guarantees that there will be one newline between docstring and function.

(If interested, & for my own reference, this is the same hacky technique I used to fix https://github.com/JuliaEditorSupport/JuliaFormatter.jl/issues/1153)

src/format_docstring.jl
```diff
@@ -119,21 +125,24 @@ function format_docstring(style::AbstractStyle, state::State, text::AbstractStri
...
+        quot * '\n' * join(lines, '\n') * indent(quot)
+    else
+        # The first line needs no indentation.
+        lines = Iterators.Stateful(lines[1:end-1]) # (drop last empty line)
+        first = popfirst!(lines) |> clean
+        isempty(lines) && return quot * first * quot
+        # Upcoming ones do.
+        prep = line -> '\n' * indent(line)
+        lines = Iterators.map(prep, lines)
+        quot * first * join(lines) * quot
```
penelopeysm, line 142:
```suggestion
        first = clean(lines[1])
        prep = line -> '\n' * indent(line)
        lines = Iterators.map(prep, lines[2:end - 1])
        quot * first * join(lines) * quot
```

I'm fairly sure this is equivalent, it passes tests.

src/format_docstring.jl
```diff
@@ -119,21 +125,24 @@ function format_docstring(style::AbstractStyle, state::State, text::AbstractStri
...
-        if !all(isspace, line)
-            write(buf, indent)
-            write(buf, line)
-        end
+    # Render into text lines, taking care of original indentation,
+    quot = is_triple_quoted ? "\"\"\"" : '"'
+    indentation = " "^state_indent
+    indent(line) = indentation * line
+    clean(line) = all(isspace, line) ? "" : line # don't write empty lines #667
+    lines = split(formatted, '\n') # Always contains at least an empty last line.
```
penelopeysm:
```suggestion
    lines = split(formatted, '\n') # Always contains at least an empty last line.
    if lines[end] != ""
        error("unreachable: `formatted` should end in empty string. Please report this at https://github.com/JuliaEditorSupport/JuliaFormatter.jl/issues!")
    end
```

I know this assertion predates this PR so not a criticism of you at all, but I'll sleep more easily at night if there's an obvious error message

penelopeysm (commented):
Thanks for looking into it!

penelopeysm:
Also, I'm sure you noticed this too but the argument needs to be added to the `jlfmt` CLI app at `src/argparse.jl`:

```julia
    option(
        "--enforce-triplequote-docstring";
        ...
    ),
```

src/format_docstring.jl
penelopeysm: Actually, come to think of it I wonder how much this breaks for things like `raw"""blah"""`. I think that would probably be orthogonal to this PR though

iago-lito: Didn't think about `""` indeed ^ ^".
I suspected that `startswith` would yield wrong results thus the former use of `first(!=('"'), ..)`? What if the docstring is *prefixed* like so?
```julia
prefix"mydocstring"
f() = 0
```

src/format_docstring.jl
iago-lito: Yeah, this felt like a weird 'guarantee' to have. I just inherited it but better check.

src/format_docstring.jl
iago-lito: Wops sorry I skipped over your second comment for some reason. Okay so I'll just assume strings always start with `"` until another PR generalizes that.

penelopeysm: Haha no worries! Maybe we can stick with the current approach but just check that it's not exactly `""`?

test/options.jl
iago-lito: Ew, I'll just trust you on this one X) Although.. does this trigger a bug whenever formatting changes the vertical size of the docstring?

test/options.jl
iago-lito: Sure, better this way :)

iago-lito:
Thank you for feedback @penelopeysm. Ready for the next round :)

test/options.jl
penelopeysm: I'm fairly sure it should be fine. The tests cover both cases I believe :)

penelopeysm (commented):
This looks really good! I'll just run a regression test and if it doesn't make anything stupendously ugly we should be fine.

penelopeysm:
!formatbot JuliaLang/julia@v1.12.6 --v2-stable-multiline-strings=true --format-docstrings=true

penelopeysm:
ah darn I haven't made it use the cli app yet

penelopeysm:
!formatbot JuliaLang/julia@v1.12.6 v2_stable_multiline_strings=true format_docstrings=true

penelopeysm:
Okay, that's pretty annoying. Base Julia contains docstrings that contain invalid Julia code, so it can't format them. That's a separate error worth fixing, but I think I'll just trust the tests for now.

Finally, just some bikeshedding: I might do a slight rename to `enforce_triplequoted_docstrings` if that's ok.

iago-lito:
Totally ok. Thank you for support <3

penelopeysm:
Fab! Thank you so much for taking this on!

iago-lito:
No worries. It did help a lot that you wrote which file to work on and that the changes would most likely be contained there :)

What's the usual release schedule?

penelopeysm:
Usually instantly, but there are a couple things I want to fix and bundle into a release, so hopefully later today.

penelopeysm:
v2.11.0 is released now :)

src/format_docstring.jl
penelopeysm: > What if the docstring is prefixed like so?

Only unprefixed docstrings are recognised, apparently:

````julia
julia> raw"""
       Hello
       """
       function f end

       """
       World
       """
       function g end
g

help?> f
search: f fd if

  No documentation found for private binding Main.f.

  f is a Function.

  # 0 methods for generic function "f" from Main

help?> g
search: g

  World
````

iago-lito: Yeah, I've noticed that. I actually use it to escape from formatting here and there ;) Although it's not meant to be used that way.

PR #1207, opened 2026-07-21
Avoid formatting invalid code in docstrings

### LINKED ISSUE #1206: `format_docstring` fails if any Julia code blocks are unparseable
This is quite annoying, because docstrings often contain code *examples* that are not meant to be literally run (but which still use Julia code blocks for e.g. syntax highlighting). For example, something like this will fail to format:

````julia
"""
    foo(a::MySpecialArg)

Do foo to a special arg. Example:

```julia
a = MySpecialArg(...) # set up your arg
foo(a)
```
"""
foo(a::MySpecialArg) = a
````

In such cases, JuliaFormatter should probably just output the text as-is.

penelopeysm:
Closes #1206


penelopeysm:
ok its a tad annoying that the formatting changes got mixed up in this pr but ig it doens't matter

penelopeysm:
OK, well, tests are failing, but that's actually good since it's flagging a format_docstrings bug whereas previously this was just completely untested. I'll merge this.

PR #1209, opened 2026-07-21
Make FormatBot use jlfmt + convert to local script

penelopeysm:
Clauded.

This turns FormatBot into a local Python script which can be run without needing to use a github runner. The workflow is changed to use the Python script.

The script works correctly when run locally, but I need to test the workflow.

Also docs haven't been updated.


penelopeysm:
!formatbot JuliaLang/julia@v1.12.6 --ignore-config --style=blue --v2-stable-multiline-strings=true

.github/workflows/formatbot-refactor.md
penelopeysm, line 1: delete this file.....

penelopeysm:
oh i forgot gha has to be merged first sigh

PR #1213, opened 2026-07-21
Emit info message when passing unused kwargs

### LINKED ISSUE #1205: warn if kwargs to format are unused
Because `format_text` etc etc take `kwargs...` it's very easy to mistype an option and think that it's doing nothing. We should warn if there are any unused kwargs.

penelopeysm:
Closes #1205


penelopeysm:
Not sure if it should be info or warn...

penelopeysm:
Claude says info which lines up with the PR I did so I'll take it

PR #1218, opened 2026-07-22
Add YASStyle to integration tests

penelopeysm:
YOLO


penelopeysm:
I'm certain that this will break but I'm mostly curious as to how bad the (lack of) idempotence is.

penelopeysm:
Eek, it doesn't even format on first try. That's no good, but at least it tells us what to fix.

PR #1228, opened 2026-07-24
Handle escapes in docstrings more correctly

### LINKED ISSUE #1224: Escape sequences in julia doctests are mangled
````julia
s = raw"""
\"""
```jldoctest
julia> @macro a.\$s
[...]
```
\"""
f
"""

ft(s; format_docstrings=true) |> println
````

gives

````julia
"""
```jldoctest
julia> a .\ $s
[...]
```
"""
f

````

which is no longer valid.

Frankly, I don't really know how to fix this. The input Julia is not really valid to begin with: it shouldn't have a backslash. The backslash is needed to avoid interpolating the actual value of `s` into the docstring (i.e., it's an escaped interpolation). But we don't know that......

penelopeysm:
Closes #1224, ~~but I'm not sure if this is enough? Particularly concerned about escaped quotes~~

Quotes are handled now

````julia
julia> s = raw"""
       \"""
       ```jldoctest
       julia> @macro a.\$s
       [...]
       ```
       \"""
       f
       """
"\"\"\"\n```jldoctest\njulia> @macro a.\\\$s\n[...]\n```\n\"\"\"\nf\n"

julia> ft(s; format_docstrings=true) == s
true

````

### DIFF
--- a/HISTORY_v2.md
+++ b/HISTORY_v2.md
@@ -1,3 +1,7 @@
+# v2.11.3
+
+Fixed a bug where Julia code in docstrings would be formatted incorrectly if the code itself contained characters that had to be escaped. (#1224, #1228)
+
 # v2.11.2
 
 Fixed a bug where YASStyle would insert a trailing comma for a function call with zero arguments, which is invalid Julia syntax. (#1225, #1226)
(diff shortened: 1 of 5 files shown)


penelopeysm:
I don't want to claim that this is perfect because I don't know if I missed something out, but it's definitely an improvement.

PR #1231, opened 2026-07-25
fix more indents

### LINKED ISSUE #1230: Indentation of multi-line vector inside macro
Since one of the recent releases (from 18.7.; I did not bisect), the code block

```
for x in @mymacro [1111111111111111111111, 222222222222222222222222,
                   33333333333333333333, 444444444]
    # !
end
```

gets formatted (with YASStyle) to

```
for x in @mymacro [1111111111111111111111, 222222222222222222222222,
          33333333333333333333, 444444444]
    # !
end
```

It seems that the formatter is ignoring the `@mymacro` characters for the indentation (but that may be a coincidence).
Was that change intended?

penelopeysm:
Closes #1230

Claude wrote this

<details><summary>Claude spam</summary>

# YAS: misaligned continuation lines after un-nesting a binary op

## Symptom

```julia
julia> s = """
       for x in @mymacro [1111111111111111111111, 222222222222222222222222,
                          33333333333333333333, 444444444]
           foo
       end""";

julia> format_text(s, YASStyle()) |> println
for x in @mymacro [1111111111111111111111, 222222222222222222222222,
          33333333333333333333, 444444444]
    foo
end
```

`33333333333333333333` should line up with `1111111111111111111111` at column 19. It comes
out at column 10 instead.

Two other right-hand sides are affected the same way — a unary operator applied to a
bracketed expression, and a `where`:

```julia
for x in !(ccccccccccccccccccccccc == 1 &&
  (ddddddddddddddddddd(eee[1]) || ffffffffffffffffffffffffffffff))     # want column 11
    foo
end

for x in Foo{A} where {Aaaaaaaaaa<:Xxxxxxxxxxxx,Bbbbbbbbbb<:Yyyyyyyyyyyy,
              Ccccccccccccccccccc<:Zzzzzzzzzzzz}                       # want column 23
    foo
end
```

## The mechanism

The relevant code is `n_binaryopcall!` in `src/styles/default/nest.jl`. The bug is in how
it *undoes* a nesting decision it made speculatively.

To lay out `x in @mymacro [111…, 222…, 333…, 444…]` at margin 92:

1. The whole thing is 97 wide, so it nests at the operator, putting the RHS on its own
   line. `in` is not an assignment, so `indent_nest` is false and the RHS is placed at
   `fst.indent` — **column 4**, the loop-body indent.

2. The RHS is nested *there*. A `MacroBlock` has no nest function of its own, so it hits
   the generic fallback, which walks `@`(1) + `mymacro`(7) + ` `(1) and nests the `Vect` at
   column 13. `n_tuple!` then sets `Vect.indent = 13 + 1 = 14` — align just past the `[`.

3. The LHS is nested, and the code re-measures: does the RHS's *first line* fit after
   `for x in `? `9 + 59 + 1 = 69 ≤ 92`, so yes — the nesting was unnecessary and gets
   undone.

4. `fst[i1] = Whitespace(1)`, `fst[i2] = Whitespace(0)`, and the RHS goes back on the same
   line, now starting at **column 9** rather than 4.

The RHS has moved sideways by +5, but everything inside it was aligned for column 4.

## The undo is a heuristic

Step 4 calls `walk(unnest!(style; dedent = true), rhs, s)`, and `dedent!`
(`src/nest_utils.jl`) simply subtracts one indent level from every non-leaf:

```julia
function dedent!(fst::FST, s::State)
    if is_closer(fst) || fst.typ === NOTCODE
        fst.indent -= s.opts.indent
    elseif is_leaf(fst) || fst.typ === StringN
        return
    else
        fst.indent -= s.opts.indent
    end
end
```

`Vect.indent` goes 14 → 10.

That is the right move for constructs indented *from the start of the line*: a block body
genuinely shifts left by one level when a level of nesting disappears. It is meaningless
for content aligned against a bracket, which has to move by however far the RHS actually
moved — here +5, to column 19 — not by −4.

## The repair step, and why it only worked sometimes

The old code knew this and patched it up afterwards, but only for two node types:

```julia
walk(unnest!(style; dedent = true), rhs, s)
# Iterables in YASStyle need to be aligned with the
# open bracket
if style isa YASStyle
    if is_unnamed_iterable(rhs)
        extra_indent = if !isempty(rhs.nodes) && is_opener(rhs[1])
            line_offset - rhs.indent + 1
        else
            line_offset - rhs.indent
        end
        add_indent!(rhs, s, extra_indent)
    elseif is_named_iterable(rhs)
        extra_indent =
            line_offset - rhs.indent + length(rhs[1]) + length(rhs[2])
        add_indent!(rhs, s, extra_indent)
    end
end
```

`is_unnamed_iterable` covers `TupleN, Vect, Vcat, Ncat, Braces, Comprehension, Brackets`;
`is_named_iterable` covers `Call, Curly, TypedComprehension, MacroCall, RefN, TypedVcat,
TypedNcat`. A `MacroBlock` is in neither list, so no branch fired and `Vect.indent` kept
the raw dedent value of 10. Likewise for `Unary` and `Where`.

## Why it had to be written per-type

The two formulas look different but are the same computation. Write `rhs_start` for the
column the RHS occupied in the nested layout:

- **unnamed iterable** — `n_tuple!` had set `rhs.indent = rhs_start + 1`, so after the
  dedent it is `rhs_start + 1 - indent`, and

  ```
  line_offset - rhs.indent + 1
      = line_offset - (rhs_start + 1 - indent) + 1
      = line_offset - rhs_start + indent
  ```

- **named iterable** — `n_call!` had set `rhs.indent = rhs_start + len(name) + 1` when it
  reached the opener, and `length(rhs[1]) + length(rhs[2])` *is* `len(name) + 1`, so

  ```
  line_offset - rhs.indent + length(rhs[1]) + length(rhs[2])
      = line_offset - (rhs_start + len(name) + 1 - indent) + len(name) + 1
      = line_offset - rhs_start + indent
  ```

Both reduce to `line_offset - rhs_start + s.opts.indent`: *how far the RHS moved, plus
undo the dedent*.

Neither branch had `rhs_start` to hand, so each one reconstructed it backwards out of
`rhs.indent` — which only works if you already know exactly how that node type's nest
function assigned `indent`. That is the whole reason there had to be one branch per type,
and the reason for the `+ 1` and `+ length(rhs[1]) + length(rhs[2])` fudge factors: they
are the offset from the node's start column to its alignment column, cancelling out the
same offset baked into `rhs.indent`.

So the bug was structural rather than an oversight about one node type. Any RHS whose nest
function sets `indent` by a rule not enumerated in those two branches gets no repair at all
and silently keeps the dedent.

## The fix

Record `rhs_start` directly, captured immediately before the RHS is nested:

```julia
# rhs
#
# Remember the column the RHS starts at in this (nested) layout. Any alignment
# that nesting the RHS sets up is relative to this column, so if the nesting is
# undone further below we need to know how far the RHS has moved sideways.
rhs_line_offset = s.line_offset
fst[end].extra_margin = fst.extra_margin
nest!(style, fst[end], s, lineage)
```

and then apply the single shift:

```julia
if style isa YASStyle && is_column_aligned(rhs)
    add_indent!(rhs, s, line_offset - rhs_line_offset + s.opts.indent)
end
```

No reverse-engineering, and the per-type fudge factors disappear. The existing iterable
cases are unchanged, since the old formulas were algebraically equal to this one.

What remains is the one question that genuinely *is* per-type, and which the old code was
conflating with the arithmetic: **does this node align to its own starting column at all,
or to the start of the line?** That is `is_column_aligned` (`src/fst.jl`):

```julia
function is_column_aligned(fst::FST)
    if is_leaf(fst)
        return false
    elseif fst.typ === MacroBlock
        # The trailing argument is the one that continuation lines follow on from.
        nodes = fst.nodes::Vector{FST}
        return !isempty(nodes) && is_column_aligned(nodes[end])
    elseif fst.typ === Unary
        # Either `!x` or `x...`; in both cases it's the operand that matters, not the
        # operator.
        nodes = fst.nodes::Vector{FST}
        idx = findfirst(n -> n.typ !== OPERATOR, nodes)
        return idx !== nothing && is_column_aligned(nodes[idx])
    end
    return fst.typ === Where || is_iterable(fst)
end
```

The distinction is real, not defensive: a first attempt applied the shift unconditionally
and broke the existing `ccc => function () … end` test, because a long-form function body
*is* line-anchored and the plain dedent was already correct for it. For the same reason a
`MacroBlock` delegates to its final argument — `@testset "x" begin … end` must not move,
`@mymacro [a, b]` must.

## Which node types can be affected

Only right-hand sides that (a) reach the un-nest path while still spanning several lines
and (b) are column-aligned.

`Binary`, `Chain`, `Comparison` and `Conditional` are column-aligned in YAS but cannot
qualify: `n_binaryopcall!` adds their **full** width to `line_margin` before deciding to
un-nest, so they only ever un-nest when they have collapsed onto a single line, at which
point there is no continuation line to misplace.

```julia
if (rhs.typ === Binary && !(op_kind(rhs) in KSet"in ::")) ||
   rhs.typ === Unary && rhs[end].typ !== Brackets ||
   rhs.typ === Chain ||
   rhs.typ === Comparison ||
   rhs.typ === Conditional
    line_margin += length(fst[end])
elseif ...
else
    # Calculate how much of the RHS we need to fit on the current line
    rw, _ = length_to(fst, (NEWLINE,); start = i2 + 1)
    line_margin += rw
end
```

Note that `Unary` is excluded from that clause exactly when its operand is a `Brackets` —
which is how the unary case became reachable.

Sweeping every `.jl` file in the repo with instrumentation on the un-nest path turned up
`MacroBlock`, `Unary` and `Where` as the reachable cases (plus `StringN`, which `unnest!`
skips, and `Begin`, which is correctly line-anchored).

## An unrelated pre-existing bug found along the way

At tight margins the output is not idempotent — but this predates the fix and is not
specific to it. A plain `Vect` right-hand side, which the old code already re-aligned,
breaks the same way:

```julia
julia> s = "for x in [aaaa, bbbb, cccc]\n    foo\nend";

julia> format_text(s, YASStyle(); margin=22) |> println
for x in [aaaa, bbbb,
          cccc]
    foo
end

julia> format_text(format_text(s, YASStyle(); margin=22), YASStyle(); margin=22) |> println
for x in
    [aaaa, bbbb,
     cccc]
    foo
end
```

On the second pass the source already contains the line break, which changes how the RHS
is measured when deciding whether to un-nest. The regression tests were placed at margins
where the output is idempotent (`test_format` checks this) rather than working around it.

</details>

### DIFF
--- a/src/fst.jl
+++ b/src/fst.jl
@@ -512,6 +512,74 @@ function is_import_expr(x::FST)
 end
 
 """
+    is_column_aligned(fst::FST)
+
+Whether YASStyle lays out the continuation lines of `fst` relative to the column at which
+`fst` itself starts, rather than relative to the start of the line it appears on.
+
+Iterables are column-aligned, because their arguments line up with the bracket that opens
+them, as is `where`, whose type parameters line up with the `{`:
+
+```julia
+f(aaaaaaaaaaaaaaaaaaaaaa,
+  bbbbbbbbbbbbbbbbbbbbbb)
+
+Foo{A} where {Aaaaaaaaaaaaaaa<:Xxxxxxxxxxxx,
+              Bbbbbbbbbbbbbbb<:Yyyyyyyyyyyy}
+```
+
+Block-like constructs are not, because their bodies are indented by one level from the
+start of the line:
+
+```julia
+function f()
+    body
+end
+```
+
+A unary operator takes this from its operand, and a macrocall written without parentheses
+(a `MacroBlock`) from its final argument, since those are the parts that the continuation
+lines belong to:
+
+```julia
+!(aaaaaaaaaaaaaaaaaaaaaa &&
+  bbbbbbbbbbbbbbbbbbbbbb)
+
+@mymacro [aaaaaaaaaaaaaaaaaaaa,
+          bbbbbbbbbbbbbbbbbbbb]
+
+@mymacro function f()
+    body
+end
+```
+
+Note that a `MacroBlock` that mixes the two (e.g. `@mymacro [a, b] begin ... end`) is
+reported as not column-aligned: the two parts would have to move in different ways, and
+only one answer can be given.
+
+`Binary`, `Chain`, `Comparison` and `Conditional` are column-aligned in YASStyle too, but
+they are deliberately not listed here because the only caller (`n_binaryopcall!`) accounts
+for their full width before it will consider undoing a nesting, so it never has to move
+one of them that has already been split over several lines.
+"""
+function is_column_aligned(fst::FST)
+    if is_leaf(fst)
+        return false
+    elseif fst.typ === MacroBlock
+        # The trailing argument is the one that continuation lines follow on from.
+        nodes = fst.nodes::Vector{FST}
+        return !isempty(nodes) && is_column_aligned(nodes[end])
+    elseif fst.typ === Unary
+        # Either `!x` or `x...`; in both cases it's the operand that matters, not the
+        # operator.
+        nodes = fst.nodes::Vector{FST}
+        idx = findfirst(n -> n.typ !== OPERATOR, nodes)
+        return idx !== nothing && is_column_aligned(nodes[idx])
+    end
+    return fst.typ === Where || is_iterable(fst)
+end
+
+"""
 Returns whether `fst` can be an iterable argument. For example in
 the case of a function call, which is of type `Call`:
 
--- a/src/fst.jl
+++ b/src/fst.jl
@@ -512,6 +512,35 @@ function is_import_expr(x::FST)
 end
 
 """
+    is_indent_nest(fst::FST)
+
+Whether nesting the binary call `fst` at its operator indents the RHS by one level from the
+start of the line:
+
+```julia
+aaaaaaaaaaaa =>
+    rhs
+```
+
+as opposed to bringing it back to the column the LHS starts at, which is what every other
+operator does:
+
+```julia
+aaaaaaaaaaaa in
+rhs
+```
+
+True for assignments, `=>`, `->` and standalone short-circuits.
+"""
+function is_indent_nest(fst::FST)
+    md = fst.metadata
+    return (!isnothing(md) && (md::Metadata).is_short_form_function) ||
+           is_assignment(fst) ||
+           op_kind(fst) in KSet"=> ->" ||
+           (!isnothing(md) && (md::Metadata).is_standalone_shortcircuit)
+end
+
+"""
     is_column_aligned(fst::FST)
 
 Whether YASStyle lays out the continuation lines of `fst` relative to the column at which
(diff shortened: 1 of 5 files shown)


src/styles/default/nest.jl
```diff
@@ -944,6 +944,11 @@ function n_binaryopcall!(
         end
 
         # rhs
+        #
+        # Remember the column the RHS starts at in this (nested) layout. Any alignment
+        # that nesting the RHS sets up is relative to this column, so if the nesting is
+        # undone further below we need to know how far the RHS has moved sideways.
+        rhs_line_offset = s.line_offset
```
penelopeysm: rename to `rhs_line_offset_in_nested_op`

penelopeysm:
This was Opus 5. So, honestly, glancing through this I think that the fix is sound i.e. it's fundamentally the right fix and isn't just papering over the cracks. I will review it carefully but I'm pretty impressed by it because I've seen LLMs generate terrible fixes for JuliaFormatter plenty of times now

src/fst.jl
```diff
@@ -511,6 +511,74 @@ function is_import_expr(x::FST)
...
+reported as not column-aligned: the two parts would have to move in different ways, and
+only one answer can be given.
+
+`Binary`, `Chain`, `Comparison` and `Conditional` are column-aligned in YASStyle too, but
+they are deliberately not listed here because the only caller (`n_binaryopcall!`) accounts
+for their full width before it will consider undoing a nesting, so it never has to move
+one of them that has already been split over several lines.
+"""
+function is_column_aligned(fst::FST)
+    if is_leaf(fst)
```
penelopeysm: `return if ...`

How about binaries?

penelopeysm: Yes confirmed there are still bugs with binaries

penelopeysm: Some bugs with binaries have been squashed but some issues with pathological inputs still remain (see #1232)

penelopeysm:
!formatbot penelopeysm/FlexiChains.jl --ignore-config --style=yas

penelopeysm:
!formatbot JuliaLang/julia@v1.12.6 --ignore-config --style=yas

penelopeysm:
Yes, that's exactly the inverse of the previous PR!

penelopeysm:
Most of those are fine. I still sense there are problems ....... .ugh

PR #1233, opened 2026-07-26
Add MinimalStyle to integration tests

penelopeysm:
Mostly curious as to what will happen


penelopeysm:
I bundled this change into #1240 already

PR #1236, opened 2026-07-26
Fix integration tests

penelopeysm:
GitHub was caching the formatted code, causing tests to run on already formatted code. That's quite embarrassing.

I just hope default/Blue is reasonably stable, or else, uh oh.


penelopeysm:
Phewwwwwwwww.

PR #1249, opened 2026-08-07
Adopt the packagedef file structure

davidanthoff:
This PR adopts the packagedef file structure. This is a quirk that allows us to use this package in more places inside the VS Code extension, i.e. with this file structure we can load this package in a private way, which is one of the things we do all over the place in the VS Code extension.

Many other packages in the community have adopted this style to help us out, like the whole Revise stack for example, so this design has been battle tested for many years now :)


penelopeysm:
Thanks @davidanthoff! I have no problems with this at all since it's just shuffling code around (indeed I am quite happy to move actual definitions outside of `JuliaFormatter.jl`) but just curious as to how this makes a difference for vscode? Web search for `julia packagedef file structure` didn't turn anything up which I suppose is consistent with your description of it as a quirk :)

penelopeysm:
(CI failures are expected so that's fine btw)

davidanthoff:
We often have cases where Julia instances run both our infrastructure code and user code. In those cases we try super hard to _not_ load our infrastructure code via the regular Julia package loading mechanism, because that would then pin any package we use to the version we use and the user can no longer load the version of that package they actually want to use. A good example is JSON: if our infrastructure code loads a specific version of the JSON package, then a user can't load any different version of that package anymore.

So what we do instead is that we load _all_ our infrastructure packages via `include` into one parent package. For that to work we need to not have any `using` or `import` statements in the packages we load, though. The way we achieve that is with this packagedef structure, you can see how that looks on our end in for example https://github.com/julia-vscode/julia-vscode/blob/main/scripts/packages/VSCodeServer/src/VSCodeServer.jl.

The key principle is that all `using` and `import` statements that are resolved via the package manager stay in the `JuliaFormatter.jl` file, and then anything beyond that is in the included packagedef.

penelopeysm:
Ah, makes a lot of sense, thanks for the explanation!

I'm going to add a comment into the file to make sure that future me or anyone doesn't inadvertently revert this, then will merge

davidanthoff:
Thanks :)

PR #1252, opened 2026-08-10
Keep whitespace around operators when gluing would change tokenization

### LINKED ISSUE #1250: MinimalStyle produces invalid/ambiguous code for trailing-dot float literals next to operators
`MinimalStyle` removes whitespace around binary operators without accounting for float literals written with a trailing dot. The result either fails to parse or triggers the "ambiguous `.` syntax" parser error, so `format_text` throws.

## MWE

```julia
julia> using JuliaFormatter  # v2.12.4, Julia 1.12.6

julia> format_text("x = A[1. .. 1., :]\n", MinimalStyle())
ERROR: formatted text could not be parsed as valid Julia
ParseError:
# Error @ line 1:11
x = A[1...1., :]
#         └ ── Expected `]`

julia> format_text("y = SA[1. + 2*cos(1.), 1.]\n", MinimalStyle())
ERROR: formatted text could not be parsed as valid Julia
ParseError:
# Error @ line 1:8
y = SA[1.+2*cos(1.), 1.]
#      └┘ ── ambiguous `.` syntax; add whitespace to clarify (eg `1.+2` might be `1.0+2` or `1 .+ 2`)
```

`DefaultStyle` is not affected because it normalizes the literal to `1.0` before joining, which removes the ambiguity.

## Expected

Keep the space when the token to the left of the operator ends in `.` (or the token to the right starts with `.`). Formatting valid input should never error.

## Context

Found by running the formatter over the 500 most-downloaded registered packages; this pattern occurs in released code, e.g.:

- AxisArrays.jl `test/sortedvector.jl`: `@test A[1. .. 1., :] == A[1:1, :]`
- DomainSets.jl `test/test_domain_ball.jl`: `@test approx_in(SA[1. + 2*cos(1.),1. + 2*sin(1.),1.], S)`

Thanks to the output re-parse validation the failure mode is an exception rather than corrupted output, but these files currently cannot be formatted with MinimalStyle at all.

davidanthoff:
Closes #1250

## Problem

`MinimalStyle` sets `trailing_zero = false`, so trailing-dot float literals like `1.` are emitted verbatim. When the whitespace around an adjacent binary operator was then removed, the output could tokenize differently than the input — or not parse at all:

```julia
julia> format_text("x = A[1. .. 1., :]\n", MinimalStyle())
ERROR: formatted text could not be parsed as valid Julia   # tried to emit A[1...1., :]

julia> format_text("y = SA[1. + 2*cos(1.), 1.]\n", MinimalStyle())
ERROR: ...ambiguous `.` syntax                             # tried to emit SA[1.+2*cos(1.), 1.]
```

A related case affected `DefaultStyle` too: `A[1 - -2]` was glued to `A[1--2]`, which does not parse (`--` is an invalid operator).

## Fix

Before removing the whitespace around an operator (`nws == 0`), `p_binaryopcall` now checks whether gluing the operator to its neighboring tokens changes the token stream: it tokenizes the glued and the spaced variant with JuliaSyntax and compares the resulting `(kind, text)` pairs. If they differ, a single space is kept on both sides.

Details:

- The check uses the **emitted** text of the boundary tokens, not the raw source: `DefaultStyle` normalizes `1.` → `1.0` (now factored into `normalize_number_literal`, shared with `p_literal`), after which gluing is safe — so `A[1. .. 1.]` still becomes `A[1.0..1.0]` there, same as before.
- Comparing token streams (not just looking for error tokens) is required because `1...1.` produces no error token, just a different split (`1`, `...`, `1.`).
- Tokens that cannot lexically merge with an operator (identifiers, closing brackets, string delimiters, …) are replaced by a neutral identifier before tokenizing, so e.g. `A["a" * "b"]` still glues to `A["a"*"b"]`. The adjoint `'` is given an identifier prefix for the same reason (`A[a' - b]` → `A[a'-b]` still works).
- Exact cases like `A[1. : 2]` → `A[1.:2]` (which tokenizes identically with or without spaces) keep gluing — this is not a blanket "float ends with a dot" heuristic.

## Tests

New `@testset "1250"` in `test/issues.jl`: both issue MWEs (MinimalStyle unchanged, DefaultStyle normalized+glued), `A[1 - -2]` across all five styles, and non-regression cases for safe gluing.


src/styles/default/pretty.jl
```diff
@@ -2707,6 +2760,94 @@ function p_pipe_to_call(
...
+    leaf, leaf_offset = result
+    k = kind(leaf)
+    offset = child_offset + leaf_offset
+    txt = getsrcval(s.doc, offset:(offset+span(leaf)-1))
+    return if k in KSet"Integer BinInt HexInt OctInt Float Float32"
+        normalize_number_literal(k, txt, s.opts.trailing_zero)
+    elseif k === K"'"
+        "x" * txt
+    elseif JuliaSyntax.is_operator(k) || k === K"."
+        txt
```
penelopeysm: I don't mind these three branches, but

src/styles/default/pretty.jl
```diff
@@ -2707,6 +2760,94 @@ function p_pipe_to_call(
...
+    else
+        "x"
+    end
```
penelopeysm, line 2817: This branch feels like a hack to me. If the token on either side can't merge with operators, then I think we should return a value that explicitly indicates this. That would allow us to avoid calling `gluing_changes_tokenization` in the case where *both* sides can't merge.

For example, once we remove the `nothing` return above, we could instead return `nothing` here to signify this.

src/styles/default/pretty.jl
```diff
@@ -2707,6 +2760,94 @@ function p_pipe_to_call(
...
+    child::JuliaSyntax.GreenNode,
+    child_offset::Int,
+    rightmost::Bool,
+)::Union{Nothing,String}
+    result = if rightmost
+        last_nonws_leaf_and_offset(child)
+    else
+        first_nonws_leaf_and_offset(child)
+    end
+    result === nothing && return nothing
```
penelopeysm: Is there any circumstance where `result` can be `nothing`? Since this is only called on binary operators, it seems to me that there should always be at least one non-whitespace token on both sides. The test suite passes locally for me with

```suggestion
    result === nothing && error("binop child has no non-whitespace node")
```

In my opinion, unless there's a reason to gracefully handle such errors, it's better to error loudly in such situations instead of silently returning `nothing`, since hitting this case would indicate that the code is not doing what we want it to do.

I will admit I haven't run this past the full integration test suite since that only runs on GHA, so there might be something I'm missing.

src/styles/default/pretty.jl
```diff
@@ -2880,6 +3021,58 @@ function p_binaryopcall(
...
+            [first(op_indices):last(op_indices)]
+        else
+            [i:i for i in op_indices]
+        end
+        for op_range in op_ranges
+            prev_idx =
+                findprev(c -> !JuliaSyntax.is_whitespace(c), childs, first(op_range) - 1)
+            next_idx =
+                findnext(c -> !JuliaSyntax.is_whitespace(c), childs, last(op_range) + 1)
+            (prev_idx === nothing || next_idx === nothing) && continue
```
penelopeysm: Again, this can't happen (I think)? Could we maybe `error()` instead of `continue`-ing here?

src/styles/default/pretty.jl
```diff
@@ -2880,6 +3021,58 @@ function p_binaryopcall(
...
+            op_text = getsrcval(
+                s.doc,
+                (child_offsets[first(op_range)]):(child_offsets[op_last]+span(
+                    childs[op_last],
+                )-1),
+            )
+            left = boundary_token_text(s, childs[prev_idx], child_offsets[prev_idx], true)
+            right =
+                boundary_token_text(s, childs[next_idx], child_offsets[next_idx], false)
+            (left === nothing || right === nothing) && continue
```
penelopeysm: If we change `boundary_token_text` to return `nothing` when no mis-tokenisation can occur, then this can instead be

```suggestion
            (left === nothing && right === nothing) && continue
```

src/styles/default/pretty.jl
```diff
@@ -2707,6 +2760,94 @@ function p_pipe_to_call(
...
+The comparison is done by actually tokenizing both variants with JuliaSyntax and comparing
+the (kind, text) pairs of the resulting non-whitespace tokens. Merely checking for error
+tokens would be insufficient: `1...1.` produces no error token at all, just a *different*
+token split (`1`, `...`, `1.`).
+"""
+function gluing_changes_tokenization(
+    left::AbstractString,
+    op::AbstractString,
+    right::AbstractString,
+)::Bool
```
penelopeysm: Finally, we'd need to change this function in order to handle the case where one side has `nothing` and the other side has actual text.

We can do this by broadening the type on `left` and `right` to accept `Union{AbstractString,Nothing}`, and then _inside_ this function if either `left` or `right === nothing` then we can replace them with the dummy `"x"` string. That's of course the same hack as before, but the logic would be self-contained within this function.

test/issues.jl
```diff
@@ -3658,6 +3658,39 @@ end
...
+        end
+
+        # Non-regression: gluing still happens whenever it is safe.
+        test_format("A[1 + 2]\n", "A[1+2]\n", MinimalStyle(); ast=true)
+        test_format("A[1 + 2]\n", "A[1+2]\n", DefaultStyle(); ast=true)
+        # `1.:2` tokenizes the same as `1. : 2`, so a trailing-dot float does not
+        # blanket-disable gluing.
+        test_format("A[1. : 2]\n", "A[1.:2]\n", MinimalStyle(); ast=true)
+        # The adjoint `'` and string delimiters next to an operator are fine to glue.
+        test_format("A[a' - b]\n", "A[a'-b]\n", MinimalStyle(); ast=true)
```
penelopeysm: Could we add one more test for `A[a'ᵀ - b]` as well? It would let me sleep more easily at night :-)

test/issues.jl
```diff
@@ -3658,6 +3658,39 @@ end
         end"""
         test_format(s_, s; join_lines_based_on_source=true)
     end
+
+    @testset "1250 no gluing operators when it changes tokenization" begin
+        # MinimalStyle has `trailing_zero = false`, so the trailing-dot floats survive
+        # as-is and gluing the operators to them would produce `A[1...1., :]` (invalid)
+        # and `SA[1.+2*cos(1.), 1.]` (ambiguous `.` syntax). The space must be kept.
+        test_format("x = A[1. .. 1., :]\n", "x = A[1. .. 1., :]\n", MinimalStyle(); ast=true)
```
penelopeysm: None of the trailing newlines in these tests are necessary (it's not a massive problem, but I prefer to not include them as they make the test a bit less targeted — on top of checking the binop handling, it is also implicitly checking that trailing newlines are preserved)

penelopeysm (changes requested):
Hi David! Thanks for the PRs and sorry it's taken a while (as you might guess, JuliaCon happened). I'll look over the others when I get a chance to. Very happy with this fix, the bug is quite nasty. I've requested some changes to keep the codebase more maintainable (or at least prevent it from descending into further unmaintainability), I don't think any of them are too onerous.

davidanthoff:
Awesome, Claude agreed with all your comments and hopefully addressed them all :)

penelopeysm (approved):
Thanks 😄

Issue #501, opened 2021-11-08
Weird multi-line string indentation


pfitzseb:
```
julia> print(JuliaFormatter.format_text("""
       write("", \"\"\"
           second
           third
           \"\"\")
       """))
write(
    "",
    """
second
third
""",
)
```
Not strictly an issue since it parses correctly, but does look weird.

Originally reported at https://github.com/julia-vscode/julia-vscode/issues/2545.

domluna:
this looks correct to me, if you look at how the text is placed in relation to the opening """ in the original text it's the same in the formatted output. It can lead to weird text in some cases, but you can indent the words more as one 1 way to get around it.

This nests because the current rule is that if a string is multiline in a function call, the function call is automatically nested.

domluna:
true, would this backfire in other cases though?

pfitzseb:
Well, Julia disregards
- a leading newline
- the whitespace all lines have in common

Sooo, sure, there are case where indentation isn't possible.

domluna:
would it be bad to align the lines to the first " if they currently appear prior to it? If it's a good idea it wouldn't be difficult to do (i think)

mortenpi:
Not 100% sure if it's related, but this

```julia
function myfunction_foo(io::IO)
    write(io, """
    """)
end
```

goes into

```julia
function myfunction_foo(io::IO)
    write(
        io,
        """
""",
    )
end
```

which does seem incorrect to me.

domluna:
that's intentional. the distance between the start and end quotes stays consistent.

mortenpi:
That doesn't seem necessary? The string wouldn't change? So it could just go into

```julia
function myfunction_foo(io::IO)
    write(
        io,
        """
        """,
    )
end
```

Or if we do insist on keeping the distance, then it could go into

```julia
function myfunction_foo(io::IO)
    write(
        io,
            """
        """,
    )
end
```

which would still be a bit weird, but slightly better? Basically, it should be aligning the ending `"""`, not the starting one.

pfitzseb:
Yeah, that's not required. Julia doesn't care about leading whitespace inside of the string, cf
```
julia> x = """
       """
""

julia> x = """
           """
""

julia> x = """

           """
"\n"

julia> x = """

       """
"\n"

julia> x = """
           ab
           """
"ab\n"

julia> x = """
           ab
       """
"    ab\n"

julia> x = """
        ab
       """
" ab\n"
```

[...]

PR #1253, opened 2026-08-10
Preserve contents of non-triple-quoted multiline string literals when re-indenting

### LINKED ISSUE #1251: Re-indentation changes the contents of multiline string literals (DefaultStyle and MinimalStyle)
When the formatter re-indents code, continuation lines that are *inside* an ordinary (non-triple-quoted) multiline string literal are shifted along with the surrounding code. For a regular `"..."` literal every leading space is part of the string's value, so this silently changes program behavior.

## MWE

```julia
julia> using JuliaFormatter  # v2.12.4, Julia 1.12.6

julia> print(format_text("""
       function f()
               x = "line one
               line two"
               return x
       end
       """, MinimalStyle()))
function f()
    x = "line one
    line two"
    return x
end
```

The string's value changes from `"line one\n        line two"` to `"line one\n    line two"`. Same result with `DefaultStyle()`. The output parses fine, so nothing catches this — the semantic change is silent.

## Expected

Multiline string/cmd/macro-string literals should be treated as opaque during re-indentation: only the line containing the opening delimiter may move; interior lines (and the closing-delimiter line for non-triple-quoted strings) must keep their exact text. (For triple-quoted strings a uniform shift can be content-preserving because of dedent semantics, but that reasoning does not apply to regular `"..."` literals.)

## Real-world impact

Found by formatting the 500 most-downloaded registered packages and comparing before/after ASTs (with `LineNumberNode`s stripped): this bug accounts for every semantic difference found — 37 files across 12 packages, including:

- Latexify.jl (11 test files): reference LaTeX strings used in `@test` comparisons — formatting the package makes its tests fail
- PlutoUI.jl (6 files): HTML/JS templates shipped to the browser
- WebSockets.jl examples: HTML served to clients
- OrdinaryDiffEq.jl (9 files): docstring/citation strings passed as macro arguments
- Error/log message strings in Flux.jl, CUDA.jl, AbstractAlgebra.jl, LoopVectorization.jl, NonlinearSolve.jl, RandomExtensions.jl
- YAML.jl: multiline `r"..."x` regex literals

Related but distinct: #501 covers odd (content-preserving) re-indentation of triple-quoted strings; this issue is about actual content changes to regular multiline literals.

davidanthoff:
Closes #1251

## Problem

Re-indenting code shifted the continuation lines *inside* ordinary (non-triple-quoted) multiline string, cmd, and string-macro literals along with the surrounding code:

```julia
julia> print(format_text("""
       function f()
               x = "line one
               line two"
               return x
       end
       """, MinimalStyle()))
function f()
    x = "line one
    line two"          # <- string value silently changed
    return x
end
```

For a regular `"..."` literal every leading space of a continuation line is part of the string's value, so this silently changes program behavior — and since the output still parses, nothing catches it. All five styles were affected.

## Fix

`p_stringh` now detects whether the literal is triple-quoted by looking at its opening delimiter token (`"""` / ` ``` `). Non-triple multiline literals are marked *opaque* via a new `is_opaque_string` flag on `Metadata`:

- their interior lines are stored **verbatim** (full original text, indent 0) instead of being dedented to the common column;
- `print_string` and `n_string!` skip the uniform indent shift for opaque nodes, so interior lines are reproduced byte-for-byte and only the opening-delimiter line moves.

Triple-quoted literals (which have dedent semantics, so uniform shifting is content-preserving — see #501) and the `format_docstrings` path (which deliberately re-indents docstrings) keep their existing behavior. Width/nesting decisions are unchanged: the `StringN` length formula gives the same result under the verbatim representation.

## Tests

New `@testset "1251"` in `test/issues.jl`: the MWE with exact expected output, value-preservation (idempotence + AST equality) across all five styles for `"..."`, `` `cmd` ``, `r"..."x`, `raw"..."`, blank interior lines, interplay with `v2_stable_multiline_strings`, plus a triple-quoted control asserting the current shifting behavior is untouched.

Context: formatting the 500 most-downloaded registered packages and comparing before/after ASTs, this bug accounted for every semantic difference found (37 files across 12 packages).


test/issues.jl
```diff
@@ -3658,6 +3658,78 @@ end
         end"""
         test_format(s_, s; join_lines_based_on_source=true)
     end
+
+    @testset "1251 re-indentation must not change multiline string contents" begin
```
penelopeysm, line 3698: Very happy with the rest of the PR, with only one request - could we run all the tests in this testset across `ALL_STYLES` just to be safe? I think the only problem with `ALL_STYLES` is usually the function return being added silently -- if you want a test that doesn't have all those issues you can replace the function with `begin ... end`.

penelopeysm: Actually, I'll just get my claude to do it. 😉

penelopeysm (changes requested):

PR #1237, opened 2026-07-26
Drop semicolons when expanding single-line constructs to blocks

penelopeysm:
This PR tweaks the output of JuliaFormatter to strip needless semicolons inside block constructs.

For example, with this PR:

```julia
julia> s = "for x in y; foo; end"
"for x in y; foo; end"

julia> format_text(s) |> println
for x in y
    foo
end
```

On main:

```julia
julia> format_text(s) |> println
for x in y
    ;
    foo;
end
```

Top-level semicolons are not removed because they carry meaning (they suppress output in the REPL). For example, this semicolon is preserved:

```julia
julia> s = "x; y"
"x; y"

julia> format_text(s)
"x;\ny"
```

Issue #594, opened 2022-05-24
Do not insert line break after =


davidanthoff:
[Here](https://github.com/julia-vscode/SymbolServer.jl/pull/248/files#r825260479) is an example where it inserts a line break after a `=`. It changes
```julia
getallns() = let allns = Base.IdSet{Symbol}(); oneverything((m, s, x, state) -> push!(allns, s)); allns end
```
into
```julia
getallns() =
    let allns = Base.IdSet{Symbol}()
        oneverything((m, s, x, state) -> push!(allns, s))
        allns
    end
```
which it shouldn't do for the VS Code default.

domluna:
@davidanthoff what would the output be?

davidanthoff:
Just wanted to bump this, I just tested this on the latest tagged version, and it still inserts line breaks for the `MinimalStyle`, which it shouldn't. Any chance we could fix this? :)

PR #1254, opened 2026-08-10
Add preserve_single_line_blocks option to keep single-line blocks on one line

### LINKED ISSUE #593: Do not rewrite semicolon separated lines as separate lines
[Here](https://github.com/julia-vscode/SymbolServer.jl/pull/248/files#r825260400) is an example where
```julia
safe_isfile(x) = try isfile(x); catch; false end
```
is turned into
```julia
safe_isfile(x) =
    try
        isfile(x)
    catch
        false
    end
```
which it shouldn't do for the VS Code default.

davidanthoff:
Closes #593

## Problem

MinimalStyle (the default style used by the Julia VS Code extension) expanded block constructs written on a single source line:

```julia
safe_isfile(x) = try isfile(x); catch; false end
```

became

```julia
safe_isfile(x) =
    try
        isfile(x)
    catch
        false
    end
```

The root cause is that every block-owning construct (`p_functiondef`, `p_try`, `p_let`, `p_if`, `p_for`, `p_do`, ...) unconditionally passes `ignore_single_line = true` to `p_block` and then places the body on its own indented line.

## Fix

A new formatting option, **`preserve_single_line_blocks`** (default `false`, set to `true` in `MinimalStyle`'s options). When enabled, a block construct written entirely on one source line keeps its single-line, semicolon-separated form; constructs spanning multiple lines are expanded as before.

Gating on the existing `join_lines_based_on_source` was not an option since YAS and SciML also enable it and deliberately expand one-liners; a new option follows the repo's "styles are option bundles" architecture and lets DefaultStyle users opt in.

Covered constructs: `try`/`catch`/`else`/`finally`, `let`, `if`/`elseif`/`else`, `for`, `while`, `function`/`macro` definitions, `do`, `struct`/`mutable struct`, `module`/`baremodule`, `begin`, and `quote`.

Implementation: a `keep_single_line(cst, s)` helper (option enabled + construct on one source line) is checked at each construct's entry; when true, the body block is prettified *without* `ignore_single_line` (so `p_block` joins its statements with `; `) and the block/keywords/`end` are joined onto the same line instead of being newline-forced via `max_padding`. Blocks without statements (e.g. `baremodule Q; end`) keep the existing behavior of dropping superfluous semicolons.

Notes:

- The line break still inserted after `=` in the `safe_isfile` example above is a separate issue (#594) with a follow-up PR stacked on this one.
- Semicolon-separated statements *inside* a multi-line construct (e.g. an `f(i); g(i)` line in a multi-line `for` body) are still split, as introduced in v2.12.0 (#1237); this PR only affects constructs that are single-line in their entirety.

Also wired up: `--preserve-single-line-blocks` CLI flag, docs (`formatting_options.md` table + section), version 2.13.0 + HISTORY entry (new-option minor bump).

## Tests

New `@testset "593"` in `test/issues.jl`: ~20 single-line constructs asserted unchanged under MinimalStyle (with idempotence and AST equality), multi-line sources still expanding, DefaultStyle default behavior unchanged, standalone option usage with DefaultStyle, and YAS/SciML non-regression. Plus CLI flag tests in `test/argparse.jl`.

### DIFF
--- a/HISTORY_v2.md
+++ b/HISTORY_v2.md
@@ -1,3 +1,7 @@
+# v2.13.0
+
+Added a new formatting option, `preserve_single_line_blocks` (default `false`; `true` for `MinimalStyle`). When enabled, block constructs written entirely on a single source line — e.g. `try isfile(x); catch; false end` or `let x = 1; x end` — are kept on one line, with their statements separated by semicolons, instead of being expanded onto multiple lines. This applies to `function`/`macro` definitions, `for`/`while` loops, `if`, `let`, `try`, `do`, `struct`/`mutable struct`, `module`/`baremodule`, `begin`, and `quote`. (#593, #1254)
+
 # v2.12.4
 
 Fixed a bug causing lack of idempotence when extra lines inside comma-separated arguments (e.g. in a vector) were present. (#1246, #1247)
--- a/docs/src/formatting_options.md
+++ b/docs/src/formatting_options.md
@@ -77,6 +77,7 @@ If you absolutely must enable them, you may need to run the formatter multiple t
 | [`max_iterations`](@ref options-max-iterations)                                     |       | `1`       | `1`         | `1`         | `1`          | `1`           |
 | [`normalize_line_endings`](@ref options-normalize-line-endings)                     | 📐    | `"auto"`  | `"auto"`    | `"auto"`    | **`"unix"`** | `"auto"`      |
 | [`pipe_to_function_call`](@ref options-pipe-to-function-call)                       | 🪃 🔥 | `false`   | **`true`**  | **`true`**  | `false`      | `false`       |
+| [`preserve_single_line_blocks`](@ref options-preserve-single-line-blocks)          | 📐    | `false`   | `false`     | `false`     | `false`      | **`true`**    |
 | [`remove_extra_newlines`](@ref options-remove-extra-newlines)                       | 📐    | `false`   | **`true`**  | **`true`**  | **`true`**   | `false`       |
 | [`sciml_margin_overrun`](@ref options-sciml-margin-overrun)                         | 📐    | unused    | unused      | unused      | **`20`**     | unused        |
 | [`separate_kwargs_with_semicolon`](@ref options-separate-kwargs-with-semicolon)     | ⚠️    | `false`   | **`true`**  | **`true`**  | `false`      | `false`       |
@@ -595,6 +596,31 @@ If true, `x |> f` is rewritten to `f(x)`, and `x .|> f` to `f.(x)`.
     *In principle*, this can be fixed, since we can determine ahead of time when the pipe will be transformed into the function call.
     However, my current judgment is that this is not really worth it since it would make the codebase more complex and brittle.
     On top of that, it's probably not a good idea to enable `pipe_to_function_call` anyway.
+
+## [`preserve_single_line_blocks`](@id options-preserve-single-line-blocks)
+
+Default: `false` (`true` for `MinimalStyle`)
+
+If true, block constructs (`function`/`macro` definitions, `for`/`while` loops, `if`, `let`, `try`, `do`, `struct`, `module`, `begin`, ...) that are written entirely on a single source line are kept on a single line, with their statements separated by semicolons, instead of being expanded onto multiple lines.
+
+For example, with this option enabled
+
+```julia
+safe_isfile(x) = try isfile(x); catch; false end
+```
+
+is left unchanged, whereas by default it is rewritten to
+
+```julia
+safe_isfile(x) =
+    try
+        isfile(x)
+    catch
+        false
+    end
+```
+
+Constructs that span multiple source lines are expanded as usual.
 
 ## [`remove_extra_newlines`](@id options-remove-extra-newlines)
 
--- a/src/argparse.jl
+++ b/src/argparse.jl
@@ -492,6 +492,13 @@ const PARSER = ArgParser(
         group = FormattingGroup,
     ),
     option(
+        "--preserve-single-line-blocks";
+        dest = :preserve_single_line_blocks,
+        type = Bool,
+        help = "Keep single-line block constructs (e.g. `try x; catch; y end`) on one line.",
+        group = FormattingGroup,
+    ),
+    option(
         "--remove-extra-newlines";
         dest = :remove_extra_newlines,
         type = Bool,
--- a/src/styles/default/pretty.jl
+++ b/src/styles/default/pretty.jl
@@ -1525,6 +1525,12 @@ function p_functiondef(
             if kind(c) !== K"Comment"
                 error("Unexpected node after function body: $(kind(c))")
             end
+            if keep_single
+                # preserve_single_line_blocks: keep the comment on the same line.
+                add_node!(t, Whitespace(1), s)
+                add_node!(t, pretty(style, c, s, ctx, lineage), s; join_lines = true)
+                continue
+            end
             if extra_block_node === nothing
                 s.indent += s.opts.indent
                 extra_block_node = FST(Block, nspaces(s))
@@ -2213,6 +2219,12 @@ function p_for(
             if kind(c) !== K"Comment"
                 error("Unexpected node after loop body: $(kind(c))")
             end
+            if keep_single
+                # preserve_single_line_blocks: keep the comment on the same line.
+                add_node!(t, Whitespace(1), s)
+                add_node!(t, pretty(style, c, s, ctx, lineage), s; join_lines = true)
+                continue
+            end
             if extra_block_node === nothing
                 s.indent += s.opts.indent
                 extra_block_node = FST(Block, nspaces(s))
@@ -2606,6 +2618,12 @@ function p_if(
             # See comments in `p_for` for explanation of this
             if kind(c) !== K"Comment"
                 error("Unexpected node after if body: $(kind(c))")
+            end
+            if keep_single
+                # preserve_single_line_blocks: keep the comment on the same line.
+                add_node!(t, Whitespace(1), s)
+                add_node!(t, pretty(style, c, s, ctx, lineage), s; join_lines = true)
+                continue
             end
             if extra_block_node === nothing
                 s.indent += s.opts.indent
--- a/src/styles/default/pretty.jl
+++ b/src/styles/default/pretty.jl
@@ -1176,6 +1176,51 @@ end
 
 # Block
 # length Block is the length of the longest expr
+"""
+    keep_single_line(cst::JuliaSyntax.GreenNode, s::State)::Bool
+
+Whether a block construct that is written entirely on a single source line should be kept
+on a single line (see the `preserve_single_line_blocks` option). For example, with the
+option enabled,
+
+    safe_isfile(x) = try isfile(x); catch; false end
+
+keeps the `try` on one line instead of being expanded to
+
+    safe_isfile(x) =
+        try
+            isfile(x)
+        catch
+            false
+        end
+
+IMPORTANT: this must be called when `s.offset` is still positioned at the start of `cst`
+(i.e. at the start of the construct, before any of its children have been prettified),
+since it compares the source line of the first and last byte of `cst`.
+"""
+function keep_single_line(cst::JuliaSyntax.GreenNode, s::State)::Bool
+    return s.opts.preserve_single_line_blocks &&
+           on_same_line(s, s.offset, s.offset + span(cst) - 1)
+end
+
+"""
+    starts_with_semicolon(fst::FST)::Bool
+
+Whether the first leaf of `fst` is a semicolon. Used when joining a single-line block onto
+the line of its parent construct: in `try; isfile(x); catch; false end` the body block
+starts with a `;`, and no space should be inserted between the `try` keyword and the `;`
+(`try; isfile(x)` rather than `try ; isfile(x)`).
+"""
+function starts_with_semicolon(fst::FST)::Bool
+    n = fst
+    while !is_leaf(n)
+        nodes = n.nodes::Vector{FST}
+        isempty(nodes) && return false
+        n = nodes[1]
+    end
+    return n.typ === SEMICOLON
+end
+
 function p_block(
     ds::AbstractStyle,
     cst::JuliaSyntax.GreenNode,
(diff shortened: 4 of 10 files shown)


penelopeysm:
Ok, I handled the merge conflicts and bumped the minor version again. I probably won't have time to review the implementation until the weekend, however, before I fixed the merge conflicts the integration tests for MinimalStyle were failing, due to idempotence failures: https://github.com/JuliaEditorSupport/JuliaFormatter.jl/actions/runs/31414368554/job/93539744210 I think the first step would be to fix that CI.

The only known failures for that workflow should be `default-jlbos` and `yas`. It was a _bit_ confusing that those were expected to fail, so i updated their names so that it's more obvious what's expected to fail. The others (including Minimal) should ideally pass otherwise it'd be a regression.

docs/src/formatting_options.md
```diff
@@ -620,6 +621,31 @@ If true, `x |> f` is rewritten to `f(x)`, and `x .|> f` to `f.(x)`.
...
+is left unchanged, whereas by default it is rewritten to
+
+```julia
+safe_isfile(x) =
+    try
+        isfile(x)
+    catch
+        false
+    end
+```
```
penelopeysm: As before it would be lovely to demo this by running `format_text(s; opts...) |> println` (there are other examples to crib from on the same page)

src/styles/default/pretty.jl
```diff
@@ -1259,6 +1259,51 @@ end
...
+    safe_isfile(x) = try isfile(x); catch; false end
+
+keeps the `try` on one line instead of being expanded to
+
+    safe_isfile(x) =
+        try
+            isfile(x)
+        catch
+            false
+        end
```
penelopeysm: The example probably not needed in the docstring (should be obvious from the reference to the option).

src/styles/default/pretty.jl
```diff
@@ -1259,6 +1259,51 @@ end
...
+
+IMPORTANT: this must be called when `s.offset` is still positioned at the start of `cst`
+(i.e. at the start of the construct, before any of its children have been prettified),
+since it compares the source line of the first and last byte of `cst`.
+"""
+function keep_single_line(cst::JuliaSyntax.GreenNode, s::State)::Bool
```
penelopeysm: I think in this function it would be nice to just do a sanity check on `kind(cst)`, something like

```julia
kind(cst) in KSet"if try for ..." || error("unreachable: keep_single_line called on non-block node")
```

(I'm not sure the exact kinds off the top of my head, but LLMs are great at that)

src/styles/default/pretty.jl
```diff
@@ -1291,8 +1336,17 @@ function p_block(
         return p_tupleblock(style, cst, s, ctx, lineage)
     end
 
+    # Block-owning constructs (function/for/if/let/try/do/...) pass
+    # `ignore_single_line = true` so that their bodies are always expanded onto separate
+    # lines. When `preserve_single_line_blocks` is enabled, a body written on a single
+    # source line keeps its single-line, semicolon-separated form instead. Blocks without
+    # any statements (e.g. the `;` in `baremodule Q; end`) keep the existing behavior of
+    # dropping the superfluous semicolons.
```
penelopeysm: This comment isn't really necessary imo- or I think it would be better to document the ignore_single_line field on PrettyContext.

src/argparse.jl
```diff
@@ -499,6 +499,13 @@ const PARSER = ArgParser(
         help = "Convert pipe operator to function calls.",
         group = FormattingGroup,
     ),
+    option(
+        "--preserve-single-line-blocks";
+        dest = :preserve_single_line_blocks,
+        type = Bool,
+        help = "Keep single-line block constructs (e.g. `try x; catch; y end`) on one line.",
```
penelopeysm:
```suggestion
        help = "Keep single-line block constructs on one line.",
```
help text ideally is short and snappy since terminal columns are limited

test/issues.jl
```diff
@@ -1388,6 +1388,77 @@ end
         test_format(s1, s2; indent=4, margin=1)
     end
 
+    @testset "593 preserve single-line blocks" begin
```
penelopeysm: General testing comments:

(1) This could be moved into a nice `test/options/preserve_single_line_blocks.jl` file rather than a single-issue testset which is a bit hard to find

(2) i think we should test explicitly with the option, rather than testing `format_text(s, MinimalStyle())`. The reason being that it's really the option that's controlling it, and MinimalStyle just so happens to set the option to true. I don't really think it's necessary to test that any given style sets any given option to any given value (it's more or less guaranteed from the way the options are set up).

(3) it would also probably be worth testing with `#= comments =#` inside the block (check that the contents are unchanged) since this has always been a massive pain point with JuliaFormatter and it's really easy to introduce bugs where the comments are just deleted or mangled.

(4) could we also add a test that unannotated struct fields get `::Any` when `preserve_single_line_blocks=true` and `annotate_untyped_fields_with_any=true`, i.e., `struct Foo x end` becomes `struct Foo x::Any end` too. This is definitely already the case but would be good to avoid regressions.

test/issues.jl
```diff
@@ -1388,6 +1388,77 @@ end
...
+            "struct A; x::Int end",
+            "mutable struct B; x end",
+            "module M; f() = 1 end",
+            "begin x = 1; y = 2 end",
+            "quote x = 1; y end",
+            # nested single-line constructs
+            "let x = 1; if x; 1 end end",
+        ]
+        for c in unchanged
+            test_format(c * "\n", c * "\n", MinimalStyle(); ast=true)
```
penelopeysm:
```suggestion
            test_format(c, c; preserve_single_line_blocks=true, ast=true)
```

also no need the newline

HISTORY_v2.md
```diff
@@ -27,6 +33,7 @@ Triple-quoted literals are unaffected. (#1251, #1253)
 
 Fixed a bug where removing the whitespace around a binary operator could change how the code tokenizes, producing invalid or ambiguous output.
 For example `A[1. .. 1.]` was formatted to `A[1...1.]` under `MinimalStyle` and `A[1 - -2]` was formatted to `A[1--2]` under `DefaultStyle`, neither of which parses. (#1250, #1252)
+>>>>>>> main
```
penelopeysm:
```suggestion
```

Sorry, this is my bad :)

penelopeysm (changes requested):
The implementation is fine imo, it's not pretty but I think that's more or less necessary. Just had some comments around the testing (on top of the idempotence issues in CI).

src/styles/default/pretty.jl
penelopeysm: > document the ignore_single_line field on PrettyContext

I can do this separately, so don't worry about lumping it into this PR :)

davidanthoff:
Awesome, I'll have Claude fix these in a second.

docs/src/formatting_options.md
davidanthoff: _(This reply was written by Claude Code, on behalf of @davidanthoff.)_

Done in a7f680b — replaced the static before/after with a `@example` block that runs `format_text(s; preserve_single_line_blocks=true) |> println` and the `=false` variant. It uses `for`/`try` one-liners rather than `safe_isfile(x) = ...`, since #594 still inserts a newline after `=` and would make the demo look odd.

src/styles/default/pretty.jl
davidanthoff: _(This reply was written by Claude Code, on behalf of @davidanthoff.)_

Done — the example is gone; the docstring now just describes what the function decides (plus the note about `s.offset` needing to be at the start of the node).

src/styles/default/pretty.jl
davidanthoff: _(This reply was written by Claude Code, on behalf of @davidanthoff.)_

Done — added a `SINGLE_LINE_BLOCK_KINDS` constant (`function macro struct module baremodule block quote let for while do try catch finally else if elseif`) and `keep_single_line` now does `kind(cst) in SINGLE_LINE_BLOCK_KINDS || error("unreachable: keep_single_line called on non-block node ...")`.

src/styles/default/pretty.jl
davidanthoff: _(This reply was written by Claude Code, on behalf of @davidanthoff.)_

The comment is gone, together with the code it described: this `p_block` change turned out to be the cause of the `base/shell.jl` idempotence failure (it kept the `;` for any *body* written on one line, even when the enclosing `if` was multi-line — e.g. `else; seg_stderr = re; end`). `p_block` is back to its original form; the keep-single-line paths in the individual constructs never relied on it.

src/argparse.jl
davidanthoff: _(This reply was written by Claude Code, on behalf of @davidanthoff.)_

Applied.

test/issues.jl
davidanthoff: _(This reply was written by Claude Code, on behalf of @davidanthoff.)_

All four done in a7f680b:

1. Moved to `test/options/preserve_single_line_blocks.jl` (and removed from `test/issues.jl`).
2. Tests now pass `preserve_single_line_blocks=true` to `test_format` instead of `MinimalStyle()`; the only style-specific tests left are the YAS/SciML "still expands" non-regression checks.
3. Added a "comments inside blocks" testset with `#= =#` before/after statements in `if`/`for`/`while`/`function`/`try`/`let`/`begin`/`struct`/`do`. Good call — it caught a real bug: `begin #= c =# x end` silently *dropped* the comment (the single-line path in `p_begin` was not routing `K"Comment"` nodes through `add_hasheq_comment!`). Fixed.
4. Added: `struct Foo x end` → `struct Foo x::Any end`, plus `struct Foo; x end` and a `mutable struct` with a mix of typed/untyped fields.

test/issues.jl
davidanthoff: _(This reply was written by Claude Code, on behalf of @davidanthoff.)_

Applied (no trailing newline, option kwarg instead of `MinimalStyle()`).

HISTORY_v2.md
davidanthoff: _(This reply was written by Claude Code, on behalf of @davidanthoff.)_

That marker was already removed in 8c40c9b (before this round), so nothing left to do here.

davidanthoff:
_(This comment was written by Claude Code, on behalf of @davidanthoff.)_

a7f680b addresses the review and the failing `minimal` Integration job. The four idempotence failures had four distinct causes:

| CI diff | Cause | Fix |
|---|---|---|
| `base/shell.jl` — `else; seg_stderr = re; end` in a multi-line `if` became `else\n    ; seg_stderr = re;` | The `p_block` change kept `;` for any *body* on one source line, even when the enclosing construct was multi-line | Reverted the `p_block` change entirely; the per-construct paths never needed it |
| `test/docs.jl` — `quote "Second docstring" Module29432 end` | A docstring is always emitted on its own line, so the "single-line" `quote` was multi-line after pass 1 | `keep_single_line` returns `false` for constructs containing a `K"doc"` node; they expand normally (which is what pass 2 produced) |
| `test/syntax.jl` — `macro z49984(s); :(let a; $(esc(s)); end); end` | `p_parens` (and `p_comprehension`) force `AlwaysNest` whenever the argument is a block, so the parens broke onto three lines and the macro was multi-line on pass 2 | Nesting is only forced when the block is actually going to be expanded (`block_arg_forces_nest`) |
| `StatsModels.jl` — `result = quote begin end end` | With `join_lines_based_on_source`, `p_begin` put the empty body's `end` on a new line | `p_begin` keeps an empty `begin end` joined when the construct is kept single-line |

Related issue found along the way: the `elseif`/`else`/`catch`/`finally` sub-nodes recomputed the single-line decision on their own span (which contains neither the construct's start nor its `end`), so a multi-line `if` with a one-line `elseif ...; ...` clause was half-joined. They now inherit the decision from the enclosing construct via a new `PrettyContext.parent_keep_single_line` field, matching the documented semantics ("constructs that span multiple source lines are expanded as usual").

Verification:
- Full test suite passes locally.
- A local replica of the Integration job (Julia 1.12's shipped `base/` + `test/`, which contain `shell.jl`/`docs.jl`/`syntax.jl`, plus a StatsModels.jl clone) is idempotent for `--style=minimal` and `--style=default` with the workflow's options. The one remaining `minimal` diff (`base/JuliaSyntax/src/parse_stream.jl`, a parenthesised range split over two lines) is identical on `main`, and that file isn't part of the CI corpus.

Two behavioural notes for the record:
- Under `MinimalStyle`, a one-line `begin end` / `quote end` now stays on one line (previously it became `begin\nend`) — a side effect of the `p_begin` fix, consistent with the option's intent.
- `f(x) = try ...; catch; ... end` still gets a line break after `=`; that's #594, handled in the stacked follow-up.

src/styles/default/pretty.jl
```diff
@@ -1259,6 +1264,93 @@ end
...
+
+Whether `arg` -- the sole argument of a bracketed construct such as `(BLOCK)` or
+`[BLOCK for x in y]`, starting at byte `offset` of the source -- is a block that will be
+expanded onto multiple lines, in which case the brackets must always nest. With
+`preserve_single_line_blocks`, a block that is kept on a single line does not force
+nesting.
+"""
+function block_arg_forces_nest(
+    arg::JuliaSyntax.GreenNode,
+    offset::Int,
```
penelopeysm:
```suggestion
    offset::Integer,
```

It fails to precompile on Windows x86 because of this (see CI).

penelopeysm:
This is looking great! I'll fix a couple of things, then release.

penelopeysm:
!formatbot JuliaLang/julia@v1.12.7 --style=default

penelopeysm:
> Formatting with base and PR is not idempotent!

~~Okay, well, that's a problem (albeit not one introduced by this PR, so it won't block merging). CI tests that it's idempotent on julia master, but I guess we should explicitly check it on 1.12.7...~~

Oh, no, it's just that I forgot the magic incantation to make things idempotent.

penelopeysm:
!formatbot JuliaLang/julia@v1.12.7 --style=default --v2-stable-multiline-strings=true

penelopeysm (commented):
OK. I think we're 95% of the way there. All tests are looking great, MinimalStyle is fine, and we also know from formatbot that the changes here don't cause any formatting differences when the option is not enabled. However, I added `--style=default --preserve-single-line-blocks=true` to the integration tests in CI and that throws up a few idempotence bugs. I looked through the logs and I think virtually all of them stem from the same construct:

```julia
julia> s1 = "if looooong f(looooooooooong) end";

julia> # Splits `f(...)` onto separate lines, but not `if ... end`
       s2 = format_text(s1; margin=10, preserve_single_line_blocks=true); println(s2)
if looooong f(
    looooooooooong,
) end

julia> # No longer a single line block so `if` and `end` get their own lines
       format_text(s2; margin=10, preserve_single_line_blocks=true) |> println
if looooong
    f(
        looooooooooong,
    )
end
```

<sup>(Incidentally, it's a similar situation to my comment in this prior PR: https://github.com/JuliaEditorSupport/JuliaFormatter.jl/pull/937#issuecomment-4563423213)</sup>

This isn't a problem with MinimalStyle, because MinimalStyle sets the margin to be some very large number. However it does mean that any other user of the option might run into problems like these.

I think it makes sense that `preserve_single_line_blocks` should override the margin for blocks that are being kept as a single line. Conceptually the fix isn't too hard: basically if we detect `keep_single_line`, then we should just disable nesting for the contents of the block.

penelopeysm:
That very much did it!

PR #1256, opened 2026-08-10
MinimalStyle: indent operator chain continuations instead of aligning

### LINKED ISSUE #592: Consistently use indent instead of aligning things for VS Code
[Here](https://github.com/julia-vscode/SymbolServer.jl/pull/248/files#r825259761) is an example where the formatter seems to fall back on aligning things. It turns
```julia
        @map({
            name = _[2]["name"],
            uuid = UUID(_[1]),
            path = _[2]["path"]
        }) |>
        @mutate(
            versions = (Pkg.TOML.parsefile(joinpath(registry_folder_path, _.path, "Versions.toml")) |>
                @map(i->{version=VersionNumber(i[1]), treehash=i[2]["git-tree-sha1"]}) |> 
                @orderby_descending(i->i.version) |>
                @take(max_versions) |>
                collect)
        ) |>
        collect
```
into
```julia
               @map({
                   name = _[2]["name"],
                   uuid = UUID(_[1]),
                   path = _[2]["path"]
               }) |>
               @mutate(
                   versions = (Pkg.TOML.parsefile(joinpath(registry_folder_path, _.path, "Versions.toml")) |>
                               @map(i -> {version = VersionNumber(i[1]), treehash = i[2]["git-tree-sha1"]}) |>
                               @orderby_descending(i -> i.version) |>
                               @take(max_versions) |>
                               collect)
               ) |>
               collect
```
which it shouldn't do.

davidanthoff:
Closes #592

## Problem

MinimalStyle (the VS Code default) aligned continuation lines of operator chains, comparisons, and binary operator calls to the column of the first operand:

```julia
x = (aaa |>
     bbb |>       # <- aligned to the column of `aaa`
     ccc)
```

and for top-level chains the continuation lines received no indent at all (`aaa |>\nbbb` kept `bbb` at column 0). The issue discussion settled on consistent indentation instead: each continuation line gets one indent level relative to the statement.

## Fix

New file `src/styles/minimal/nest.jl` with `n_chainopcall!`, `n_comparison!`, and `n_binaryopcall!` overrides for `MinimalStyle`, following the existing BlueStyle pattern (`src/styles/blue/nest.jl`): they forward to the default implementations with `indent = fst.indent + s.opts.indent` instead of letting them fall back to align-to-`s.line_offset`. Standalone short-circuit chains skip the extra indent since `n_block!` already adds one level for them.

The `indent` keyword only affects the non-`is_indent_nest` branch of `n_binaryopcall!`; assignments, short-form defs, `=>`, and `->` already used indent-based nesting and are unchanged (so this is independent of #1255/#594).

Example output for the pipeline from the issue:

```julia
@map({
        name = _[2]["name"],
        uuid = UUID(_[1]),
        path = _[2]["path"]
    }) |>
    @mutate(
        versions = (Pkg.TOML.parsefile(joinpath(registry_folder_path, _.path, "Versions.toml")) |>
            @map(i->{version=VersionNumber(i[1]), treehash=i[2]["git-tree-sha1"]}) |>
            @orderby_descending(i->i.version) |>
            @take(max_versions) |>
            collect)
    ) |>
    collect
```

This matches the "+1 indent level per chain continuation" target from the issue thread, with one nuance: interior lines of a multiline *first* operand (the `{...}` braces above) also shift one level, since the chain indent applies uniformly to all of the chain's continuation lines. The output is idempotent.

## Tests

New `@testset "592"` in `test/issues.jl`: top-level and nested `|>` chains, aligned input converging to the indented form, chains in function bodies/`return`, comparisons in `if` conditions, standalone `&&` chains (no double indent), and non-regression tests asserting DefaultStyle and YASStyle keep their aligned behavior.


penelopeysm (commented):
I'm going to start by saying that personally I _much_ prefer the indented version that this PR does. But personally, I think this is such an integral formatting choice, and I don't really see how this relates to the "minimal" in `MinimalStyle` -- so I don't really think that it should be implemented as a `MinimalStyle` overload. IMO we should do one of two things:

1. Treat it as a core philosophy of JuliaFormatter, i.e., just make it the default everywhere (but people will complain about changed formatting), or
2. Make it a config option that is disabled for Default but enabled for Minimal (well, people might still complain, but it will mostly impact VSCode users, so that's your call 😄).

Given that the code changes are not very large, I am inclined to believe that it's not too difficult to make (2) happen? I think basically `n_binaryopcall!` needs to be moved to an inner function, say `n_binaryopcall!_inner`, and then `n_binaryopcall!` just needs to check `s.opts.indent_chains` (or whatever the option is called) and dispatch accordingly. Likewise for the other overloaded functions.

Alternatively we could just modify every call site of `n_binaryopcall!` to check the option. I haven't counted, but if there's only one or two call sites maybe that's easier.

And personally I would be quite happy to have this as a config option with blue style :)

penelopeysm:
If we make it a config option, then the tests can go into `test/options/indent_chains.jl` (or whatever the name is) :)

davidanthoff:
Ok, just pushed a version that Claude created that makes it a config option, that does seem like the best idea.

HISTORY_v2.md
```diff
@@ -1,3 +1,17 @@
...
+
+New `indent_chains` option. When enabled, the continuation lines of operator chains (e.g. `|>` pipelines), comparisons, and binary operator calls are indented by one level relative to the expression instead of being aligned to the column of the first operand:
+
+```julia
+x = (aaa |>     # indent_chains = false
+     bbb)
+
+x = (aaa |>     # indent_chains = true
+    bbb)
+```
```
penelopeysm: this example could be clearer since the difference is only one space

docs/src/formatting_options.md
```diff
@@ -419,6 +420,50 @@ The number of spaces used for one level of indentation.
...
+and binary operator calls are laid out when the expression has to be split over several
+lines.
+
+When set to `false` the continuation lines are aligned to the column at which the expression
+starts:
+
+```julia
+x = (aaa |>
+     bbb |>
+     ccc)
```
penelopeysm: it would be ideal to run these as `@example` blocks with/without the option

penelopeysm (commented):
Given that I've been so tardy on this, I think it's only right that I fix my remaining comments, but just wanted to leave them for future reference!

docs/src/formatting_options.md
```diff
@@ -419,6 +420,50 @@ The number of spaces used for one level of indentation.
 
 There is at present no option to use tabs for indentation: please open an issue if you want this feature.
 
+## [`indent_chains`](@id options-indent-chains)
```
penelopeysm: I realise I used indent_chains in my prev comment, but truthfully this works for all binary operators anyway;

```julia
format_text("a + b"; margin=1, indent_chains=true) # --> "a\n    + b"
```

Having looked at it more closely, I think it makes sense to call it something more like `indent_binary_continuations` instead. That covers comparison operators too, which JuliaSyntax parses differently, but syntactically are really just binary operators as well.

davidanthoff:
> Given that I've been so tardy on this

I think these are great response times, I often sit for years on stuff :) But thanks for finishing it up in any case!

